Issue #672, opened 2014-10-17
[FEATURE] flux:form icon to show in page module


Idleworks:
In TYPO3 6.2 only icons of the content elements are displayed (names are hidden). Content elements of Flux always shows the TYPO3 icon, although in flux: form an icon was specified. 
Is it possible to show that icon in list/page module instead of default icon?

![ce-icon](https://cloud.githubusercontent.com/assets/6685173/4683882/f56bcafe-562a-11e4-8495-408e698027eb.jpg)

NamelessCoder:
Pull request with this feature is currently being processed - closing the issue.

PR #681, opened 2014-10-28
[FEATURE] Display icon of Form associated with record

NamelessCoder:
Close: #672


xf-:
Works only with icon set in options and not with Resources/Public/Icons/Content/Element-id.(gif|png..).
Should also automatically find the icon.

NamelessCoder:
Rebased, not yet adjusted to auto-icon.

NamelessCoder:
Adjusted to auto-icon; cache logic improved.

PR #704, opened 2014-11-06
[TASK] Add PreviewView and deprecate flux:widget.grid

danilobuerger:


NamelessCoder:
Picked as https://github.com/FluidTYPO3/flux/commit/23477b07d7b44e65f586aa19eb8500ebf0dc4fa8 - with the injection of PreviewView changed to a small object getter. HEREDOCs might be changed later but for now, it's working like it should. Thanks!

PR #769, opened 2015-02-20
[TASK] Small CodeStyle issue

cedricziel:


cedricziel:
Would have pushed directly, but have gone through a rebase and thought there may have been artifacts :)

Go get some points!

PR #777, opened 2015-03-04
Even more cleanups

cedricziel:


cedricziel:
Can you please spend extra care on https://github.com/cedricziel/flux/commit/fc29a0a1a3988bbcd3a6e8b8f6e8c1fa14fce439 while reviewing? I'm not a 100 percent sure, but i _think_ i used the right variable :)

Classes/Hooks/WizardItemsHookSubscriber.php
```diff
@@ -146,7 +146,7 @@ protected function getWhiteAndBlackListsFromPageAndContentColumn($pageUid, $colu
 		// Grid returned by the Provider that applies to the parent element's type and configuration
 		// (admitted, that's quite a mouthful - but it's not that different from reading the values from
 		// a page template like above; it's the same principle).
-		if (0 < $parentRecordUid && FALSE === empty($fluxAreaName)) {
+		if (isset($parentRecordUid) && 0 < $parentRecordUid && FALSE === empty($fluxAreaName)) {
```
NamelessCoder: I would initialise `$parentRecordUid = 0` up where `defaultValues` is set, then leave this condition as-is. Should be safe that way and we avoid a potentially undeclared variables.

Classes/ViewHelpers/AbstractFormViewHelper.php
```diff
@@ -37,7 +35,7 @@ public function render() {
 		$container = $this->getContainer();
 		$container->add($component);
 		// rendering child nodes with Form's last sheet as active container
-		$this->viewHelperVariableContainer->addOrUpdate(self::SCOPE, self::SCOPE_VARIABLE_EXTENSIONNAME, $extensionName);
+		$this->viewHelperVariableContainer->addOrUpdate(self::SCOPE, self::SCOPE_VARIABLE_EXTENSIONNAME, NULL);
```
NamelessCoder: I think this one should use `$this-> getExtensionName()` as default, rather than `NULL`. Most of our Form ViewHelpers will override this method anyway - but this way we will resolve the current name from VH var container if it is set.

NamelessCoder:
Not sure @cedricziel but I *think* most of this is redundant now - at least, all the stuff I commented on I have fixed separately. There might be two cases of not imported classes used in annotations and of potentially uninitialised values - but I'm guessing it would be much easier to kill those off in a fresh PR.

PR #779, opened 2015-03-13
[BUGFIX] Wrong sort order of  all RootPaths

monosize:
Here is the Problem
Wrong sort order when using more then one RootPath.

### Example
```typoscript
view {
  templateRootPaths {
    10 = EXT:a/Resources/Private/Templates/
    20 = EXT:b/Resources/Private/Templates/
  }
 ...
}
```

In the method `fillFromTypoScriptArray` 

Array templateRootPaths before krsort 
0 = EXT:a/Resources/Private/Templates/
1 = EXT:b/Resources/Private/Templates/

Array templateRootPaths after krsort 
1 = EXT:b/Resources/Private/Templates/
0 = EXT:a/Resources/Private/Templates/

Array templateRootPaths after `array_values(array_unique( ...))` 
0 = EXT:b/Resources/Private/Templates/
1 = EXT:a/Resources/Private/Templates/

Later in the method `resolveTemplateFileForControllerAndActionAndFormat` 
```php
foreach (array_reverse($this->templateRootPaths) as $templateRootPath) { ...}
````
with array_revers now the first used Path is
1 = EXT:a/Resources/Private/Templates/ 
Thats wrong!

In this pull request krsort is removed and a ksort is inserted in the method `extractPathArrays`
now the rootPath with the higher numeric key is used.

### DIFF
--- a/Classes/View/TemplatePaths.php
+++ b/Classes/View/TemplatePaths.php
@@ -255,9 +255,6 @@ class TemplatePaths {
 		// sort and unique, ensuring that the keys used in TypoScript become
 		// the dictated order of paths. The lower the number, the lower the
 		// path's priority is.
-		krsort($templateRootPaths, SORT_NUMERIC);
-		krsort($layoutRootPaths, SORT_NUMERIC);
-		krsort($partialRootPaths, SORT_NUMERIC);
 		$templateRootPaths = array_values(array_unique($templateRootPaths));
 		$partialRootPaths = array_values(array_unique($partialRootPaths));
 		$layoutRootPaths = array_values(array_unique($layoutRootPaths));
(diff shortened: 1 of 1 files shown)


NamelessCoder:
@monosize sorry to hijack the fix like this but it turns out the problem was a little deeper. There was one last thing to be done before the behavior was completely in sync with what you expect when you set view paths in (for example) a standard Extbase plugin.

See the commit for explanation - the TL;DR is that in vanilla Extbase, `getViewProperty` on `ActionController` will actually sort these. Flux doesn't use this method and needs the paths sorted also when using them outside of controllers - so 2+2, Flux TemplatePaths now sorts like `ActionController` would, plus support for `overlays.`.

PR #786, opened 2015-03-21
[TASK] Two foreach although one is enough

Madii:
I think it's enough that the last method iterates.

### DIFF
--- a/Classes/View/TemplatePaths.php
+++ b/Classes/View/TemplatePaths.php
@@ -199,7 +199,7 @@ class TemplatePaths {
 	protected function resolveFilesInFolders(array $folders, $format, $controllerName = '') {
 		$files = array();
 		foreach ($folders as $folder) {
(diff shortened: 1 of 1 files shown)


NamelessCoder:
I would prefer this stay the way it is, although technically, it makes sense the way you suggest. The reason is **semantic** but nonetheless I think it weighs heavy enough to dismiss this change.

Only when you operate in a `Template` context does it make sense to be able to specify a controller name - and none of the other contexts `Layout` or `Partial` have any requirement for an "append path segment"-type behaviour which would allow a controller name to be appended through a meaningful generic API.

Hope you understand! True, there is what appears to be a redundant iteration - but we are after all only talking about a handful of iterations total per template rendering and we avoid confusing the generic parts of the API with context-specific parameters.

Madii:
Yeah i understand that and i like that. I was learned to minifie code as possible and that's why i propossed that although my semantic sense didn't like it as you mentioned :)

Issue #573, opened 2014-05-29
Paste (copy/move), drag-n-drop and localisation bugs


NamelessCoder:
This issue is the **single** place to report **any** type of issue of this type:

> "My record is incorrectly positioned, contains incorrect values or was not created as expected after performing a paste/drag-n-drop command or after creating a localised version of the original record"

If your bug fits this description, report it here!

If any bugs in the mentioned areas still exist **after May 28th 2014** please report them here instead of creating new issues. Before you report problems, make sure your Flux repository is checked out on the `development` branch and that you have pulled (or possibly reset) to obtain the latest changes.

### Currently known bugs in copying:

* [Issue with missing re-referencing FAL elements in a copy of a record](#issuecomment-44843860) **SOLVED** - most recent TYPO3 CMS core `master` branch plus most recent flux `development` branch has this bug solved.

#### Related core patches:

https://review.typo3.org/#/c/30633/ which ensures passing the field name while TYPO3 iterates over fields of a record. Without this patch, the logic which retrieves files stored in FlexForm references, may in some cases fail to detect the existence of said references, because Flux receives a NULL field name and thus may return an incorrect Form instance (read: resolves the Provider with highest priority for the table while ignoring the field name in priority verdict) which does not contain the correct definitions for the record field currently being processed. Furthermore, due to Flux tracking which Provider was called with a certain field, Flux was unable to execute any but the first Provider detected through this logic in TYPO3. Applying the patch passes the field name and Flux returns the correct data structure when TYPO3 iterates record fields to detect file references in FlexForm fields.

https://review.typo3.org/#/c/30632/ adds an error prevention for records which have FlexForm fields that contain references that were not correctly processed (due to the bug above). Install this patch only if you are being faced with an Exception when attempting to edit affected records.

### Currently known bugs in moving:

* "Move record up" using sorting arrows causes record to disappear when child content record is **second** in list.

![687474703a2f2f692e696d6775722e636f6d2f6d697659336d632e706e67](https://cloud.githubusercontent.com/assets/2228052/4264020/e8797578-3c10-11e4-8208-cad3b920d50d.png)

### Currently known bugs in drag and drop:

* After dropping a content element from one `colPos` to another and then *immediately thereafter* using the "insert content after" link associated with the content record that was moved, the created content element is created inside the *original* `colPos` value. The content object may become invisible in page editing mode - to retrieve, use the list view and set proper `colPos` and as needed, `Relations` values.

### Currently known bugs in localisation:

[...]

PR #798, opened 2015-04-01
[BUGFIX] Add special provider resolve handling for moved records

coding46:
Fixes https://github.com/FluidTYPO3/fluidcontent/issues/233


CarstenD:
Fixes https://github.com/FluidTYPO3/flux/issues/573

misterboe:
this Fix only works if i move the Content under the Parent Element, if the Content is moved over the Parent Element the colPos is still 18181 and the Element is not Visible in BE

misterboe:
i also had to clear Cache to see the Changes in FE, after Content was moved

Issue #794, opened 2015-03-30
Disable Form work only in template with uid = 1


joneppie:
```plugin.tx_fluidbootstraptheme.forms.alert.enabled = 0```
form https://github.com/FluidTYPO3/flux/commit/8fe7e6efd1e60e81f5281e65d77343c7edf177d1 only work, if set in template with uid=1 and disable alert in all pages in the whole TYPO3-Installation

Version:
- TYPO3 6.2.11
- flux develop and master

joneppie:
I found a reason: When edit page the function getCurrentPageId in Class BackendConfigurationManager returns always 1 (which is the page with the first root-template)

PR #800, opened 2015-04-07
[BUGFIX] Set possible wrong currentPageId to null

joneppie:
Fixes https://github.com/FluidTYPO3/flux/issues/794

### DIFF
--- a/Classes/Configuration/BackendConfigurationManager.php
+++ b/Classes/Configuration/BackendConfigurationManager.php
@@ -56,6 +56,7 @@ class BackendConfigurationManager extends CoreBackendConfigurationManager implem
 			return $this->currentPageId;
 		}
 		$pageUids = $this->getPrioritizedPageUids();
+		$this->currentPageId = 0; // parent::getCurrentPageId() in getPrioritizedPageUids() set possible wrong value
 		while (TRUE === empty($this->currentPageId) && 0 < count($pageUids)) {
 			$this->currentPageId = array_shift($pageUids);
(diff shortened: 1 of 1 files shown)


joneppie:
```parent::getCurrentPageId()``` in line 74 return an id AND set ```$this->currentPageId``` to this value.  After return to ```public function getCurrentPageId``` in line 58 the while-loop will entry at no time because ```currentPageId``` has a value

File: Classes/Configuration/BackendConfigurationManager.php [hunk 1]
```diff
@@ -66,13 +66,15 @@ public function getCurrentPageId() {
...
 	protected function getPrioritizedPageUids() {
-		return array(
+		$pageUids = array(
 			$this->getPageIdFromGet(),
 			$this->getPageIdFromPost(),
 			$this->getPageIdFromRecordIdentifiedInEditUrlArgument(),
 			$this->getPageIdFromContentObject(),
 			parent::getCurrentPageId(),
 		);
+		$this->currentPageId = 0; // parent::getCurrentPageId() set possible wrong value
```
NamelessCoder: This line should be at line 59 instead so that right before the loop, we reset the internal variable. The reason for the problem if you have is that the loop never runs because the first check is immediately false (empty($this->currentPageId)). Rather than reset the value in this getter/resolver, which is bad, it should be reset right before determining the *new* active page UID.

NamelessCoder:
Perfect - thanks!

PR #804, opened 2015-04-09
[BUGFIX] Since TYPO3 CMS 6.2 the 'script' configuration is deprecated,  ...

monosize:
...use 'module' instead.

**Update: removed TYPO3_version check, its obsolete.**
**Sorry - Now also Updated ColorPicker, Select, Slider and Suggest, this was missing in the last pull request**
```createComponent``` in ```AbstractFormComponent``` overrides the defined name in the wizard classes to 'wizard', so only one wizard was displayed.
In the add wizard the 'params' definition was missing.
See also http://docs.typo3.org/typo3cms/TCAReference/6.2/AdditionalFeatures/WizardsConfiguration/

### DIFF
--- a/Classes/Form/Wizard/Select.php
+++ b/Classes/Form/Wizard/Select.php
@@ -34,11 +34,6 @@ class Select extends AbstractWizard {
 	/**
 	 * @var string
 	 */
-	protected $script = 'wizard_list.php';
-
-	/**
-	 * @var string
-	 */
 	protected $mode = 'substitution';
 
 	/**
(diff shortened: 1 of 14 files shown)

### CODE REVIEW

File: Classes/Form/Wizard/Select.php [hunk 1]
```diff
@@ -32,9 +32,11 @@ class Select extends AbstractWizard {
 	protected $icon = 'list.gif';
 
 	/**
-	 * @var string
+	 * @var array
 	 */
-	protected $script = 'wizard_list.php';
+	protected $module = array(
+		'name' => 'wizard_list'
```
NamelessCoder: This should probably be `'name' => 'wizard_select'` - that's the only thing I could spot in this PR :)

monosize:
There was a bug in  $structure array in Classes/Form/Wizard/Link.php - it's now corrected. 
https://github.com/FluidTYPO3/flux/commit/f5a12d5ca5ae4977741572f8e4f5e55e95b2d0d8#diff-d051c01dc06638b3494c469d743bc617L80

bjo3rnf:
Manually picked in 3f6512e9904b9ff1c9a8df0d4ad2dc55294540a3. Thanks @monosize.

PR #809, opened 2015-04-18
Fix extensionName override support

pkerling:
I'm a bit unsure about the AbstractFormComponent changes and their implications. I've removed the default extensionName setting as as far I could see it will get explictly set anyway and it simplifies the logic in resolveLocalLangValueOfLabel, but that assumption might be wrong.

### DIFF
--- a/Classes/Form/AbstractFormComponent.php
+++ b/Classes/Form/AbstractFormComponent.php
@@ -63,7 +63,7 @@ abstract class AbstractFormComponent implements FormInterface {
 	/**
 	 * @var string
 	 */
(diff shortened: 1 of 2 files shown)


NamelessCoder:
Looks like the test failures are caused by incomplete mocks and perhaps a bad expectation statement, and it looks like it's in the AbstractFormComponentTest as well as AbstractFormViewHelperTest but I didn't check for a more exact reason. Need a hand?

pkerling:
Haven't even run the tests locally... I'll take a look and come back at you if necessary.

PR #810, opened 2015-04-18
Fix extensionName override support

pkerling:
Follow-up for #809 

Tests run now, but $root->getExtensionName() is now completely gone. Is this a problem?

### DIFF
--- a/Classes/Form/AbstractFormComponent.php
+++ b/Classes/Form/AbstractFormComponent.php
@@ -156,6 +156,7 @@ abstract class AbstractFormComponent implements FormInterface {
 		$component->setLabel($label);
 		$component->setLocalLanguageFileRelativePath($this->getLocalLanguageFileRelativePath());
 		$component->setDisableLocalLanguageLabels($this->getDisableLocalLanguageLabels());
+		$component->setExtensionName($this->getExtensionName());
 		return $component;
 	}
 
--- a/Classes/Form/AbstractFormComponent.php
+++ b/Classes/Form/AbstractFormComponent.php
@@ -261,12 +261,11 @@ abstract class AbstractFormComponent implements FormInterface {
 		}
 		$name = $this->getName();
 		$root = $this->getRoot();
+		$extensionName = $this->extensionName;
 		if (FALSE === $root instanceof Form) {
 			$id = 'form';
-			$extensionName = $this->extensionName;
 		} else {
 			$id = $root->getName();
-			$extensionName = $root->getExtensionName();
 		}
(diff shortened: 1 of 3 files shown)


NamelessCoder:
> $root->getExtensionName() is now completely gone. Is this a problem?

Yes, that would cause forms constructed using PHP objects to no longer be able to inherit the extension name from the root element - but perhaps an easy way to solve that and maybe also avoid the entire ViewHelperVariableContainer situation:

https://github.com/FluidTYPO3/flux/blob/development/Classes/Form/AbstractFormComponent.php#L158

If an additional line here was would always copy `$this->extensionName` to the created component, and then the AbstractFormViewHelper used `$container` rather than `$form` to create new components, then this should actually cause the right behavior in all contexts. This is, if I am remembering everything around this code correctly!

NamelessCoder:
By the way, great stuff - thanks a lot for taking care of all these!

pkerling:
AbstractFormViewHelper does not use createComponent at all, components are created by the subclasses (ex: AbstractFieldViewHelper::getPreparedComponent, SheetViewHelper::render). From the looks of it they all use `$this->getForm()->create[...]`. Additionally, they also propagate the extension name themselves `$sheet->setExtensionName($this->getExtensionName())`. Now inserting the code you mentioned into `createComponent` should solve the problem with PHP-created forms and the `$root->getExtensionName()` call being deleted. For ViewHelper-created forms, I see two possibilities:
* Go through every view helper creating form components and make sure they use `$this->getContainer()->create[...]` instead and call `setExtensionName` themselves _only_ if it is explicitly set by some source (e.g. `extensionName` argument). Also, make sure the root component always gets the extension name from the ControllerContext as fallback.
* Just leave it as it is and use the ViewHelperVariableContainer

NamelessCoder:
Ach, right... the ViewHelpers do have a tendency to override the render method and don't share a "prepareComponent" method, at least not on the base class.

I'll vote for the solution that combines the PHP usage safeguard against removed $root combined with use of ViewHelperVariableContainer as suggested! Solves both problems with a minimum of changes in ViewHelpers' code.

pkerling:
Should I see if I can do something about the failed coverage check?

NamelessCoder:
You don't have to worry about those 0.02% - when we reach 99% I'll add new test cases.

NamelessCoder:
...but we do still need the `createComponent` setting of parent name to be implemented since getting rid of root as source of extension name!

pkerling:
Now I'm confused. Thought I addressed that in 8cc5c36, did I miss anything?

PR #812, opened 2015-04-24
[BUGFIX] Fix convention based form icons

### LINKED ISSUE #795: Convention-based template icons not set correctly
The automatically chosen template icons are not shown as the image src is not correct.
Currently (Flux 7.2.0), the absolute server path is used.

Temporarily, I amended line 85 in file [MiscellaneousUtility.php](https://github.com/FluidTYPO3/flux/blob/7.2.0/Classes/Utility/MiscellaneousUtility.php#L70) to get it to work:
```diff
- return (TRUE === is_array($filesInFolder) && 0 < count($filesInFolder) ? reset($filesInFolder) : NULL);
+ return (TRUE === is_array($filesInFolder) && 0 < count($filesInFolder) ? str_replace(PATH_site, '../', reset($filesInFolder)) : NULL);
```

coding46:
Fixes #795

### DIFF
--- a/Classes/Hooks/ContentIconHookSubscriber.php
+++ b/Classes/Hooks/ContentIconHookSubscriber.php
@@ -106,9 +106,8 @@ class ContentIconHookSubscriber {
 						if (NULL !== $form) {
 							$icon = MiscellaneousUtility::getIconForTemplate($form);
 							if (NULL !== $icon) {
-								$iconFileReference = '../../../' . $icon;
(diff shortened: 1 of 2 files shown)


NamelessCoder:
One of these days this really should be refactored a bit into perhaps its own class. It's something we'll be using a lot so it'd make sense to have this properly covered by tests that don't just check that this works with mocked and virtual values. But, works for now ;)

Issue #468, opened 2014-02-25
FAL fields always return 1 as value


JostBaron:
Steps to reproduce (flux dev 16cb322fd61c59874741796e9a99a105fedb6c56, TYPO3 6.2-dev b20aa02b0435c52d362cafffc697d85959276d41):

Create a form with a inline.fal field, fill that field (account for #431).

Echo the field value in the frontend - it is always "1".

PR #815, opened 2015-04-28
[BUGFIX] Make FAL working inside a section

kraftb:
Instead of creating a default FAL configuration using "foreign_field"
omit this field. This will result in a comma separated list of
sys_file_reference UIDs being stored in the flexform data field.

### DIFF
--- a/Classes/Form/Field/Inline/Fal.php
+++ b/Classes/Form/Field/Inline/Fal.php
@@ -156,6 +156,7 @@ class Fal extends AbstractInlineFormField {
 	 */
 	public function buildConfiguration() {
 		$configuration = $this->prepareConfiguration('inline');
+		unset($configuration['foreign_field']);
 		$configuration['appearance']['createNewRelationLinkTitle'] = $this->getCreateNewRelationLinkTitle();
 		return $configuration;
 	}
--- a/Classes/ViewHelpers/Field/Inline/FalViewHelper.php
+++ b/Classes/ViewHelpers/Field/Inline/FalViewHelper.php
@@ -74,7 +74,8 @@ class FalViewHelper extends AbstractInlineFieldViewHelper {
 		$component = $this->getPreparedComponent('Inline/Fal');
 		if (FALSE === is_array($this->arguments['foreignMatchFields'])) {
 			$component->setForeignMatchFields(array(
-				'fieldname' => $this->arguments['name']
+				// @todo: Retrieve this dynamically
+				'fieldname' => 'pi_flexform'
 			));
 		}
 		$component->setForeignSelectorFieldTcaOverride(array(


NamelessCoder:
This *definitely* has to be controlled by some parameter (for the VH as well as form component). Without that, this would break all existing records' saved values which we can't do to our users.

kraftb:
True. I am working on a new project where this problem occurs so I didn't think about a transition.
I have an alternate variant which "truly" implements FAL usage inside flexforms. But this requires a patch of the core and would also break backwards compatibility.

In the current situation sys_file_reference simply doesn't contain the information to have a correc reference to the flexform field.

I guess some kind of upgrade wizard would be necessary to handle this. The wizard could assume that no one uses flux+sections+fal currently. As this doesn't work anyways (https://github.com/FluidTYPO3/flux/issues/468).

Is there some place for upgrade wizards in flux?

NamelessCoder:
> Is there some place for upgrade wizards in flux?

https://github.com/FluidTYPO3/flux/blob/development/class.ext_update.php

But updating the DB is only part of the issue. The other part is the way FAL relations are currently consumed in PHP and Fluid templates - changing the way the value is stored also means every place that uses the values must also be changed. That's why this *must* be a flag where we preserve the current behavior as default. Such a flag can then later be deprecated, default can be changed and such - on a form-to-form basis.

The other issue is that this effectively turns the component from an IRRE to a CSV which logically means the ViewHelper name should even change.

The right solution is one that, in the TYPO3 core, fixes the problem with field names used in relations not being unique when the IRRE field is placed inside a section object - because this implies the name must be unique.

But it still implies a major breaking change any way you do it, because like before, relations are referenced by the field name and this field name must be changed to correctly retrieve the value. That's true whether it is IRRE or CSV.

In order to pass this you may need to submit it to the core with a toggle for the behavior and/or deprecation of the current method along with logging - as well as an update script.

Another way to gain this ability without breaking anything would be to add a toggle attribute for `flux:field.file` which when enabled, creates this CSV-as-IRRE configuration and a different component type - and when disabled (default) does what the ViewHelper currently does. Another major benefit of that method would be that the CSV-as-IRRE value storage is 100% compatible with a "group" field with internal_type "file" which is what `flux:field.file` currently creates.

kraftb:
This is my core patch which fixes the FAL/Flexform problems on the TYPO3 side: https://github.com/FluidTYPO3/flux/pull/815

It involves adding a flexfield_pointer_foreign field to sys_file_reference. In my opinion using the flexfield-name for the value in "field_foreign" of sys_file_reference is somewhat wrong (current situation). But I see the reasons for it: I made FAL fields outside of sections work - one just had to take care not to name a flexform field like "image" because this already natively exists in tt_content.

kraftb:
The linked core patch alone wouldn't break anything in flux or other setups. As long as the DS (TCA) configuration option "foreign_flexfieldPointer" isn't added to a DS the field will not get used. This field works in the same was as "foreign_field'": It names a field in the foreign table (sys_file_reference) which contains a reference to the field hosting the FAL references. But instead of pointing to just a field it contains a flexform pointer to the flexfield.

kraftb:
Of course the patch misses unit tests and a changelog entry :(

NamelessCoder:
There are still a number of problems that won't be solved by this or the core patch. The inner problem is this:

FlexForm objects inside sections are repeatable, yet have no unique ID.

Here's the current situation in Flux:

Flux tries to at least mitigate this by assigning an `id` attribute (hidden) to each object in XML when it is saved. This value would then be possible to place in HTML as a unique, permanent ID associated with just that object, and to use as (dynamic) relation target field name. **This allows the editor to sort objects inside sections without worrying about the reference**.

This is the solution we need. One that transparently is capable of assigning an additional XML attribute before saving objects and then also capable of 1) including this value in the FlexForm data - I suggest by using the "id" as keys in the array of objects in the section, and 2) using this value when posting data from TCEforms fields and 3) considering this "id" the "source ID of relation" by using it in `foreignMatchFields` if the existing configuration is empty. The major shift in this solution would be that objects inside flexform sections are no longer referenced by their numeric index - they become indexed only by their object IDs. However: it's very easy to write a conversion script which can convert existing relations as well as existing FlexForm XML.

But this should be handled by the core. We just have to be honest with ourselves: it's a temporary measure. The truly ideal situation is a switch away from FlexForms to a relational database with a nice migration script but until we have that, the above would be the second best option.

Sorry to reject/redirect this - given that this simply moves the problem to another location I can't accept this one. I would however happily review and advertise about a core patch that implements the suggested ID-attribute solution!

kraftb:
I am not sure about it, but AFAIK the index of flexform sections is constant. They do not get reindexed. So there would be no need for an attribute or additional ID element. I see your concerns about this issue. With my patch one would have to rely on static indexes for section elements ...

kraftb:
Right now I just came across such a case: In sys_refindex the path is generated exactly like I do it. Indeed I created the flexform-sections path exactly the way sys_refindex does so it is compatible.

kraftb:
Just search for "/el/" in a dump of sys_refindex table.

Issue #822, opened 2015-05-08
Impossible to move FCE inside another FCE in TYPO3 CMS 7.2


robsonrobi:
I just installed TYPO3 CMS 7.2. Everything looks great with the dev versions of fluidpages, flux and fluidcontent. But i am not able to move a nested content element. Trying to grab f.e. a content element inside a left column to put it in the right column, move the complete container CE|CE, not just the element inside.

Moving Content Elements up and down does not work either.

Any help would be great!

robsonrobi:
Is there any progress? I really love to use FLUX in TYPO3 7.2 :-)

georg-tiefenbrunn:
Same here using TYPO3 7.3-dev (TYPO3 and FluidTYPO3 dev using composer).

@robsonrobi: Using cut/paste you can achieve your goal:

1. You have your container FCE1 and create (or choose) a FCE2 outside (after) the container
3. create a new FCE3 *inside* the container FCE1
4. cut FCE2 and paste it *after* FCE3
5. delete FCE3

But you have to be very careful:

* pasting a FCE inside another FCE at first position does not work (the FCE keeps its position)
* pasting a FCE from inside a FCE to outside the FCE (into a column) does not work (no icon "Paste in this position" is displayed for columns)
* pasting a FCE from inside a FCE into another FCE at first position does not work; the FCE is unlinked (youl'll have to correct fields ``tt_content.tx_flux_column`` and ``tt_content.tx_flux_parent`` in DB)
* moving records down doesn't seem problematic. But moving a FCE to position 1 using "Move record up" will unlink the FCE.

@NamelessCoder: Should these TYPO3 7.2/7.3 specific issues also be added to #573 or should we wait until the new PreviewView (sub)class mentioned in https://github.com/FluidTYPO3/flux/pull/823#issuecomment-100647141 is ready?

robsonrobi:
@georg-tiefenbrunn : Thanks for your answer. Me as a programmer it is okay doing that workaround, but for the rest it is not possible to use it in that way. Looking really really forward to a stable fix in the next version. :-)

Thumbs up for FLUX!

NamelessCoder:
@georg-tiefenbrunn Please wait for the reworked preview-view class for 7.2+, that's likely to remove some of the problem(s) - and feel free to coordinate with @monosize if perhaps you've got more time to implement something like that!

robsonrobi:
Is there date you plan to finish it?

[...]

PR #823, opened 2015-05-08
[BUGFIX] Flux grid not working in TYPO3 CMS 7.2

monosize:
Tested in TYPO3 CMS 7.1 and TYPO3 CMS 7.2 
There is a little cosmetic bug: While moving a FCE inside another FCE: The dropzone of the dragged element is displayed twice, when the element has moved over other dropzones.
#822

### DIFF
--- a/Classes/View/PreviewView_7_2.php
+++ b/Classes/View/PreviewView_7_2.php
@@ -1,0 +1,87 @@
+<?php
+namespace FluidTYPO3\Flux\View;
+
+/*
+ * This file is part of the FluidTYPO3/Flux project under GPLv2 or later.
+ *
+ * For the full copyright and license information, please read the
+ * LICENSE.md file that was distributed with this source code.
+ */
+
+use FluidTYPO3\Flux\Form\Container\Column;
+use FluidTYPO3\Flux\Service\ContentService;
+use TYPO3\CMS\Backend\View\PageLayoutView;
+
+/**
+ * @package Flux
+ */
+class PreviewView_7_2 extends PreviewView {
+
+	/**
+	 * @param array $row
+	 * @param Column $column
+	 * @return string
+	 */
+	protected function drawGridColumn(array $row, Column $column) {
+		$colPosFluxContent = ContentService::COLPOS_FLUXCONTENT;
+
+		$dblist = $this->getInitializedPageLayoutView($row);
+		$this->configurePageLayoutViewForLanguageMode($dblist);
+		$records = $this->getRecords($dblist, $row, $column->getName());
+
+		$content = '';
+		foreach ($records as $record) {
+			$content .= $this->drawRecord($row, $column, $record, $dblist);
+		}
+
+		$id = 'colpos-' . $colPosFluxContent . '-page-' . $row['pid'] . '--top-' . $row['uid'] . '-' . $column->getName();
+		$target = $this->registerTargetContentAreaInSession($row['uid'], $column->getName());
+
+		return <<<CONTENT
+		<td colspan="{$column->getColspan()}" rowspan="{$column->getRowspan()}" style="{$column->getStyle()}">
+			<div data-colpos="{$colPosFluxContent}" class="t3js-sortable t3js-sortable-lang t3js-sortable-lang-{$dblist->tt_contentConfig['sys_language_uid']} t3-page-ce-wrapper ui-sortable" data-language-uid="{$dblist->tt_contentConfig['sys_language_uid']}">
+				<div class="fce-header t3-row-header t3-page-colHeader t3-page-colHeader-label">
+					<div>{$column->getLabel()}</div>
+				</div>
+				<div class="t3-page-ce t3js-page-ce" data-page="{$target}">
+					<div class="t3js-page-new-ce t3-page-ce-wrapper-new-ce" id="{$id}" style="display: block;">
+                        {$this->drawNewIcon($row, $column)}
+						{$this->drawPasteIcon($row, $column)}
+						{$this->drawPasteIcon($row, $column, TRUE)}
+					</div>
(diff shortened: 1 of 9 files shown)


NamelessCoder:
Thanks for the patch, @monosize - however, since this is one area that's likely to change again in the future (it already did from 6.2 -> 7.1, but fortunately those markups were still compatible with the old classes/containers) we prefer to split this into a new PreviewView (sub)class which can then be used when the version of TYPO3 is detected. This other approach is a lot cleaner and testable, and once we no longer need to support a specific version of TYPO3 we can simply remove the entire class and the condition case that would use it. The subclass then only overrides the methods that actually need to change.

In this case it seems fine to use the replacement PreviewView class from 7.1 and upwards. When the new LTS comes out we'll make the base class support that and change the condition that loads the alternative PreviewView.

monosize:
Here is a patch with subclasses. 

May be there is a better solution for the TYPO3 version switcher...

lsoendergaard:
Hi,

I merged this locally on 76a28ef (flux 7.2.1) and tried this on a TYPO3 7.2 Installation. It still has a few issues, when trying to drag and drop a content element in the first position of a flux grid element (any other position seems to work fine).

Sometimes the content elements just disappear (and I have to retrieve them from "List"), sometimes they just pop up somewhere else. e.g. I moved a content element from "col2" of grid element uid=30, and tried to move it to "col1" of the same grid element at the first position (any other position worked fine).

After reloading the page, the content element somehow showed up in "col3" (not col1) of grid element with the uid=28. But consistent, even if I create a new content element, this exact case is reproducible.

Copy & Pasting works fine with flux 7.2.1 after that one bug was fixed.

Classes/View/PreviewView_7_2.php
```diff
@@ -0,0 +1,87 @@
...
+ */
+
+use FluidTYPO3\Flux\Form\Container\Column;
+use FluidTYPO3\Flux\Service\ContentService;
+use TYPO3\CMS\Backend\View\PageLayoutView;
+
+/**
+ * @package Flux
+ */
+class PreviewView_7_2 extends PreviewView {
```
NamelessCoder: Correct me if I'm wrong, but: is it not possible to combine this markup (the DOM structure as well as class names) generated by both these functions, in the default PreviewView? As far as I can tell from reading, those added containers should not cause problems. That way you wouldn't need this class, the Preview_7_2 class or the switching that goes on in `ext_localconf.php`.

Resources/Public/js/VersionSevenPointTwo/DragDrop.js
```diff
@@ -0,0 +1,159 @@
...
+ *
+ * For the full copyright and license information, please read the
+ * LICENSE.md file that was distributed with this source code.
+ */
+
+/**
+ * Important! The original DragDrop.js must be replaced with this one.
+ * so we undefine the module via requirejs!
+ */
+requirejs.undef("TYPO3/CMS/Backend/LayoutModule/DragDrop");
```
NamelessCoder: Is there no way to avoid this? Personally I don't consider it safe to replace core JavaScript. What is it precisely that we need these modified methods to do - would it be possible to contribute this to the core itself, as simple flexibility/compatibility precautions that increase support for modules/hooks that produce HTML markup in the page module.

ext_localconf.php
```diff
@@ -32,8 +32,17 @@
 $GLOBALS['TYPO3_CONF_VARS']['SC_OPTIONS']['t3lib/class.t3lib_tcemain.php']['clearCachePostProc'][] = 'FluidTYPO3\Flux\Backend\TceMain->clearCacheCommand';
 $GLOBALS['TYPO3_CONF_VARS']['SC_OPTIONS']['t3lib/class.t3lib_tstemplate.php']['includeStaticTypoScriptSources']['flux'] = 'FluidTYPO3\Flux\Backend\TypoScriptTemplate->preprocessIncludeStaticTypoScriptSources';
 $GLOBALS['TYPO3_CONF_VARS']['SC_OPTIONS']['GLOBAL']['extTablesInclusion-PostProcessing']['flux'] = 'FluidTYPO3\Flux\Backend\TableConfigurationPostProcessor';
-$GLOBALS['TYPO3_CONF_VARS']['SC_OPTIONS']['cms/layout/class.tx_cms_layout.php']['tt_content_drawItem']['flux'] = 'FluidTYPO3\Flux\Backend\Preview';
 
+$previewClassPostfix = '';
+$version = ((float) substr(TYPO3_version, 0, 3));
+switch (TRUE) {
```
NamelessCoder: Depending on the above comments this may be a non-issue, but I should mention it anyway:

The naming of this class is invalid and likely to cause problems with code detecting wether or not a class name uses namespaces. Usually, we replace the implementations of classes by using the "class name registry" which allows overruling the implementation that ObjectManager will load for each class that can be loaded.

On a side note, in this case I think it would be preferable to have a "Legacy" class or sub-scope and fall back to it if the version is LTS, otherwise use the default class. Since from now on we'll only ever be requiring support for most recent LTS and most recently released iteration. But avoiding splitting the classes would be preferred ;)

smichaelsen:
Can't comment on any code details, but I can verify that with this patch drag 'n' drop works with current TYPO3 master. Thanks!

NamelessCoder:
Thanks for confirming, @smichaelsen - we will definitely be going in this direction, we just need to determine how much we truly need to override for this to work. Of course, we aim for an absolute minimum of overriding.

jmverges:
We also need to take care of this one: https://github.com/FluidTYPO3/fluidpages/issues/246 it could be fixed just adding the typo3pageModule.js included in fluidpages extension

xf-:
Fixed via #854

PR #833, opened 2015-05-15
[BUGFIX] Prevent infinite nesting of FCEs

### LINKED ISSUE #824: No protection against inserting a content element as its own child
It is somehow possible (don't know why, and can't reproduce it, but an editor did it) to insert a content element as its own child. Doing this has the effect that the page module can't be opened any longer for that page because of an infinite recursion. 

This could be mitigated by adding two checks:

 1. If a content element has children, and a child has the same ID than the element itself, don't render it (in previews and in the frontend). Maybe print a warning instead, or silently ignore it.
 2. If a content element is pasted into another element, and both uids are the same, don't do it and show a warning.

BenjaminBeck:
Prevent infinite nesting by aborting row processing in ContentService.
Show a flash message in the backend if row processing has been aborted.

Fixes #824

### DIFF
--- a/Classes/Service/ContentService.php
+++ b/Classes/Service/ContentService.php
@@ -11,8 +11,11 @@ namespace FluidTYPO3\Flux\Service;
 use FluidTYPO3\Flux\Utility\MiscellaneousUtility;
 use TYPO3\CMS\Backend\Utility\BackendUtility;
 use TYPO3\CMS\Core\DataHandling\DataHandler;
+use TYPO3\CMS\Core\Messaging\FlashMessage;
+use TYPO3\CMS\Core\Messaging\FlashMessageService;
 use TYPO3\CMS\Core\SingletonInterface;
 use TYPO3\CMS\Core\Utility\GeneralUtility;
+use TYPO3\CMS\Extbase\Utility\LocalizationUtility;
 
 /**
  * Flux FlexForm integration Service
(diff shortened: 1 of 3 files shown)


NamelessCoder:
Until a nested content solution is present in the core, please keep this as simple as possible - it is obvious that no parent should be able to copy or reference itself into children and IMHO it is perfectly valid and sufficient to silently ignore this. If you insist on having messaging/logging then please fire a simple "devLog" logging instruction along with the otherwise silent ignoring ;)

The alternative and ideal solution is one that simply removes the "paste" icon(s) when rendering the nested content grid, if parent is detected in clipboard, but I'm not sure that is possible given how we use the content element renderer and table lister.

BenjaminBeck:
Ok - then i will remove the Messaging.
Because pasting is possible on many different ways (context-menu, icons, list-view, ..) i think this solution is good to easily protect from all cases.
One thing that came to my mind is that with the current implementation it still is possible to create circular nesting if the paste action is done inside a sub-sub-grid (gridelement inside a gridelement)  - because only the direct parent is checked..
Should i also check for that?

NamelessCoder:
Yes; when checking the parent we should definitely be checking the entire root line of UIDs to avoid that situation.

JostBaron:
Thanks for your work :-)

BenjaminBeck:
Sorry for the long wait - i will do this in the next days.

NamelessCoder:
I'm afraid Juan beat you to it, Benjamin ;)

Thank you nonetheless both for reporting and providing a fix, much appreciated!

PR #834, opened 2015-05-19
Fix for editing children elements in workspace

thomasc-clio:
If you are in workspace, child elements can not be edited by pressing on pencil icon. Permissions error will be thrown, becaus alt_doc.php does not receive id of edit record. This small commit fixes that, but it could propably be done more nicely.


xf-:
I didn't tested it. 
Please use one commit and a prefix in your commit message.

NamelessCoder:
@xf- would you mind cherry-picking/squashing this one?

NamelessCoder:
Manually picked - thanks, Thomas!

Issue #831, opened 2015-05-13
Template fallback quit working at this commit


Outdoorsman:
When I updated to the most recent verion of Flux I found that my template fallback just stopped working. I have isolated it to this @NamelessCoder commit in Flux where it quit working:
https://github.com/FluidTYPO3/flux/commit/7a541dd5777a2db7a11347034a438594ecf3e0f0

Here is the way that I'm implementing fallback...

My provider_extension has this in setup.txt
```
plugin.tx_fluidbootstraptheme.view {
    templateRootPaths.1 = {$plugin.tx_providerextension.view.templateRootPaths.0}
    partialRootPaths.1 = {$plugin.tx_providerextension.view.partialRootPaths.0}
    layoutRootPaths.1 = {$plugin.tx_providerextension.view.layoutRootPaths.0}
}
```

fluidbootstraptheme (v2.0 branch, compatible with TYPO3 7.x) has this in setup.txt
```
plugin.tx_fluidbootstraptheme.view {
    templateRootPaths.0 = {$plugin.tx_fluidbootstraptheme.view.templateRootPaths.0}
    partialRootPaths.0 = {$plugin.tx_fluidbootstraptheme.view.partialRootPaths.0}
    layoutRootPaths.0 = {$plugin.tx_fluidbootstraptheme.view.layoutRootPaths.0}
}
```

doerler:
I think, this happens because since this commit: aacefc0 the default view configuration is added with the array key 10. Later on all configuration paths are sorted by key: [TemplatePaths.php](../blob/7.2.1/Classes/View/TemplatePaths.php#L337).

This means that all overlay paths with a key below 10 are overriden by this new default path.

Outdoorsman:
@doerler you are correct. As soon as I changed from 1 to 11 (or anything larger than 10), it works. I believe that Flux should be changed from 10 to 0 though since it is the base template. I noticed that this is the way that the news extension does it and I have adopted it elsewhere because it really makes sense.
```
plugin.tx_fluidbootstraptheme.view {
    templateRootPaths.11 = {$plugin.tx_providerextension.view.templateRootPaths.0}
    partialRootPaths.11 = {$plugin.tx_providerextension.view.partialRootPaths.0}
    layoutRootPaths.11 = {$plugin.tx_providerextension.view.layoutRootPaths.0}
}
```

PR #840, opened 2015-05-26
[BUGFIX] Change from 10 to 0 for templateRootPaths.0.xxxx

Outdoorsman:
I recommend changing this to 0 as I noticed others doing with extensions like the news extension. This already cause me a lot of grief because I did not know that 10 was the starting number... see my issue here, https://github.com/FluidTYPO3/flux/issues/831.

### DIFF
--- a/Classes/Service/FluxService.php
+++ b/Classes/Service/FluxService.php
@@ -287,9 +287,9 @@ class FluxService implements SingletonInterface {
 	protected function getDefaultViewConfigurationForExtensionKey($extensionKey) {
 		$extensionKey = ExtensionNamingUtility::getExtensionKey($extensionKey);
 		return array(
-			TemplatePaths::CONFIG_TEMPLATEROOTPATHS => array(10 => 'EXT:' . $extensionKey . '/Resources/Private/Templates/'),
-			TemplatePaths::CONFIG_PARTIALROOTPATHS => array(10 => 'EXT:' . $extensionKey . '/Resources/Private/Partials/'),
-			TemplatePaths::CONFIG_LAYOUTROOTPATHS => array(10 => 'EXT:' . $extensionKey . '/Resources/Private/Layouts/'),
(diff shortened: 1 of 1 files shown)


Outdoorsman:
Any feedback on this? I really believe this is a good thing to do.

NamelessCoder:
I agree this should be done. However, this currently breaks the unit tests (but does not afaics have negative impact on current installations) so we need those fixed - I'll take the liberty of calling @cedricziel here to do a manual cherry-pick of this. Hope he doesn't mind.

There are some places in documentation where the `.10` index is used - Charles would you mind taking a look at submitting pull requests for those? Possible places are in README.md files of flux, fluidpages, fluidcontent and fluidcontent_core and in the `documentation` repository that is rendered online :)

jmverges:
I can confirm .10 index is not in any README.md file. It only appears in documentation https://github.com/FluidTYPO3/documentation/blob/f080e06e7b920714229c95aad2fce824779feb28/3.Templating/3.1.ProviderExtension/3.1.5.ConfigurationFiles.md where could stay as is .10 because is referring to a provider extension.
It also appears in: https://github.com/FluidTYPO3/fluidcontent_core/blob/6389f37047de0afb7f78d0cee6c657669ca3b481/Tests/Unit/Provider/CoreContentProviderTest.php
https://github.com/FluidTYPO3/flux/blob/4bb25d45e36df72586e5ffef19a79b84035d1d86/Tests/Unit/Service/FluxServiceTest.php
https://github.com/FluidTYPO3/flux/blob/4bb25d45e36df72586e5ffef19a79b84035d1d86/Tests/Unit/Service/FluxServiceTest.php
https://github.com/FluidTYPO3/fluidcontent/blob/b3b4da75f9a338d266e1c1f94b7b7c6719083fcb/Tests/Unit/Service/ConfigurationServiceTest.php
not in builder, not in fluidpages, not in site

NamelessCoder:
Picked by Juan, now merged! Thanks Charles and Juan :)

Issue #491, opened 2014-03-23
Allowing specific fluidcontent elements in <flux:form.variable name="allowedContentTypes">


twirsing:
There is probably a good reason why this is not possible but I am suggesting it anyway. 

Changing typo3conf/ext/flux/Classes/Hooks/WizardItemsHookSubscriber.php:147 from

```
if (FALSE !== strpos($name, '_') && FALSE === in_array($item['tt_content_defValues']['CType'], $whitelist)) {       
```

to 

```
if (FALSE !== strpos($name, '_') && FALSE === in_array($item['tt_content_defValues']['CType'], $whitelist) && FALSE === in_array($name, $whitelist)) {
```

would enable to define specific fluidcontent elements. The definition terms are certainly a bit strange but it works for me.

PR #848, opened 2015-06-17
Further restrict available content elements by list_type

krokodok:
I really suck at PHP so I most certainly won't understand why this is a bad idea, but I had to modify the WizardItemsHook slightly.

I really just copied the idea from here: https://github.com/FluidTYPO3/flux/issues/491

In this issue the idea was to further restrict available content elements in the wizard for fluidcontent-elements. This issue was closed because there is now official support for this feature in fluidcontent. However, it is still not possible to further restrict the available plugins by list_type. To do that I had to re-implent the code from https://github.com/FluidTYPO3/flux/issues/491 on line 277.

    if (FALSE !== strpos($name, '_') && FALSE === in_array($item['tt_content_defValues']['CType'], $whitelist) && FALSE === in_array($name, $whitelist)) {

I would be not the least bit surpirsed if there's a better way to this, however, I realy need this feature. I'd be glad to use the official feature upon release instead of my little hack.

### DIFF
--- a/Classes/Hooks/WizardItemsHookSubscriber.php
+++ b/Classes/Hooks/WizardItemsHookSubscriber.php
@@ -274,7 +274,8 @@ class WizardItemsHookSubscriber implements NewContentElementWizardHookInterface 
 		$whitelist = array_unique($whitelist);
 		if (0 < count($whitelist)) {
 			foreach ($items as $name => $item) {
-				if (FALSE !== strpos($name, '_') && FALSE === in_array($item['tt_content_defValues']['CType'], $whitelist)) {
+				if (FALSE !== strpos($name, '_') && FALSE === in_array($item['tt_content_defValues']['CType'], $whitelist) && FALSE 
+=== in_array($name, $whitelist)) {
 					unset($items[$name]);
 				}
 			}


jmverges:
Hi @krokodok I don't understand at all what is this about. Could you provide a example of how are you using it? I would like to be able to reproduce your problem first and then understand what is your solution. However for the next time you want to make a PR you should follow our contribution guide: https://fluidtypo3.org/documentation/contributing/contribution-guide.html

krokodok:
I'm sorry to hear that, I'm not very familiar with GitHub and rarely use it.

The problem:

In Flux it is possible to restrict the available content types in the new-content-element-wizard with the attribute _variables="{allowedContentTypes: ''}"_.

Example:
><flux:grid.column name="main" label="LLL:flux.settings.grid.main"
>	style="background-color:#ddf;"
>	variables="{allowedContentTypes:
>	'text,
>	image,
>	uploads,
>	fluidcontent_content',
>	Fluidcontent: {allowedContentTypes: 'providerExtension:TwoColumns.html, providerExtension:Accordion.html'}}" />

In this list it is only possible to allow or disallow the usage of the plugin-element. It is not possible to further restrict _which plugins_ should be allowed or disallowed. By adding the line from my first post, which I got from https://github.com/FluidTYPO3/flux/issues/491, this becomes possible again.

Example:
><flux:grid.column name="main" label="LLL:flux.settings.grid.main"
>	style="background-color:#ddf;"
>	variables="{allowedContentTypes:
>	'text,
>	image,
>	uploads,
>	plugins_tx_myExtension_plugin1,
>	plugins_tx_myExtension_plugin2,
>	plugins_tx_myExtension_plugin3,
>	fluidcontent_content',
>	Fluidcontent: {allowedContentTypes: 'providerExtension:TwoColumns.html, providerExtension:Accordion.html'}}" />

jmverges:
Hi @krokodok 

after we discussed it and for further plans we decided don't accept this feature.

If you have your extension and you have several plugins and you want to be able to restrict any of them in a container element you should create a custom content type for each one.

You can get more information about how to do it here: http://stackoverflow.com/questions/18464356/how-to-quickly-create-custom-content-elements-in-typo3-6-x

Regards

PR #854, opened 2015-07-22
[BUGFIX] Fix drag and drop issues for current and 6.2 LTS

jmverges:
This is based on @monosize https://github.com/FluidTYPO3/flux/pull/823
It solves problems with 7 > and 6.2 LTS: https://github.com/FluidTYPO3/fluidpages/issues/246

### DIFF
--- a/Classes/Backend/LegacyPreview.php
+++ b/Classes/Backend/LegacyPreview.php
@@ -1,0 +1,40 @@
+<?php
+namespace FluidTYPO3\Flux\Backend;
+
+/*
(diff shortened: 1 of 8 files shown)


xf-:
code looks good, but needs intense testing. Could break a lot of stuff.
Is this also master T3 compatible?

jmverges:
I think so, I tested it manually with two Vagrants one with 6-2 and the other one with 7-3, both are working fine at the same time. UT are also passing. We should remove the js file from fluidpages once this is merged.

Issue #853, opened 2015-07-16
Template icons not displayed in Preview (Page mode)


sdelcroix:
Hi,

Since Flux 7.2, template icons of our fluidcontent contents are not displayed in Preview, in TYPO3 "Page" mode.

We used both methods (path in arguments "options" and the convention-based one), none worked.
After code inspection, we found that the path of the icon file is relative to the TYPO3 main directory, not to the path of the current PHP script that calls/displays the icon.

We patched it by adding backpath context to the *getIconForTemplate* method in *MiscellaneousUtility.php* file (line 82) as following.

`$iconRelFolder = $GLOBALS['BACK_PATH'] . ExtensionManagementUtility::extRelPath($extensionKey) . 'Resources/Public/Icons/' . $controllerName . '/';`

xf-:
Works fine in my 7.3.1 instance.
I needed to clear the cache.
Please show me the 404, the path generated or somethingsimilar.

deerfish:
Guten Tag
 
Vielen Dank für Ihre Nachricht.

Ich bin vom Montag, 20. Juli  2015 bis und mit Freitag, 14. August 2015 ferienhalber abwesend. 
 
In dringenden Fällen dürfen Sie sich bei Hugo Stalder, 041 266 0034, hugo.stalder@appline.ch melden.


Gute Grüsse und frohe Sommerwochen wünscht

Damian Parpan



Damian Parpan
appline
Hirschengraben 52
6003 Luzern
Switzerland

T: +4141 266 0034
M: +4179 303 4878

Email: damian.parpan@appline.ch
www.appline.ch

xf-:
Please reopen it, if the issue still exists.
(Worked also in my 7.2 a view months ago)

sdelcroix:
Hi xf,

7.3.1 ? Don't you mean 7.2.1 ?

Here the wrong icon file URL generated in Page mode (with convention-based path) : 

`http://_domain_/typo3/sysext/cms/typo3conf/ext/rvvnfluidphotoalbums/Resources/Public/Icons/Content/MyContentElement.gif`

See that is relative to typo3/sysext/cms/

Caches have been cleared multiple times.

flash4light:
i can confirm this issue. The relativ path of the icons in the preview mode is wrong. Source is something like "../typo3conf/ext/my_ext/Resources/Public/Icons/icon-fluid-cols.gif" which must be
[...]

PR #856, opened 2015-07-24
[BUGFIX] Fix for icons in Preview for 6.2 LTS and master

jmverges:
Closes: #853

As solution we are returning absolute urls instead of relative paths. I thought about another "if" but it is not possible since the issue only occurs in Preview and not in Edit.


xf-:
i prefer avoiding domains. reason: caching proxies...
Absolute path:
```php
$iconAbsoluteUrl = '/' . str_replace(PATH_site, '', $iconFolder);
```
should work

PR #859, opened 2015-07-25
[BUGFIX] Disallow pasting element inside it self (endless loop)

xf-:
Please note:
If you put gird in gird in gird without direct parent it will pass and still create a endless loop.
We would need always to search to the top. Create a proper page layout.


xf-:
Fix https://github.com/FluidTYPO3/fluidcontent/issues/220

NamelessCoder:
Please fix CGL issues. Code looks good (but prevents only direct parent pasting; we need to also consider the rootline but it's fine to do that in another pull request).

xf-:
CGL fixed

PR #868, opened 2015-07-31
change copy/move and paste method 

Marc-Bauer:
including workspaces and fixes bugs in TYPO3 7.X


NamelessCoder:
Needs a rebase - solution does look good, but should be placed in ContentProvider instead (LightweightProvider is since removed). We also need to address the code style and commit issues, but a rebase is first on the list :)

Outdoorsman:
PING - I'm just hoping to see something happening here as the work is already done :)

NamelessCoder:
Replaced by https://github.com/FluidTYPO3/flux/pull/910 with many thanks to Marc for pointing to the source of the problem!

PR #869, opened 2015-08-02
[TASK] CGL - Remove multiple variable assignment, fix indents and exceptions

xf-:


Classes/View/LegacyPreviewView.php
```diff
@@ -54,17 +54,17 @@ class LegacyPreviewView extends PreviewView {
...
-     * @param $colPosFluxContent
-     * @param $dblist
-     * @param $target
-     * @param $id
-     * @param $content
-     * @return string
-     */
+	 * @param array $row
+	 * @param Column $column
+	 * @param $colPosFluxContent
```
NamelessCoder: While you are doing CGL anyway: please add parameter types for this and the next handful of arguments.

xf-: thx, variables already included, param still missing

NamelessCoder: What I mean is that the `@param $colPosFluxContent` should be `@param integer $colPosFluxContent`, `@param $dblist` should be `@param DatabaseRecordList $dblist` etc.

xf-:
Done

Issue #864, opened 2015-07-28
Translated content got "tx_flux_parent=1" in tt_content, although there was no parent


adudtkowski:
I had the mistake that translated content got "tx_flux_parent=1" in tt_content, although there was no parent. Thereby, there were errors in the sort.

I fixed it in flux\Classes\Service\ContentService.php in initializeRecordByNewAndOldAndLanguageUids():
```php
protected function initializeRecordByNewAndOldAndLanguageUids($row, $newUid, $oldUid, $newLanguageUid, $languageFieldName, DataHandler $tceMain) {
		if (0 < $newUid && 0 < $oldUid && 0 < $newLanguageUid) {
			$oldRecord = $this->loadRecordFromDatabase($oldUid);
			if ($oldRecord[$languageFieldName] !== $newLanguageUid && $oldRecord['pid'] === $row['pid']) {
                // look for the translated version of the parent record indicated
                // in this new, translated record. Below, we adjust the parent UID
                // so it has the UID of the translated parent if one exists.
                if ($oldRecord['tx_flux_parent'] > 0) {
                    $translatedParents = (array)$this->workspacesAwareRecordService->get('tt_content', 'uid', "t3_origuid = '" . $oldRecord['tx_flux_parent'] . "'");
                    $translatedParent = reset($translatedParents);
                } else {
                    $translatedParent = NULL;
                }

                $sortbyFieldName = TRUE === isset($GLOBALS['TCA']['tt_content']['ctrl']['sortby']) ?
                    $GLOBALS['TCA']['tt_content']['ctrl']['sortby'] : 'sorting';
				$overrideValues = array(
					$sortbyFieldName => $tceMain->resorting('tt_content', $row['pid'], $sortbyFieldName, $oldUid),
					'tx_flux_parent' => NULL !== $translatedParent ? $translatedParent['uid'] : $oldRecord['tx_flux_parent']
				);
				$this->updateRecordInDatabase($overrideValues, $newUid);
			}
		}
	}
```

xf-:
Hi,
you maybe want to contribute and create a pull request?
We can test that better - simple checkout, discuss code and you contribute :)

jmverges:
Hi @adudtkowski,

here you have our contribution guide: https://fluidtypo3.org/documentation/contributing/contribution-guide.html It is quite easy

Regards

jmverges:
I can confirm this issue.
[...]

PR #872, opened 2015-08-04
[BUGFIX] Fix tx_flux_parent always equal to 1 when is translated

jmverges:
Close: #864

### DIFF
--- a/Classes/Service/ContentService.php
+++ b/Classes/Service/ContentService.php
@@ -283,8 +283,12 @@ class ContentService implements SingletonInterface {
 				// look for the translated version of the parent record indicated
 				// in this new, translated record. Below, we adjust the parent UID
 				// so it has the UID of the translated parent if one exists.
-				$translatedParents = (array) $this->workspacesAwareRecordService->get('tt_content', 'uid', "t3_origuid = '" . $oldRecord['tx_flux_parent'] . "'");
-				$translatedParent = reset($translatedParents);
+				if ($oldRecord['tx_flux_parent'] > 0) {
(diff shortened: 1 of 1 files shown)


Classes/Service/ContentService.php
```diff
@@ -283,8 +283,12 @@ protected function initializeRecordByNewAndOldAndLanguageUids($row, $newUid, $ol
 				// look for the translated version of the parent record indicated
 				// in this new, translated record. Below, we adjust the parent UID
 				// so it has the UID of the translated parent if one exists.
-				$translatedParents = (array) $this->workspacesAwareRecordService->get('tt_content', 'uid', "t3_origuid = '" . $oldRecord['tx_flux_parent'] . "'");
-				$translatedParent = reset($translatedParents);
+				if ($oldRecord['tx_flux_parent'] > 0) {
+					$translatedParents = (array)$this->workspacesAwareRecordService->get('tt_content', 'uid', "t3_origuid = '" . $oldRecord['tx_flux_parent'] . "'");
```
NamelessCoder: Space between casting type and variable being cast. Other than that, looks good!

woodyc79:
Hello!

When will this find it's way to master or TER?
It cost me a whole day, a little bit of reputation of TYPO3 and me and additional 50€ for recovering the whole database.... because of this bug simply copying the Content Element with uid 1 led to copying all those referenced 500 additional Content Elements, ... and finaly deleting that Content Element deleted ALL translated Contents from a LIVE Page!  

Thank you!
Regards,

nabossha:
Hi - would appreciate to see this merged in the TER releases too - because i had the same problem as @woodyc79 with about 2'500 elements.... 
thank you very much!

PR #873, opened 2015-08-04
[BUGFIX] for issue #573: move record up to top position

galoppi:


Classes/Service/ContentService.php
```diff
@@ -222,7 +221,15 @@ public function moveRecord(array &$row, &$relativeTo, $parameters, DataHandler $
 				$row['tx_flux_column'] = $relativeToRecord['tx_flux_column'];
 				$row['colPos'] = $relativeToRecord['colPos'];
 				$row['sorting'] = $tceMain->resorting('tt_content', $relativeToRecord['pid'], 'sorting', abs($relativeTo));
+			} elseif (0 < (integer) $relativeTo) {
+				// moving to first position in colPos, means that $relativeTo is the pid of the containing page
+				$row['sorting'] = $tceMain->getSortNumber('tt_content', 0, $relativeTo);
+				// neither change fields tx_flux_column nor tx_flux_parent here!
+			} else {
+				$row['tx_flux_column'] = $row['tx_flux_parent'] = NULL;
```
xf-: In future we try to avoid multiple assignments for better readable code.
Variable is always in first block.

NamelessCoder:
On a side note here, I think this would be the opportune time to convince the `gridelements` team that *violating the rules this way and intentionally and abusively passing a value that is invalid is bad, mkay. It ruins it for everyone else*.

PR #874, opened 2015-08-04
Bugfix/871/exeption page module

galoppi:


Classes/Provider/AbstractProvider.php
```diff
@@ -274,7 +274,7 @@ public function getForm(array $row) {
 				$form = call_user_func_array(array($formClassName, 'create'), array($row));
 			} else {
 				$templatePathAndFilename = $this->getTemplatePathAndFilename($row);
-				if (NULL !== $templatePathAndFilename) {
+				if (NULL !== $templatePathAndFilename && '' != $templatePathAndFilename) {
```
jmverges: I think it should be TRUE === empty($templatePathAndFilename)

xf-:
Please only push 1 issue in a PR.
-> 2 commits with different purposes.

NamelessCoder:
The fix in e3aede76c3d34ef7f1d6ed346bb3eec34ec46c98 is already applied earlier; at the root of the problem rather than where template file gets used. Subject message of commit needs to be updated. Needs rebasing or cherry picking.

NamelessCoder:
Both changes manually picked - thanks @galoppi!

Issue #871, opened 2015-08-04
Uncaught TYPO3 Exception: file_get_contents(): Filename cannot be empty in Backend Page module


galoppi:
related to commit bbe0a21bdea7163c6b8f82ec3ba6f4447b537ee4:

You can't access the page module anymore because of this exception.
Tested with TYPO3 6.2.14, latest development-Branches of flux, fluidcontent,fluidpages,vhs.

Cause:
Flux is now also registered as Content Provider but has no implementation of the method
getTemplatePathAndFilename().
So the AbstractProvider::getTemplatePathAndFilename() method is called that returns an empty string and not NULL.

Possible fix:
\FluidTYPO3\Flux\Provider\templatePathAndFilename line 277:

replace 
`if (NULL !== $templatePathAndFilename) {`
with
`if (NULL !== $templatePathAndFilename && '' != $templatePathAndFilename) {`

But I don't know if flux should really be triggered here.
Before the mentioned commit only fluidcontent was resolved as provider in \FluidTYPO3\Flux\Service\FluxService::resolveConfigurationProviders.

xf-:
hi,
https://github.com/FluidTYPO3/flux/blob/development/Classes/Provider/ContentProvider.php#L80 FALSE loads avoids the error. Not tested yet, we are writing internally.

jmverges:
@xf- I can confirm the error in master branch of TYPO3

PR #876, opened 2015-08-04
[BUGFIX] for issue #871: avoid exception in TYPO3 BE page module

galoppi:


xf-:
Hi,

won't fix the issue. if you remove the return in trigger it will be rendert in backend preview.

monosize:
Hi,

works for me in TYPO3 7.2, 7.3 and also in 7.4

PR #881, opened 2015-08-08
[FEATURE] Conveniently take defaults when new record is created

cedricziel:
This patch introduces the `useColumnsForDefaultValues` setting in TCA
 for tt_content and will enable copying of the mentioned columns to newly created
 records.

 The effect should be that the new record will be a sibling in a possible
 parent container.

References #868


NamelessCoder:
We have this in fluidcontent currently, I think. I requested they be moved to Flux but maybe that has not happened yet. In any case: please make sure it isn't there already, and create a corresponding PR to remove it from fluidcontent.

xf-:
wrong @NamelessCoder 
https://github.com/cedricziel/flux/blob/wsFixes/ext_tables.php#L13

NamelessCoder:
> wrong

Not wrong - like I said, I just wasn't sure if both parts had been done yet. I really did request it moved and it really was in fluidcontent originally ;)

PR #887, opened 2015-08-11
[BUGFIX] In the web layout module the checkbox "Show hidden content e…

monosize:
…lements" has no effect for nested flux grids.

### DIFF
--- a/Classes/View/PreviewView.php
+++ b/Classes/View/PreviewView.php
@@ -482,6 +482,7 @@ class PreviewView {
 		// records based on them having remove placeholders.
 		$condition = "AND tx_flux_parent = '" . $row['uid'] . "' AND tx_flux_column = '" . $area . "' ";
 		$condition .= "AND colPos = '" . ContentService::COLPOS_FLUXCONTENT . "' ";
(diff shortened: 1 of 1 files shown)


Classes/View/PreviewView.php
```diff
@@ -482,6 +482,7 @@ protected function getRecords(PageLayoutView $view, array $row, $area) {
 		// records based on them having remove placeholders.
 		$condition = "AND tx_flux_parent = '" . $row['uid'] . "' AND tx_flux_column = '" . $area . "' ";
 		$condition .= "AND colPos = '" . ContentService::COLPOS_FLUXCONTENT . "' ";
+		$condition .= $view->tt_contentConfig['showHidden'] ? '' : 'AND hidden = 0 ';
```
xf-: Please write a compare like TRUE ===
Also you can use a Elvis operator ?: 'AND hidden = 0 '

xf-:
Commit message should be a little bit shorter. (you can use multiline)

monosize:
A Compare with TRUE === and an Elvis operator ?:'AND hidden = 0' is not possible - the value from $GLOBAL['BE_USER']... is a string "0" or "1". Now the value is converted to an integer.

xf-:
+1

PR #888, opened 2015-08-11
[TASK] Translate select options

xf-:
Finally resolve select options.
Example:
```html
<flux:field.select name="test" items="1,2,3,4" />
```

```xml
			<trans-unit id="flux.example.test.option.1" xml:space="preserve">
				<source>Option</source>
			</trans-unit>
			<trans-unit id="flux.example.test.option.2" xml:space="preserve">
				<source>Option 2</source>
			</trans-unit>
			<trans-unit id="flux.example.test.option.3" xml:space="preserve">
				<source>Option 3</source>
			</trans-unit>
```
Result
builder8

This also allows to translate headlines in fluidcontent_core
Headline XY....
Normal Headline

### DIFF
--- a/Classes/Form/Field/Select.php
+++ b/Classes/Form/Field/Select.php
@@ -67,7 +67,11 @@ class Select extends AbstractMultiValueFormField {
 		} elseif (TRUE === is_string($this->items)) {
 			$itemNames = GeneralUtility::trimExplode(',', $this->items);
 			foreach ($itemNames as $itemName) {
-				array_push($items, array($itemName, $itemName));
+				$resolvedLabel = $this->resolveLocalLanguageValueOfLabel('', $this->name . '.option.' . $itemName);
+				if (0 === strpos($resolvedLabel, 'LLL:') ) {
+					$resolvedLabel = $itemName;
+				}
+				array_push($items, array($resolvedLabel, $itemName));
 			}
 		} elseif (TRUE === is_array($this->items) || TRUE === $this->items instanceof \Traversable) {
 			foreach ($this->items as $itemIndex => $itemValue) {


NamelessCoder:
This is not the right way to achieve this, sorry. The method used (and indeed any method to do the same) would inevitably break other functionality and would be created in a way that relies on the edge case behavior of showing LLL references when no translation exists. And you put significant load on the backend form rendering in that Flux has to attempt translating every single value regardless of it having a translation or not. We just can't trust that the values coming from this CSV can be used as segments in an LLL reference.

The other aspect is that `items` on the select field were never intended to be manipulated apart from allowing the ViewHelper to support CSV and oppositely indexed (key is field value, value is label) arrays.

The right way to go about this would be to make the `items` an extended array:

```xml
<flux:field.select items="{0: 'LLL:EXT:myext/Resources/....', 1: 'LLL:EXT:...'} />
```

I know it's not pretty, nor is it easy to remember, but it both performs better and won't break any other use cases or generate invalid reference (unless, of course, you intentionally enter an invalid reference).

More magic is not always a good thing ;)

xf-:
Fixed

xf-:
closed the wrong one

smichaelsen:
I have made pull request with (IMO) improvements on @xf- 's patch. Maybe that was not "the github way" to do it, but here it is: https://github.com/xf-/flux/pull/3

NamelessCoder:
Please address the concerns in the comment above and suggest an alternative solution, for example one that triggers this behavior only when a certain `option` (Flux options) is set on the field. It is okay to have this, but it absolutely has to be a conscious decision made by the developer in each case.

NamelessCoder:
Manually picked and put behind a flag: `translateCsvItems="1"` on `flux:field.select` will now trigger this behavior.

See https://github.com/FluidTYPO3/flux/commit/c5c79ea3f6b42c5abe48b24fab0a96d326cbc7df

PR #915, opened 2015-09-10
[TASK] Drop custom paste icons

NamelessCoder:
Reasons are as follows:

* Paste icons require yet another custom format on the two supported versions.
* Paste icons currently broken on both versions.
* Paste icons require custom styling on each version.
* Paste icons have no style equivalent on 7.x
* 7.x in general has no paste icons (uses clickmenu)
* Expecting core solution (possibly not until core's container content feature is done).

Suggest dropping those icons now, as part of Flux 7.3 release.

### DIFF
--- a/Classes/View/LegacyPreviewView.php
+++ b/Classes/View/LegacyPreviewView.php
@@ -77,7 +77,7 @@ class LegacyPreviewView extends PreviewView {
 			$column->getLabel(),
 			$target,
 			$id,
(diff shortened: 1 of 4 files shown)


xf-:
we should split legacy brnach before

xf-:
@NamelessCoder please rebase - i crated a legacy branch

NamelessCoder:
Done :)

PR #918, opened 2015-09-11
[BUGFIX] Show only the preview if CType is equal to 'fluidcontent_con…

monosize:
…tent'

 Avoid rendering of the preview if the field 'tx_fed_fcefile' isn't empty, and also CType isn't equal to 'fluidcontent_content'.
 This happens, if you change the CType of an 'fluidcontent_content' element.

### DIFF
--- a/Classes/Backend/Preview.php
+++ b/Classes/Backend/Preview.php
@@ -86,23 +86,25 @@ class Preview implements PageLayoutViewDrawItemHookInterface {
 		} else {
 			$itemContent = '<a name="c' . $row['uid'] . '"></a>' . $itemContent;
 		}
-		$providers = $this->configurationService->resolveConfigurationProviders('tt_content', $fieldName, $row);
-		foreach ($providers as $provider) {
-			/** @var ProviderInterface $provider */
-			list ($previewHeader, $previewContent, $continueDrawing) = $provider->getPreview($row);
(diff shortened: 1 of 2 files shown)


NamelessCoder:
The `Preview` section must be supported for any content type that contains Flux fields. The current suggested change would break, among other things, fluidcontent_core - and would cause E_NOTICE when using Flux without fluidcontent. It's essential that we avoid patching Flux to accommodate other extensions (there is one single exception to that right now... a temporary measure to support the extensive manipulation Gridelements does to copy/paste commands etc.).

If this change should be placed anywhere, it would be in fluidcontent itself. If you want to suppress the preview rendering, it would have to be in the ContentProvider (possibly needs overridden getPreview method) by returning NULL as preview content and header. To me it sounds perfectly reasonable to suppress the preview like that, if the `tx_fed_fcefile` field is empty (or even testing if a template file can be resolved which would also prevent rendering if you for example have a single, bad template path).

My apologies for just closing like this but hopefully you agree with the reasons above ;)

And thanks for all your work by the way!

PS: you're always welcome to stop by the IRC channel to discuss bugs/features or whatnot. We'd be happy to see you :)

monosize:
Ahh, you are right!
I will take a look on the ContentProvider of fluidcontent ;)

PR #921, opened 2015-09-16
Over-all Patch for Localization/Move/Drag & Drop for TYPO3 CMS 6.2.14

galoppi:


NamelessCoder:
https://github.com/sitegeist/flux/commit/e8f86b0ff075f41baf5b00afbeb5563ab5431b0b manually picked and adjusted to avoid direct array access syntax - as https://github.com/FluidTYPO3/flux/commit/737c4a92b99ff95d3b94603a6a61a2c797910e23

NamelessCoder:
https://github.com/sitegeist/flux/commit/ac96c6f024b053510c1d75cb0ae3034bf17d88db manually picked and CGL-adjusted, and modified to use a temporary variable for column name, as https://github.com/FluidTYPO3/flux/commit/7c67a554cb55db19832d48261d71be0a9471be62

NamelessCoder:
https://github.com/sitegeist/flux/commit/debefc7440279d13a968e869e367ca91b27210e1 manually picked, fitted with unit test, fixed class header and CGL - as https://github.com/FluidTYPO3/flux/commit/c0864b4f02ccd7382146e6f15737bf3576d2ed0e

mkoehler2911:
I tried the new dev version for flux:
7.2.2 dev and typo3 6.2.14

1. The localization button is available but the element I tried to localize is literally invisible in column view until I activate the visibility of my localized element.
2. The localized element is at the wrong position.
3. Elements inside of a container aren't visible in list view anymore.

galoppi:
@mkoehler2911: Thanks for your bugreport!
Could you please try to uncheck and recheck the checkbox "Show hidden elements" at the bottom of the page module? Do the localized elements appear after that?
And at which position was the original element in the default language? (First, somewhere in the middle or last element or the only element in the container)?

mkoehler2911:
@galoppi unchecking/checking the "Show hidden elements" checkbox did it. Now, those elements are visible.
The original element is in the middle of my container and the localized element appears at the end of the container.
Thanks for that quick reply!

robsonrobi:
I am having the same problem with the position. The localized object is put at the bottom, but has to be somewhere in the middle. Would be so cool if that can be fixed, too. Localization is then again fun! Great job so far.

galoppi:
@robsonrobi: For the correct positioning is a pull request pending: #930. And there is another bugfix to fix some trouble with deleted localized flux containers and disappearing localized children: #929

robsonrobi:
@galoppi : Thanks. One further question. Is it difficult to integrate the fixes for a working version in TYPO3 7.4 ?

NamelessCoder:
@galoppi Except for the functional tests, we've merged all these right? In any case, I'll close here and we'll discuss about the functional tests via email ;)

PR #923, opened 2015-09-17
[TASK] Added unit test for moving a content element to first position in a column

galoppi:


NamelessCoder:
Manually picked as https://github.com/FluidTYPO3/flux/commit/cbc2b63d55257bfe2f61b577b05681f13141c9e0 - thanks!

PR #926, opened 2015-09-18
[BUGFIX] Only show flux preview in page modul if CType of content ele…

galoppi:
…ment is fluidcontent_content

Avoids trouble with elements that have been flux elements and have been changed from editors to f.e. shortcuts. These shortcuts have been displayed with the flux grid before this bugfix

### DIFF
--- a/Classes/View/PreviewView.php
+++ b/Classes/View/PreviewView.php
@@ -129,6 +129,9 @@ class PreviewView {
 	 * @return string
 	 */
 	public function getPreview(ProviderInterface $provider, array $row) {
+		if ($row['CType']!=='fluidcontent_content') {
+			return '';
+		}
 		$form = $provider->getForm($row);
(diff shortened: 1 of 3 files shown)


NamelessCoder:
Hi Alex,

This one is already implemented in fluidcontent: https://github.com/FluidTYPO3/fluidcontent/pull/278

We avoid all special handling of other extensions in Flux itself. There is a single exception currently with a legacy reason behind it: a special marker from Gridelements is supported for some edge case systems running 6.2. So since this one specifically targets fluidcontent, it must go in fluidcontent.

The solution seems to do the same as your proposed fix here but please confirm.

galoppi:
ok, I'll check that, thanks!

galoppi:
Ok, I checked it. FluidTYPO3/fluidcontent#278 fixed this problem.

NamelessCoder:
Excellent - thanks for confirming!

PR #930, opened 2015-09-23
[BUGFIX] Correct positioning for localizations of new inserted conten…

galoppi:
…t elements

### DIFF
--- a/Classes/Provider/ContentProvider.php
+++ b/Classes/Provider/ContentProvider.php
@@ -140,6 +140,9 @@ class ContentProvider extends AbstractProvider implements ProviderInterface {
 					$this->contentService->moveRecord($row, $relativeTo, $moveData, $reference);
 				}
 			}
+			if ('localize' === $command) {
+				$this->contentService->fixPositionInLocalization($id, $relativeTo, $row, $reference);
+			}
 			self::trackMethodCallWithClassName(__CLASS__, __FUNCTION__, $id);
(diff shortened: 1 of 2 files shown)


NamelessCoder:
Picked and converted to sprintf instead of concatenation as https://github.com/FluidTYPO3/flux/commit/58602dd9166e2e12fb14e0755d6bccd5455b728a

mkoehler2911:
I'm very excited of this bugfix!
For 2 languages everything works fine (localizing, positioning etc.).
I've a backend with 5 languages and it seems that this bugfix can't handle this. 

- I created an element inside a container element (L=0).
- I localized that element with the help of the button (in L=1).
- The element seemed to be localized but I couldn't find it. 
- I checked the database and found my localized element in the wrong language (L=3).

I checked the language (L=3) and the element was at the correct position but wrong language.
Any solution for that?

Typo3 6.2.15
flux 7.2.3

BenjaminBeck:
hi @mkoehler2911 
your issue should be fixed in development by #946

mkoehler2911:
Great job!
Localization makes me finally happy again.

I found one small bug while localizing elements:
- I create an element in my default language and localize that element in my 2nd language (works fine)
- I create a container in my default language and pull my created (and localized) element inside this container
- trying to localize the container (with my element)

bug: my container is at the right position but my element jumps to the top of my page

PR #936, opened 2015-09-26
[BUGFIX] Icon position in production

xf-:
Icon position worked only development mode.

### DIFF
--- a/Resources/Public/css/icon.css
+++ b/Resources/Public/css/icon.css
@@ -10,7 +10,7 @@
 
 .fluidcontent-icon {
 	margin-top: 7px;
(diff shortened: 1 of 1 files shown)


Resources/Public/css/icon.css
```diff
@@ -10,7 +10,7 @@
 
 .fluidcontent-icon {
 	margin-top: 7px;
-	position: absolute;
+	position: absolute!important;
```
NamelessCoder: Can we fix this without `!important` please. Other people will care ;)

xf-:
reopen - 1 moment

PR #941, opened 2015-09-30
[BUGFIX] Make PathUtility work as expected with arrays

Cash2m:
- Drop unnecessary array key check
- Check for types explicitly
- Use GeneralUtility for checking absolute paths

### DIFF
--- a/Classes/Utility/PathUtility.php
+++ b/Classes/Utility/PathUtility.php
@@ -38,19 +38,33 @@ class PathUtility {
 	 * @return mixed
 	 */
 	public static function translatePath($path) {
-		if (is_array($path) == FALSE) {
-			$path = (0 === strpos($path, '/') ? $path : GeneralUtility::getFileAbsFileName($path));
+		if (is_string($path)) {
+			$path = GeneralUtility::isAbsPath($path) ? $path : GeneralUtility::getFileAbsFileName($path);
 			if (is_dir($path)) {
 				$path = realpath($path) . '/';
 			}
 			$path = GeneralUtility::fixWindowsFilePath($path);
-		} else {
-			foreach ($path as $key => $subPath) {
-				if (TRUE === in_array($key, self::$knownPathNames)) {
-					$path[$key] = self::translatePath($subPath);
+		} elseif (is_array($path)) {
+			$templatePaths = self::getTemplatePathFields($path);
+			foreach ($templatePaths as $field => $subpath) {
+				if (is_array($subpath)) {
+					$path[$field] = array_map(array(__CLASS__, __METHOD__), $subpath);
+				} else {
+					$path[$field] = self::translatePath($subpath);
 				}
 			}
 		}
 		return $path;
+	}
+
+	/**
+	 * Get fields from view configuration that are template paths
+	 *
+	 * @param array $viewConfiguration
+	 * @return array
+	 */
+	public static function getTemplatePathFields(array $viewConfiguration) {
+		$knownPathKeys = array_fill_keys(self::$knownPathNames, NULL);
+		return array_intersect_key($viewConfiguration, $knownPathKeys);
 	}
 }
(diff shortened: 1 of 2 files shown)


Classes/Utility/PathUtility.php
```diff
@@ -19,36 +18,21 @@
...
 	 * Translates an array of paths or single path into absolute paths/path
 	 *
 	 * @param mixed $path
 	 * @return mixed
 	 */
 	public static function translatePath($path) {
-		if (is_array($path) == FALSE) {
-			$path = (0 === strpos($path, '/') ? $path : GeneralUtility::getFileAbsFileName($path));
+		if (is_string($path)) {
+			$path = GeneralUtility::isAbsPath($path) ? $path : GeneralUtility::getFileAbsFileName($path);
```
NamelessCoder: Is `isAbsPath` available on 6.2 as well?

Cash2m: yes

Classes/Utility/PathUtility.php
```diff
@@ -19,36 +18,21 @@
 class PathUtility {
 
 	/**
-	 * @var array
-	 */
-	private static $knownPathNames = array(
```
NamelessCoder: We can drop the legacy names but we do need to check for the path being one of the recognised paths - the reason is that the view configuration array may contain other flags (`enabled` which is used by Fluidpages, for example) and we don't want to translate those.

Cash2m: Okay, now I see the point

Classes/Utility/PathUtility.php
```diff
@@ -38,19 +38,26 @@ class PathUtility {
...
+		return NULL;
+	}
+
+	/**
+	 * Get fields from view configuration that are template paths
+	 *
+	 * @param array $viewConfiguration
+	 * @return array
+	 */
+	public static function getTemplatePathFields(array $viewConfiguration) {
```
NamelessCoder: I see where you're going with this extra method, and I'm not opposed to adding new methods - but the `translatePaths` method is widely used also in dependencies and the behavior of it must be preserved. We've moved quite far from the original problem now and have accumulated some side effects. It's overkill to avoid checking if the path is in a known array.

We could consider adding a second argument to `translatePath` called `$translateAny` which defaults to `FALSE` and if set to `TRUE` skips the `in_array` check in the recursive part of `translatePath`. The default should be to work exactly as before.

Thanks for adding a unit test btw ;) I didn't even know we lacked that one.

Cash2m: Sorry for the inconvenience. The original problem was that the method was not working 100% on `templateRootPaths/partialRootPaths/...`. Before we go full circle give me one last chance to keep the original behavior and only fix the little bug with that additional method.

NamelessCoder:
Assuming you're ready for review ;)

I understand the intention - to allow `translatePath` to work with *any* type of path, not just template paths, when passing an array to it (given that `translatePath` already allows *any* type of path to be translated if passed as a string).

And I understand the use case for this - being able to call the method in your own code and/or to implement it in various places in FluidTYPO3 in the future.

But, and it's a rather heavy "but":

I don't see a sane way of doing this without the thing you're in the progress of doing, but this approach also means that *every place we use `translatePath` currently and assume it only translates template paths, we have to switch to another method*. And given that it is a rather edge use case (wanting to pass arbitrary arrays of paths for translation) that already has an existing way that does exactly that (`array_map($ref, $pathsArray)`) it's quite a big ask.

Not to sound dismissive but in order to save us both some time:

* If you found a problem that makes the original `translatePath` method fail on any of the intended supported path configuration names, please point out the problem and I'll either fix it or suggest a fix that doesn't require this much API change.
* The `GeneralUtility::isAbsPath` is very much desired so please don't hesitate to create *that* as an individual pull request!
* The behavior change is not desired. If you want this new behavior I suggest a second argument for `translatePath` as mentioned earlier. But the behavior must remain the same without requiring changes in third-party extensions.

Cash2m:
Hi, thanks for you patience and the excessive review! I will close this and maybe split this into multiple PR somewhere in the future. Have a nice weekend!

PR #942, opened 2015-09-30
[BUGFIX] Use relative path for template icons

Cash2m:
Page icon path always started with `/` which only works if your TYPO3 installation is on root level

### DIFF
--- a/Classes/Utility/MiscellaneousUtility.php
+++ b/Classes/Utility/MiscellaneousUtility.php
@@ -61,7 +61,7 @@ class MiscellaneousUtility {
 	}
 
 	/**
-	 * Returns the icon for a template
+	 * Returns the relative http icon path for a template
 	 * - checks and returns if manually set as option or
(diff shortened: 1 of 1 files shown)


xf-:
the problem is v6 vs v7 - v7 adds a prefix.

Cash2m:
What kind of prefix? Just tried with latest 6.2, 7.4 and 7.5 and saw no issue.

xf-:
you tried add a icon manually, via detection and it worked in wizard and fluidpage in all versions?

Cash2m:
Page icons are working now in all three stated versions.  
But the FCEs are not showing up at all in TYPO3 7... In 6.2 they work though. Sounds like another bug?

jmverges:
Hey guys,

I'm going to close this since it was fixed somehow from other PR for 7LTS. Feel free to reopen this if you experiment the problem again.

PR #952, opened 2015-10-07
translate only in non-default language

lsascha:
Make sure page values are only used from pages_language_overlay when not in default language.


lsascha:
looking over it again. I think it also does not take into account that language_overlays can be deleted in backend and column "deleted" is only written to 1 in the Database. Shouldn't it also check if the overlay is not deleted?

PR #953, opened 2015-10-07
[TASK] Only show children of a content element if it has the correct …

JostBaron:
Usability improvement


JostBaron:
This is wrong here, better put that into EXT:fluidcontent.

PR #956, opened 2015-10-08
Remove deprecated dependency "cms"

RauchF:
The TYPO3 CMS sysext `cms` has been deprecated, and as such, extensions which define a dependency on `cms` may cause the dependency notice to be logged in the deprecation log:

> Extension "flux" defines a dependency on ext:cms, which has been removed. Please remove the dependency.

It is not a big deal, but unnecessary.


xf-:
CGL - Commit message, please use a prefix

NamelessCoder:
Oops - pushed this one, didn't see the pull request! Also removed from other packages.

Issue #961, opened 2015-10-12
Flux loses element if moved in workspaces


JostBaron:
In the sequence flux loses content elements:
1. There are two content element types: A container element having a grid for children, and "normal" elements.
2. In LIVE workspace, create a container element, and one child for it.
3. Switch to a draft workspace
4. Create a second container element
5. Move the child of the first container to the second one (either drag & drop or copy & paste)
6. The child is gone (if you used drag & drop you need to reload the view).

Here is a simple provider extension to reproduce this:

https://github.com/JostBaron/fluxworkspacestest

I have some time assigned to check this stuff out - can you tell me where stuff like this is handled?

JostBaron:
There seem to be two parts to this problem.

**First:** The command that is executed in this case is `'version'`. Neither the hook subscriber [here][1] nor the one in the superclass handles that command.

**Second:** Even if one adds the `'version'` command to the `$pasteCommands`, it does not work.In the [`moveRecord()`-method][2], no move placeholder is created, and there is no preexisting one, because the record was not previously versioned.

So it seems there are two things that need to be done here:

 1. Handle `'version'` commands
 2. Make sure that placeholder records are created before they are updated. Not sure whose responsibility this is, but probably the one of EXT:workspaces.

 [1]: https://github.com/FluidTYPO3/flux/blob/development/Classes/Provider/ContentProvider.php#L129
 [2]: https://github.com/FluidTYPO3/flux/blob/development/Classes/Service/ContentService.php#L200

JostBaron:
Ok, here is the text from pull request #962, so I can close that:

One problem is, that provider methods are not always executed, because of the caching mechanism that tracks the method calls and limits them to 1 call/request. The problem is that this mechanism only tracks method calls based on these things:

 * method name
 * ID of the modified record
 * Class name of the provider.

But when using workspaces, multiple commands can be issued per request that create the same cache identifier. E.g. in the case described in #961 there are two commands issued:

 * One with `'version'` as command. The provider method of `FluidTYPO3\Flux\Provider\ContentProvider` does nothing for this one, except tracking the method call, in two places: `FluidTYPO3\Flux\Provider\ContentProvider->postProcessCommand`, and also in `FluidTYPO3\Flux\Backend\TceMain->executeConfigurationProviderMethod`.
 * One with `'move'` as command. The provider method should do something on this command, but because the method call has been tracked, nothing happens here.

[...]

PR #962, opened 2015-10-13
[BUGFIX] Don't call provider methods for command 'version'.

JostBaron:
This is a partial fix for #961, but it needs improvement before, so don't merge it please.

The problem there is that provider methods are not always executed, because of the caching mechanism that tracks the method calls and limits them to 1 call/request. The problem is that this mechanism only tracks method calls based on these things:

 * method name
 * ID of the modified record
 * Class name of the provider.

But when using workspaces, multiple commands can be issued per request that create the same cache identifier. E.g. in the case described in #961 there are two commands issued:

 * One with `'version'` as command. The provider method of `FluidTYPO3\Flux\Provider\ContentProvider` does nothing for this one, except tracking the method call, in two places: `FluidTYPO3\Flux\Provider\ContentProvider->postProcessCommand`, and also in `FluidTYPO3\Flux\Backend\TceMain->executeConfigurationProviderMethod`.
 * One with `'move'` as command. The provider method should do something on this command, but because the method call has been tracked, nothing happens here.

The pull request prevents tracking of the method call, but this is probably the wrong solution. A proper solution would fix the tracking of method calls.

Any ideas on that?

### DIFF
--- a/Classes/Backend/TceMain.php
+++ b/Classes/Backend/TceMain.php
@@ -97,9 +97,11 @@ class TceMain {
 	 * @return void
 	 */
 	public function processCmdmap_preProcess(&$command, $table, $id, &$relativeTo, &$reference) {
-		$record = array();
-		$arguments = array('command' => $command, 'id' => $id, 'row' => &$record, 'relativeTo' => &$relativeTo);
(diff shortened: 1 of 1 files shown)


JostBaron:
This is not a good idea, closing it.

PR #964, opened 2015-10-14
[BUGFIX] Call provider methods for different commands.

### LINKED ISSUE #961: Flux loses element if moved in workspaces

JostBaron:
Not only once for each method/id combination.

Partly resolves #961. Moving the first content element into a newly created container works now, but moving elements after this first element does still not work.

### DIFF
--- a/Classes/Backend/TceMain.php
+++ b/Classes/Backend/TceMain.php
@@ -99,7 +99,7 @@ class TceMain {
 	public function processCmdmap_preProcess(&$command, $table, $id, &$relativeTo, &$reference) {
 		$record = array();
 		$arguments = array('command' => $command, 'id' => $id, 'row' => &$record, 'relativeTo' => &$relativeTo);
-		$this->executeConfigurationProviderMethod('preProcessCommand', $table, $id, $record, $arguments, $reference);
+		$this->executeConfigurationProviderMethod('preProcessCommand', $table, $id, $command, $record, $arguments, $reference);
 	}
 
 	/**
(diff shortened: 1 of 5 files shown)


NamelessCoder:
Hi Jost,

Please rebase this on current development branch and remove the `[!!!]` from the commit message - this does not break current implementations since you've been clever enough to make the `$command` optional with a default value ;)

Everything else looks good and we'd like to merge this before our LTS compatible release.

JostBaron:
Hi Claus,

I think this is breaking, but not for callers of the method. It's breaking for implementors of the interface - they have to add the new parameter in the method implementations.

Will do the rebase later today.

Regards,
Jost

JostBaron:
Rebase is done - I couldn't test it though, because of not having a TYPO3 at 7.4 or greater, and not having time to set one up.

NamelessCoder:
Ah, right about the interface... but I'd judge that it is extremely unlikely for any Provider implementation to override the `shouldCall` method. Since we don't use deprecation markings in the subject please remove it there and add to the commit message that it is considered breaking exclusively if you override the `shouldCall` method in a custom Provider (no FluidTYPO3 extension does this so in that regard, it is safe).

JostBaron:
Updated the commit message.

NamelessCoder:
Perfect - thanks a lot Jost!

PR #967, opened 2015-10-15
[BUGFIX] Fix #966 Search pageId only when need

### LINKED ISSUE #966: plugin.xxx.forms.xxx.enabled =0 doesn't work on different root templates
If you use plugin.xxx.forms.xxx.enabled =0 with different setting in different root templates, it doesn't work, because the pageId for used typoscript isn't found right. It is a new version of #794. Patch is coming soon.

joneppie:
Only call getPageIdFromRecordUid for searching pageId when id is not a pageID itself


NamelessCoder:
Please shorten the subject line to around <100 characters as absolute max. Good ones are around 85 characters. You can continue the description of what the commit fixes in the commit message body.

deerfish:
Guten Tag
 
Ich bin vom Donnerstag, 22. Oktober 2015 bis und mit Freitag, 23. Oktober 2015 abwesend. 
 
In dringenden Fällen dürfen Sie sich bei Hugo Stalder, 041 266 0034, hugo.stalder@appline.ch melden.


Einen guten Tag wünscht

Damian Parpan



Damian Parpan
appline
Hirschengraben 52
6003 Luzern
Switzerland

T: +4141 266 0034
M: +4179 303 4878

www.appline.ch

Classes/Configuration/BackendConfigurationManager.php
```diff
@@ -110,7 +110,7 @@ protected function getPageIdFromRecordIdentifiedInEditUrlArgument() {
 		if ('pages' === $table && 'new' === $command) {
 			// if TYPO3 wants to insert a new page, URL argument is already the PID value.
 			return $id;
-		} elseif (0 <> $id) {
+		} elseif (0 <> $id && 'pages' !== $table) {
```
NamelessCoder: I suggest instead in line 127 where `'pid'` is used, to replace the straight usage of `'pid'` with: `'pages' !== $table ? 'pid' : 'uid'`. This would give us the same behavior but would be easier to understand in a year or two ;)

joneppie: That  suggestion create a database query like this one: ```SELECT uid FROM pages WHERE uid = 2```

NamelessCoder: What if line 110 instead removes the `&& 'new' === $command` check? That'd cover this case, too. There's also the consideration: what happens when `$id` is a negative value (e.g. insert after)?

joneppie: Change line 110 to ```if ('pages' === $table && 0 < $id) {``` ?

NamelessCoder:
Cherry-picked as https://github.com/FluidTYPO3/flux/commit/f9813abc9c21e22a17c235941033435813d04dd4 - thanks :)

Issue #954, opened 2015-10-07
not correctly using page_overlay values


lsascha:
i found a problem in flux using Typo3 7.5 (and probably before)

that it does not always correctly use page_overlays in languages, and on my install even displayed the language_overlay in the default language.

I was able to fix it in changing ext/flux/Classes/Provider/AbstactProvider.php
on line 468 in the method getPageValues()

i changed it to the following:
```php
protected function getPageValues() {
    $record = $GLOBALS['TSFE']->page;
    if ($GLOBALS['TSFE']->sys_language_uid != 0) {
        $localisation = $this->recordService->get('pages_language_overlay', '*', "pid = '" . $record['uid'] . "'");
    }
    if (FALSE === empty($localisation)) {
        $record = RecursiveArrayUtility::merge($record, reset($localisation));
    }
    return $record;
}
```
(added the check with if ($GLOBALS['TSFE']->sys_language_uid != 0) )
also it seems it does not take into account that page_language_overlays can be marked as "deleted" in the database. So it might select an already deleted language_overlay (that additional check is not yet in the code above)

it seems the Unit Test for AbstactProvider fails then, but since it works on my installation i am not sure if the unit test is even correct.

pixelbrackets:
Had the same issue in an installation.

Whenever a page was translated by an editor and I used the page-object, the translated version was used. Regardless of currently selected language. My page had a german default language and an english translation. The {page.title} value always showed the english title, because the current code does not check what language is selected but always loads a translation instead.

The patch above is good, but does not check which language is currently selected. In my case the page may also have a russian translation, then the method would still return english (first translation found). 
I have modified the patch and added a check for the currently used language.

I'll send a pull request.

lsascha:
i am not sure and i would have to dig a lot deeper into the code, but wasn't the correct language overlayed before already without problems? The only problem there was that the first langauge_overlay was also applied in the default language. But no apparent problem if a language was selected anyway. Would have to test it more or dig deeper into the code.

anyway, your fix does not seem to make any issues yet for me so i guess its ok.

loco8878:
you should use 

`$localisation = $GLOBALS['TSFE']->sys_page->getRecordOverlay(
				'pages',
				$record,
				$GLOBALS['TSFE']->sys_language_uid
			);`

to get the correct translations

NamelessCoder:
Fixed by #980 - thanks guys for your patience and contributions!

PR #980, opened 2015-10-30
[BUGFIX] Dont apply localisation to page values in default language

pixelbrackets:
Do not load language overlays of a page if the
currently used language language is the default language.

If the current language is not the default language,
then load the appropriate language overlay only.

Refs #954


pixelbrackets:
See https://github.com/FluidTYPO3/flux/issues/954#issuecomment-152462682 for some notes about this one

mneuhaus:
:+1:

PR #995, opened 2015-11-06
[BUGFIX] Returns the correct value

jmverges:


andreaswolf:
There’s also a patch for the root cause of this pending now in Gerrit: https://review.typo3.org/44599

PR #1003, opened 2015-11-09
[DOC] Warn about <flux:field.inline.fal> inside sections

cweiske:
See https://github.com/FluidTYPO3/fluidpages/issues/285

### DIFF
--- a/Classes/ViewHelpers/Field/Inline/FalViewHelper.php
+++ b/Classes/ViewHelpers/Field/Inline/FalViewHelper.php
@@ -30,6 +30,20 @@ use TYPO3\CMS\Fluid\Core\Rendering\RenderingContextInterface;
  *         -> v:iterator.first()
  *         -> v:variable.set(name: 'settings.files')}
  *
+ * ### Usage warning
+ *
+ * Due to [TYPO3 core bug #71239](https://forge.typo3.org/issues/71239), using
+ * FAL references within sections (`<flux:form.section>`) in content elements
+ * or within the page configuration does not work.
+ *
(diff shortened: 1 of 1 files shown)


Classes/ViewHelpers/Field/Inline/FalViewHelper.php
```diff
@@ -30,6 +30,21 @@
...
+ *
+ * Due to [TYPO3 core bug #71239](https://forge.typo3.org/issues/71239), using
+ * FAL references within sections (`<flux:form.section>`) in content elements
+ * or within the page configuration does not work.
+ *
+ * When choosing a file in one section element, you will see it in all sections.
+ * When choosing a file in a page configuration, it will be visible in the subpages
+ * configuration, too.
+ *
+ * This issue will most likely not fixed before TYPO3 8, so do not use it.
```
NamelessCoder: I think a `be` got missed - `This issue will most likely not be fixed before TYPO3 8, so do not use it.`

Classes/ViewHelpers/Field/Inline/FalViewHelper.php
```diff
@@ -30,6 +30,21 @@
...
+ *
+ * Alternatively, you could use `<flux:field.file>`.
+ *
+ *
```
NamelessCoder: One too many blank lines ;)

NamelessCoder:
I like this way of warning about the core bug - hopefully we will see a lot less people fall into this trap!

cweiske:
Changed.

PR #1005, opened 2015-11-10
[TASK] Align styling for new content button for 7LTS

jmverges:


Classes/View/PreviewView.php
```diff
@@ -1,3 +1,5 @@
+# Vagrant for BadHorns
```
NamelessCoder: Oops :)

NamelessCoder:
I think you can also drop `function isLegacyCoreVersion` as part of this one. IIRC it doesn't get used anymore now?

PR #1007, opened 2015-11-11
[FEATURE] Allow any Provider to manipulate any TCA detail

NamelessCoder:
No need to use a complicated core API - a single Provider method can now be used to manipulate everything about the TCA related to the table/type that made the provider trigger.

### DIFF
--- a/Classes/Backend/FormEngine/ProviderProcessor.php
+++ b/Classes/Backend/FormEngine/ProviderProcessor.php
@@ -1,0 +1,40 @@
+<?php
+namespace FluidTYPO3\Flux\Backend\FormEngine;
+
+use FluidTYPO3\Flux\View\ViewContext;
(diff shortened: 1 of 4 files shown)


cedricziel:
Can only be merged on a new major version as it breaks `ProviderInterface`

NamelessCoder:
I disagree: the method is new and Providers are pretty much required to also subclass the AbstractProvider which implements the method in a NOOP style. I'll add a note in the commit and changelog that it breaks edge cases of Providers that do not subclass AbstractProvider.

cedricziel:
I would wholeheartedly agree if interfaces had default methods but they dont. I'm no friend of changing contracts.

NamelessCoder:
We will raise the minor version with next release and I think we can justify this change as a minor ;)

cedricziel:
ACK

NamelessCoder:
Rebased and resolved!

PR #1014, opened 2015-11-12
[BUGFIX] Move element to first position in grid column

monosize:
This bugfix resolves the following problems:
1. drag and drop to the first position = the element was not moved.
2. moving the element with the arrow buttons to the first position = the element was disappeared from the page

### DIFF
--- a/Classes/Service/ContentService.php
+++ b/Classes/Service/ContentService.php
@@ -207,6 +207,7 @@ class ContentService implements SingletonInterface {
 				list ($parent, $column) = $this->getTargetAreaStoredInSession($relativeTo);
 				$row['tx_flux_parent'] = $parent;
 				$row['tx_flux_column'] = $column;
+				$row['sorting'] = $tceMain->getSortNumber('tt_content', 0, $row['pid']);
 			} elseif (0 <= (integer) $relativeTo && FALSE === empty($parameters[1])) {
 				list($prefix, $column, $prefix2, , , $relativePosition, $relativeUid, $area) = GeneralUtility::trimExplode('-', $parameters[1]);
 				$relativeUid = (integer) $relativeUid;
--- a/Resources/Public/js/DragDrop.js
+++ b/Resources/Public/js/DragDrop.js
@@ -154,6 +154,23 @@ define(['jquery', 'jquery-ui/sortable'], function ($) {
      */
     return function() {
         DragDrop.initialize();
+        // Workaround move with the element with the up button to the first position in grid column
+        // The 'href' of the up button href becomes the 'href' of the down button of the first element
+        $('[data-colpos="18181"]').each(
+            function(i){
+                var gridColumn = $(this);
+                var pageId = parseInt(gridColumn.find(' > [data-page]').data('page'));
+                if (pageId<0) {
+                    var downArrow = gridColumn.find('> .t3js-page-ce-sortable span.icon-actions-move-down').first().closest('a');
+                    var upArrow = gridColumn.find('> .t3js-page-ce-sortable span.icon-actions-move-up').first().closest('a');
+                    // Check if up and down arrow exists
+                    if (upArrow.length && downArrow.length) {
+                        // Set 'href' of the up button to the 'href' of the down button of the first element in the grid column
+                        upArrow.attr('href', downArrow.attr('href'));
+                    }
+                }
+            });
+
         return DragDrop;
     }();
 });


NamelessCoder:
I'm not sure about this one... re: 1) that's fixed in another commit just this evening and re: 2) that's the same issue with one addition: I don't think we should look up the first record but rather just set `$row['sorting'] = 0` which should always place it on top (and resort other content when that record is saved).

monosize:
```$row['sorting'] = 0``` makes it simpler. The look up for the first record is then unnecessary.

File: Classes/Service/ContentService.php [hunk 1]
```diff
@@ -225,8 +226,6 @@ public function moveRecord(array &$row, &$relativeTo, $parameters, DataHandler $
 			} elseif (0 < (integer) $relativeTo) {
 				// moving to first position in colPos, means that $relativeTo is the pid of the containing page
 				$row['sorting'] = $tceMain->getSortNumber('tt_content', 0, $relativeTo);
-				$row['tx_flux_parent'] = NULL;
```
NamelessCoder: These must not be removed; they were added as part of a recent bugfix. Perhaps you need a rebase?

monosize: From my side it is correct.

There is a problem, when the element is moved with the arrow button to the first position. 
If  tx_flux_parent and tx_flux column are NULL, then the element will lose the connection to the gridcolumn and t will no longer appear on the page.

NamelessCoder: But removing those lines causes the element to not be moved if you drag it from inside a nested container to the top column position.

References:

https://github.com/FluidTYPO3/flux/commit/fab20a070409112b04c38c94c8b7ef69352fa72f
https://github.com/FluidTYPO3/flux/commit/7e20161288e6a830565e288bab91e908ca196ea3
https://github.com/FluidTYPO3/flux/commit/c0864b4f02ccd7382146e6f15737bf3576d2ed0e

Enabling the first feature means you can sort elements using the arrows and face the problem you have. The second one is a patch that fixes the drag and drop. Last is the original patch that made child records invisible and unsortable in list view (but sortable using the Relations tab of the parent).

I'm pretty sure at this point that we must have these lines and somehow learn to accept the problem of list view sorting child elements using the arrow icons...

monosize:
### The up button problem:
After some tests I found out, that there is only a functionally solution with javascript.

I have changed the DragDrop.js with the simplest idea however ;-)

The "href" of the up button from the second element in the column becomes the href of the down button form the first element in the column. So we do a down movement on the first element, when you click on the up button of the second element.

```javascript
// Workaround move the element with the up button to the first position in grid column with colpos=18181
// The 'href' of the up button href becomes the 'href' of the down button of the first element
$('[data-colpos="18181"]').each(
    function(i){
        var gridColumn = $(this);
        var pageId = parseInt(gridColumn.find(' > [data-page]').data('page'));
        if (pageId<0) {
            var downArrow = gridColumn.find('> .t3js-page-ce-sortable span.icon-actions-move-down').first().closest('a');
            var upArrow = gridColumn.find('> .t3js-page-ce-sortable span.icon-actions-move-up').first().closest('a');
            // Check if up and down arrow exists
            if (upArrow.length && downArrow.length) {
                // Set 'href' of the up button to the 'href' of the down button of the first element in the grid column
                upArrow.attr('href', downArrow.attr('href'));
            }
        }
    });
```

### The drag & drop problem:
But there was another problem.  We can't set the sorting = 0, because, if you move another element to the top of a column, then this elements has also a sorting of 0. The sorting isn't automatically recalculate. So we use 
```php
$row['sorting'] = $tceMain->getSortNumber('tt_content', 0, $row['pid']);
```
to calculate a correct sorting value.

monosize:
todo: After a drag & drop there must be done a reload of the page, in order to show the correct up and down buttons.

monosize:
There could be another way in ```FluidTYPO3\Flux\View\PreviewView``` with drawGridColumn, drawRecord and drawElement, if we can overwrite the tt_contentData. Then we can remove the changes in the DragDrop.js

monosize:
In TYPO3 7.6 w/o FluidTYPO3 extensions there are no up and down buttons shown!

see here: https://github.com/TYPO3/TYPO3.CMS/blame/master/typo3/sysext/backend/Classes/View/PageLayoutView.php#L562

and here: https://github.com/TYPO3/TYPO3.CMS/commit/fd1bdd076cd62a81da12fa1f3b32c63c2a5e9ada

Only drag & drop is used.

monosize:
I close this pull request. 

Since TYPO3 7.6 the up and down buttons are hidden.

I'm working on a new solution which also respect the content permissions in Web->Page.

NamelessCoder:
@monosize Feel free to drop by IRC to discuss these with the team ;)

PR #1015, opened 2015-11-13
[FEATURE] Bring back the paste and reference buttons.

monosize:
* Works in TYPO3 7.6.
* Now in the TYPO3 7 design.

image


NamelessCoder:
Note: we really need a core feature for referencing content; Flux content references aren't fully fledged (for example, not recursive).

PR #1029, opened 2015-11-19
[BUGIFX] Merge settings declared before by other extensions

jmverges:


Classes/Controller/AbstractFluxController.php
```diff
@@ -103,9 +103,9 @@ protected function initializeSettings() {
 		$extensionKey = $this->provider->getExtensionKey($row);
 		$extensionName = ExtensionNamingUtility::getExtensionName($extensionKey);
 		$pluginName = $this->request->getPluginName();
-		$this->settings = (array) $this->configurationManager->getConfiguration(
+		$this->settings = RecursiveArrayUtility::merge($this->settings, (array) $this->configurationManager->getConfiguration(
```
NamelessCoder: I think the order of the two arrays should be reversed - now it overrides anything already existing in $this->settings if it is defined in the typoscript array. Were it opposite, people could use $this->settings to override anything - rather than only those settings that are not in the typoscript.

PR #1032, opened 2015-11-27
[BUGFIX] Initialize full $row in FlexForm hook

smichaelsen:
$row which is passed to our FlexForm hook by TYPO3 only contains the
record uid in TYPO3 master, so we have to assure to get the whole record
in order to build the proper DS

### DIFF
--- a/Classes/Backend/DynamicFlexForm.php
+++ b/Classes/Backend/DynamicFlexForm.php
@@ -9,6 +9,7 @@ namespace FluidTYPO3\Flux\Backend;
  */
 
 use FluidTYPO3\Flux\Service\FluxService;
+use TYPO3\CMS\Backend\Utility\BackendUtility;
 use TYPO3\CMS\Core\Utility\GeneralUtility;
 use TYPO3\CMS\Extbase\Object\ObjectManagerInterface;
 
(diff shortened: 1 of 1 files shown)


cweiske:
See https://review.typo3.org/44914

NamelessCoder:
After chewing on this for a few days I think this should be a bug fix for this regression in the TYPO3 core. These things used to work, but more importantly, even if this wasn't considered a regression it would still be a problem in general that when $row is either completely new or only consist of a UID, developers would have to implement hooks like this for every table containing a nested IRRE. At the very least, parent field configuration should be ensured by the core.

And if I'm right about that, it would also appear that we will not need this fix once the "real" problem is fixed - but please correct me if I'm wrong about that!

smichaelsen:
As far as I see this problem still exists and wasn't fixed on the TYPO3 side either. It's a major problem for our client. How can we proceed on this one?

mneuhaus:
Hey, i'm currently trying to reproduce and analyse this a bit in-depth, yet i can't seem to reproduce the issue. Here's what i did:

1. fetch ```smichaelsen/flux.git``` and rebased to latest flux:development
2. opened the form in the backend for a fluidcontent element to create a **new** content element

as far as i understood the issue i should now get an $row that's not what we'd expect (proper array with table fields) as an argument for getFlexFormDS_postProcessDS.

But, as far as i can see, i get a proper array like this:


If i open an existing content element i get this data as $row:



i'm currently testing with a TYPO3 7.6.2 version here. Could you outline the specific Steps to reproduce the issue?

NamelessCoder:
Not required after https://github.com/FluidTYPO3/flux/commit/6508a5fb26af1511ab5d6309d7e6ffffcc222f1d which does this  plus more for new records - but thank you so much for the contribution nonetheless!

PR #1034, opened 2015-11-29
[FEATURE] Adds renderType and format arguments to TextViewHelper to be able to render t3editor

jmverges:


Classes/Form/Field/Text.php
```diff
@@ -50,6 +60,10 @@ public function buildConfiguration() {
 		} else {
 			$configuration['defaultExtras'] = $defaultExtras;
 		}
+		if (FALSE === empty($this->getRenderType())) {
```
NamelessCoder: This one I think causes an E_NOTICE; you have to assign an actual variable with $renderType then check that with empty().

Classes/Form/Field/Text.php
```diff
@@ -36,6 +36,16 @@ class Text extends Input implements FieldInterface {
 	protected $enableRichText = FALSE;
 
 	/**
+	 * @var string
+	 */
+	protected $renderType;
```
NamelessCoder: Is there a reasonable default value we can apply here, e.g. the $renderType that results in a standard `<textarea>`?

jmverges: I really have no idea about the default $renderType. I've been taking a look in sysext and looks like there are more options now for renderType. If I understood well, looks like ¿maybe wizards going to be deprecated and renderType will sustite at least some of them? Don't know...

Classes/ViewHelpers/Field/TextViewHelper.php
```diff
@@ -27,6 +27,9 @@ public function initializeArguments() {
 		$this->registerArgument('rows', 'int', 'Number of rows in editor', FALSE, 10);
 		$this->registerArgument('defaultExtras', 'string', 'FlexForm-syntax "defaultExtras" definition, example: "richtext[*]:rte_transform[mode=ts_css]"', FALSE, '');
 		$this->registerArgument('enableRichText', 'boolean', 'Shortcut for adding value of TS plugin.tx_flux.settings.flexform.rteDefaults to "defaultExtras"', FALSE, FALSE);
+		$this->registerArgument('renderType', 'string', 'Render type', FALSE, '');
```
NamelessCoder: Last nit-pick: please describe the purpose of the arguments in their help text here - doesn't have to be the full monty, but should at least mention the most common possible values for render type and format arguments.

PR #1035, opened 2015-12-01
[FEATURE] Maintain uid and pid of standard page

orangenpresse:
I had some problems when I depend on the uid or the pid of a page after the getPageValues function.
Because now I get the uid and pid from the pages_overlay.

For example I build an menu on a Page with the vhs menu page helper. I use the uid from the page as PageUid in the view helper.
Now if I change the language, I get a completely different result because the uid from the pages_overlay is used. But the menu depends on the PageUid not the pages_overlay.

First I thought about to check if a translated page is used and then use the pid instead. But that will add more logic to the templates. Also If you need the UID of the pages_overlay you can use the fields:
_PAGES_OVERLAY
_PAGES_OVERLAY_UID

But for default behavior I only expect that the language relevant information is changed.
I also added support for the enabled fields in the function, maybe we don't need them here because it's still checked from TYPO3. But I think it adds more safety.

Please let me know if I get something wrong or I if I should improve something.

### DIFF
--- a/Classes/Backend/DynamicFlexForm.php
+++ b/Classes/Backend/DynamicFlexForm.php
@@ -100,7 +100,7 @@ class DynamicFlexForm {
 				$dataStructure[$index] = $this->patchTceformsWrapper($subStructure, $index);
 			}
 		}
-		if (isset($dataStructure['config']) && $parentIndex !== 'TCEforms') {
+		if (isset($dataStructure['config']['type']) && $parentIndex !== 'TCEforms') {
 			$dataStructure = array('TCEforms' => $dataStructure);
 		}
 		return $dataStructure;


cedricziel:
Hi @orangenpresse ! :)

Would you please squash the commits together?

Cheers! (And thank you for the contribution!)

Classes/Provider/AbstractProvider.php
```diff
@@ -464,11 +464,30 @@ protected function getCurrentValuePointerName() {
...
+				$restoreVars[] = array(
+					'uid' => $record['uid'],
+					'pid' => $record['pid']
+				);
+			}
+			$localisation = $this->recordService->get(
+				'pages_language_overlay',
+				'*',
+				'pid = "' . $record['uid'] .
+				'" AND sys_language_uid = "' . $GLOBALS['TSFE']->sys_language_uid . '"' .
```
NamelessCoder: We should have this extended query for sure - but the $restoreVars approach I'm not too sure about. It should probably not be an array-within-an-array, and I'd instead define a new $mergedRecord:

```php
$mergedRecord = RecursiveArrayUtility::merge($record, $localisation);
$mergedRecord['uid'] = $record['uid'];
$mergedRecord['pid'] = $record['pid'];
```

alternatively:

```php
unset($localisation['uid'], $localisation['pid']);
$mergedRecord = RecursiveArrayUtility::merge($record, $localisation);
```

It would read better and we probably don't expect to be adding additional properties to this backup/restore behavior.

NamelessCoder:
This should be renamed to a `[BUGFIX] Maintain uid and pid of original page`. Plus the comment in code. Other than that, looks good!

PR #1038, opened 2015-12-07
[FEATURE] Package API

NamelessCoder:
Flux Manifest and Package Concept
=================================

### Intention:

To define a JSON-based manifest for Flux, enabling packages (later including non-extensions) to gain control over which features the package provides and how it provides them. And to define an API, PHP-based, that allows the package a dynamic way to affect the features. The manifest and packages being closely linked - the manifest being used to “seed” the Package instance.

### Purpose:

The following are the main purposes for using the package/manifest feature:

* To enable configuration-free integration with Flux features based on class name convention lookup (which can be overridden).
* Provide common API on a per-package basis: allowing packages to do things like deliver custom class names Flux will use as substitutes for Flux's inner, semi-public API).
* To allow modifying parameters stored in the manifest by using TypoScript (with the traditional use cases: manipulating third-party exts, multisite usage, etc).

### Precautions:

Some features of Flux may need to be fitted with an Interface (e.g. the ViewContext, TemplatePaths, ExposedTemplateView etc. classes. To be discussed in team which ones, if any.

### Files and classes involved:

Packages can, in addition to current methods of registering provisioning, also provide:

a manifest named flux.json which resides in the package’s root directory (alongside composer.json etc.)
A single class named $vendor\$extensionName\FluxPackage which returns manifest configuration values for your package.

The FluxPackage instance can be seeded with properties of the manifest; a common FluxPackage implementation is seeded with those properties for those packages that do not provide their own FluxPackage class[1]. No other files or classes are required in order to utilise the package/manifest feature - but additional classes can be delivered via the FluxPackage instance or defined in the manifest.

### Sample manifest file including all properties currently imagined by this author:

	{
	  “package”: “MyVendorName.MyExtension”,
	  “providers”: [“Content”, “Page”],
	  “view”: {
	    “templateRootPaths”: [“path/one/“, “path/two/“],
	    “partialRootPaths”: [“path/one”],
	    "namespaces": {
	      // namespaces made available in all templates automatically, but only inside this extension
	      "myext": "MyVendor\\MyExtension\\ViewHelpers",
	      "flux": "FluidTYPO3\\Flux\\ViewHelpers"
	    }
	  },
	  “features”: {
	    “transformation”: false, // circumvent all form data transforms inside this extension
	    “preview”: false, // render no previews at all from within this extension
	    “dataHandling”: false // circumvent record processing by Providers in this extension
	  },
	  “implementations”: {
	    “view”: “MyVendorName\\MyExtension\\View\\MySpecialExposedView”, // custom implementation of ExposedTemplateView (Flux-only pattern)
	    “viewContext”: “MyVendorName\\MyExtension\\View\\MyCustomViewContext”, // allows replacing the special ViewContext (Flux-only pattern)
	    "renderingContext": "MyVendor\\MyExtension\\Rendering\\RenderingContext", // allows replacing the rendering context with a custom version
	    “form”: “MyVendor\\MyExtension\\Form\\MySpecialStandardForm”, // default class of Form instances, will be used when ViewHelpers or TypoScript create forms for the extension
	    “templatePaths”: “MyVendor\\MyExtension\\View\\TemplatePaths”, // note: corresponding TemplatePaths pattern implemented in standalone fluid; Flux’s version will subclass it in future
	    “templateVariableProvider”: “MyVendor\\MyExtension\\View\\MyCustomVariableProvider”, // note: this is in preparation for standalone fluid as dependency, will not work on 7.6
	    “viewHelperResolver”: “MyVendor\\MyExtension\\View\\MySpecialViewHelperResolver”, // in preparation for standalone fluid
	    “viewHelperInvoker”: “MyVendor\\MyExtension\\View\\MySpecialViewHelperInvoker”, // in preparation for standalone fluid
	  },
	  “compatibility”: {
	    “7.6”: {features: {transformation: false}}, // complete copy of entire manifest possible. This case: disables “transformation” feature on TYPO3 7.6 and above..
	    “8.0”: {providers: [“Content”]} // overrule available providers on TYPO3 version 8.0 and above.
	  }
	}

With the absolute minimal version being:

	{
	  “package”: “MyVendorName.MyExtension”
	}

Simply put, this flux.json becomes a manifest much like the one used by Composer but dealing specifically with Flux (and Fluid features that we can manipulate via FluxPackages once standalone Fluid arrives with the improved PHP API and open ability to replace implementations). And like Composer the manifest has a “Package” class associated with it - but again, allowing Flux- and TYPO3-specific operation. FluxPackage instances then become the public-to-internal bridge for many Flux features that currently require significant skill to utilise.

Another purpose is to allow toggling OFF individual features - in the future we might change these default states when we deprecate features; which would in effect allow some extensions to continue subscribing to deprecated features while others do not. It can also be seen as a way to optimise performance by skipping that are fairly costly steps, e.g. data transformation (which runs on every value, tries to find a corresponding Form component for it, reads transformation instruction and finally does the transform, all in a loop. Short circuiting such logic when it is determined not to be required, but being able to toggle it on later, would be ideal. These “feature flags” (a simplification bordering on misuse) can be further described as constants within Flux - giving an additional layer of warning against deprecation as well as a fast way to search for places where FluidTYPO3 uses the feature flag.

Most of the implementation-controlling attributes can be automatically detected by class name conventions as well (which is why the keys in the “implementation” object are not FQNs, rather identifiers!), for example allowing Flux to resolve an expected Form class name based on the $vendor that’s specified in this manifest.

By making the vendor name the only absolutely required setting we guarantee that class name convention based lookups can function without the additional configuration defining the expected class name. Only when the class names differ from convention, or when extension A wishes to use a certain implementation provided by extension B, for example a custom Form class name, is the additional configuration of implementations required.

Although one might argue that overriding default values of ViewHelper arguments might first of all not make a lot of sense to define in a manifest like this, second of all should be natively supported by the core… with that considered, it makes a lot of sense to integrate this in a “Package” style concept - and since there is no such concept in the core (yet) we may as well provide it in Flux. The overrides defined here are every bit as possible using “just” PHP (for example via a controller that manipulates the view’s ViewHelperResolver) but are allowed here because controllers in Flux are optional. Our ViewHelperResolver implementation then takes care of modifying these directly via ArgumentDefinitions.

And lastly: by incorporating the CompatibilityRegistry in the seeding of FluxPackages we can allow seeding with different values according to minimum TYPO3 version - without ever touching PHP. and having every root variable right in the same file as your version based overrides.

Applesauce!

### Suggested FluxPackage interface:

	interface FluxPackageInterface {
	  public function __construct($seedArrayOrExtensionKeyOrManifestPath) : void;
	  public function getVendorName() : string;
	  public function getExtensionName() : string;
	  public function getViewPaths() : TemplatePaths;
	  public function isProviderFor($controllerName) : boolean;
	  public function isFeatureEnabled($featureName) : boolean;
	  public function getImplementation($implementationName) : object;
	  public function modify(array $alternativeManifestDeclaration) : FluxPackageInterface;
	}

There’s not a whole lot to explain here, but what should be explained is:

* getViewPaths() will return an instance of TemplatePaths; this is for ease of use (such as chaining calls to read partial paths or resolve a specific template file via FluxPackage).
* Default implementations of isProviderFor() and isFeatureEnabled() will check in the arrays defined in the manifest, if not defined, then attempt to resolve a general yes/no verdict.
* getImplementation() always returns a viable instance of the correct type; or fails hard with an exception about the nature of the problem (desired class missing/unloadable, not of the correct type, etc.)
* modify() returns self simply to allow easy chaining - the original instance is also modified.

Together these functions provide a reliable API for the features that consume Provider Extensions.

### Page context awareness

Since TYPO3 uses the page context to allow TypoScript, pageTSconfig etc. to be changed per page, Flux will do the same and allow the attributes of a given FluxPackage to be manipulated directly via TypoScript and fetching via ConfigurationManager. Any request that then has a page UID as context will then automatically use the TS which applies - and to handle this, a method similar to the modify() method on Form components: accepting an array of input (structured identically to the JSON manifest) which then merges with any already loaded attribute values, overriding ones that were previously set.

This particular feature is likely to be infrequently used but, I expect, should become quite popular with multisite setups (in which each branch could customise how each Provider Extension works under that branch). It also is one more way for Flux to erase the boundaries between PHP, TS, Fluid and a handful of TYPO3 core APIs.

### Footnotes

[1] By wrapping all current ways Flux ingests these manifest properties in a common class which delivers default implementations for all features depending on detected classes/files, we can completely replace all current usages with this FluxPackage concept from day one. We would merely need to have this default FluxPackage to seed itself with Provider controller names etc. that were registered via FluidTYPO3\Flux\Core - which we can then deprecate, or keep as wrapper API for an indefinite time.

### DIFF
--- a/Classes/FluxPackage.php
+++ b/Classes/FluxPackage.php
@@ -1,0 +1,236 @@
+<?php
+namespace FluidTYPO3\Flux;
+
+/*
+ * This file is part of the FluidTYPO3/Flux project under GPLv2 or later.
+ *
+ * For the full copyright and license information, please read the
+ * LICENSE.md file that was distributed with this source code.
+ */
+
+use FluidTYPO3\Flux\Form;
+use FluidTYPO3\Flux\Package\FluxPackageInterface;
+use FluidTYPO3\Flux\Package\PackageNotFoundException;
+use FluidTYPO3\Flux\Utility\CompatibilityRegistry;
+use FluidTYPO3\Flux\Utility\ExtensionNamingUtility;
+use FluidTYPO3\Flux\Utility\RecursiveArrayUtility;
+use FluidTYPO3\Flux\View\ExposedTemplateView;
+use FluidTYPO3\Flux\View\TemplatePaths;
+use FluidTYPO3\Flux\View\ViewContext;
+use TYPO3\CMS\Core\Utility\ExtensionManagementUtility;
+use TYPO3\CMS\Core\Utility\GeneralUtility;
+use TYPO3\CMS\Fluid\Core\Rendering\RenderingContext;
+use TYPO3\CMS\Fluid\Core\ViewHelper\TemplateVariableContainer;
+
+/**
+ * Class FluxPackage
+ */
+class FluxPackage implements FluxPackageInterface {
+
+	const IMPLEMENTATION_VIEW = 'view';
+	const IMPLEMENTATION_VIEWCONTEXT = 'viewContext';
+	const IMPLEMENTATION_RENDERINGCONTEXT = 'renderingContext';
+	const IMPLEMENTATION_FORM = 'form';
+	const IMPLEMENTATION_TEMPLATEPATHS = 'templatePaths';
+	const IMPLEMENTATION_TEMPLATEVARIABLEPROVIDER = 'templateVariableProvider';
+	const IMPLEMENTATION_VIEWHELPERRESOLVER = 'viewHelperResolver';
+	const IMPLEMENTATION_VIEWHELPERINVOKER = 'viewHelperInvoker';
+
+	const FEATURE_PREVIEW = 'preview';
+	const FEATURE_TRANSFORMATION = 'transformation';
+	const FEATURE_DATAHANDLING = 'dataHandling';
+
+	/**
+	 * @var array
+	 */
+	protected $manifest = array();
+
+	/**
+	 * @var array
+	 */
+	protected $defaultImplementations = array(
+		self::IMPLEMENTATION_VIEW => ExposedTemplateView::class,
+		self::IMPLEMENTATION_VIEWCONTEXT => ViewContext::class,
+		self::IMPLEMENTATION_RENDERINGCONTEXT => RenderingContext::class,
+		self::IMPLEMENTATION_FORM => Form::class,
+		self::IMPLEMENTATION_TEMPLATEPATHS => TemplatePaths::class,
+		self::IMPLEMENTATION_TEMPLATEVARIABLEPROVIDER => TemplateVariableContainer::class,
+		self::IMPLEMENTATION_VIEWHELPERRESOLVER => NULL, // @TODO: fill when standalone fluid arrives
+		self::IMPLEMENTATION_VIEWHELPERINVOKER => NULL // @TODO: fill when standalone fluid arrives
+	);
+
+	/**
+	 * @param mixed $seedArrayOrExtensionKeyOrManifestPath
+	 * @return FluxPackageInterface
+	 */
+	public static function create($seedArrayOrExtensionKeyOrManifestPath) {
+		return new static($seedArrayOrExtensionKeyOrManifestPath);
+	}
+
+	/**
+	 * Constructor - takes an array of manifest data in the
+	 * same structure as in the manifest JSON file, or an
+	 * extension key in which case the expected manifest
+	 * is resolved using that and naming convention of the
+	 * manifest file. Or takes a full path to the manifest
+	 * file in which case the manifest is read from there.
+	 *
+	 * Note: applies CompatibilityRegistry-resolved versioned
+	 * manifest configuration values immediately.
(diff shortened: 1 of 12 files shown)


mneuhaus:
powerful stuff here, i didn't really get all of it to be honest, but i'm all in favor of putting in a api layer to prepare for better future seperation/modular use. sounds a bit complex im some places, but it's no easy task to do i guess.

jmverges:
Question, when you talk about Page context awareness are you talking also about plugin.tx_ext.settings.MyOwnPageLayout.enabled = false ?

NamelessCoder:
Page context aware means it will use and respect the TypoScript you define on pages. That is: the optional TypoScript that can override properties in the FluxPackage. It doesn't have any impact on the particular TS setting you mention - that's used only when you resolve Form instances, not FluxPackages.

mneuhaus:
looks pretty neat, lean and clean at first glance :+1: from me :)

PR #1047, opened 2015-12-21
[BUGFIX] Missing 'TCEforms' wrapping on fields in flexform datastructure

monosize:
see also my comment on 0aaa843287c175cf506c7693974a633c2e71bd58

Exception on saving changed data when using a select field with renderType "selectCheckBox". The form data for the select field is an array but must converted to a string.

Without the TCEformsthere there is no conversion and we got an exception.

### DIFF
--- a/Classes/Form/AbstractFormField.php
+++ b/Classes/Form/AbstractFormField.php
@@ -217,6 +217,9 @@ abstract class AbstractFormField extends AbstractFormComponent implements FieldI
 		if (TRUE === $this->getRequestUpdate()) {
 			$fieldStructureArray['onChange'] = 'reload';
 		}
+		// We need a TCEforms wrap
+		$fieldStructureArray = array('TCEforms' => $fieldStructureArray);
+
 		return $fieldStructureArray;
 	}
 
--- a/Tests/Unit/Form/Field/AbstractFieldTest.php
+++ b/Tests/Unit/Form/Field/AbstractFieldTest.php
@@ -171,7 +171,7 @@ abstract class AbstractFieldTest extends AbstractFormTest {
 		$instance = $this->createInstance();
 		$instance->setClearable(TRUE);
 		$result = $this->performTestBuild($instance);
-		$this->assertNotEmpty($result['config']['wizards']);
+		$this->assertNotEmpty($result['TCEforms']['config']['wizards']);
 	}
 
 	/**
--- a/Tests/Unit/Form/Field/TextTest.php
+++ b/Tests/Unit/Form/Field/TextTest.php
@@ -64,7 +64,7 @@ class TextTest extends InputTest {
 		$instance = $this->createInstance();
 		$instance->setDefaultExtras(NULL)->setEnableRichText(TRUE);
 		$result = $this->performTestBuild($instance);
-		$this->assertArrayHasKey('defaultExtras', $result['config']);
+		$this->assertArrayHasKey('defaultExtras', $result['TCEforms']['config']);
 	}
 
 	/**
@@ -75,7 +75,7 @@ class TextTest extends InputTest {
 		$instance = $this->createInstance();
 		$instance->setDefaultExtras('richtext[*]');
 		$result = $this->performTestBuild($instance);
-		$this->assertNotEmpty($result['defaultExtras']);
+		$this->assertNotEmpty($result['TCEforms']['defaultExtras']);
 	}
 
 }


NamelessCoder:
@monosize It sounds more to me like there's a problem in FormEngine. The patch you link to was created because it is impossible to create the required FormEngine components using TCA as input, when the TCA contains this extra dimension.

So what I'm thinking is we have one of the following situations, in order of likelihood:

1. Something in the component's structure or FormEngine's reading of the structure is broken or requires new configuration that we currently do not provide, which will cause it to save values as strings from that component.
2. Something in FormEngine is changed to intentionally dispatch arrays (for some particular component types). This was the case not too long ago - see https://review.typo3.org/#/c/43408/
3. Something about the execution order or point in logic means you receive an extracted record rather than a compressed one (as is legacy behavior). If you look at the history of https://review.typo3.org/#/c/43408/ it's claimed that first, this is by design and I'm thinking about it wrong when I request that the records do not get expanded or at least the expanded record is provided as a *new* property name, since it's now apparently wrong to think of this as actual database records. Second, because FormEngine is expected to change *again* and even more in this particular area.

Here's me without filter applies: I dislike the way the new FormEngine conversion was undertaken. I strongly feel much valid input from myself and others has been disregarded. And I am of the position that FormEngine is still (partially) broken in some very important areas. Because of that, I'm hesitating to implement any sort of backwards compatibilities due to the message that these parts are *still* expected to change - and because it's very hard to tell what is a regression and what is intentional changes in FormEngine that make it incompatible with existing (third-party) code. Worst case situation: we create workarounds to avoid these issues (like this wrap) and the core then removes the regression that caused it to be required in the first place. Best case: there's quite an amount of inconsistencies now in FormEngine regarding which components (rendering types) require a legacy format and which require the new format.

This situation creates frustration for users and team members alike and it generates a significant workload for us just trying to keep up with the regressions and how to work around them in ways that once they're fixed, it won't break again and require *more* work from us to fix the fixes. Sigh. Sorry, I digress.

I honestly don't know what to do about this. I don't want to merge code that works around core regressions because those really should be made in the core itself, but when I try to make patches for such regressions, they are rejected since the behavior change is now by design, code will change, someone else will fix, etc.

My greatest worry about this is that regardless what I and the team do, the result is inevitably either a truckload of things like condition based workarounds to support both regressed and non-regressed FormEngine components - or a truckload of work for myself or someone else in first studying the FormEngine sufficiently that our opinions and merge requests will matter, then actually provide those fixes. And hope it doesn't get a "someone is working on that" / "someone else will do that" response. If we don't do either, the result is user frustration.

I'd be happy to take suggestions or assistance, but I have to say I think that patching Flux back in time should be our *last* option. Right now, Flux components deliver an array that is 100% compatible with what FormEngine expects as constructor "data" argument - if FormEngine then does something unexpected with that, I have to say it's a regression and needs to be fixed in the core.

monosize:
@NamelessCoder The array just seems as if it would be compatible. After reviewing the source code of TYPO3 I found out: "All flexforms in TYPO3 7.6.2 source still used TCEforms." The TCEforms are apparently eliminated only internally for processing in the form engine but not for fields with type "flex"!!! 

Background: In other places in the source code of TYPO3 the datastructure is generated with ```BackendUtility::getFlexFormDS```wich call ```FluidTYPO3\Flux\Backend\DynamicFlexForm::getFlexFormDS_postProcessDS``` and delivers DS without TCEforms. And there is then expected TCEforms. 

**A field of type "flex" loads additionally the DS**

Look here:  ```\TYPO3\CMS\Core\DataHandling\DataHandler::checkValueForFlex``` using ```BackendUtility::getFlexFormDS``` internally and is called during the form engine process and checked ```['TCEforms']['config']```--> And that results into the error!

NamelessCoder:
Sorry for hijacking this one @monosize but I thought it'd be quicker this way. See the commit for a bit more detail - and thanks for doing the digging!

monosize:
@NamelessCoder Isn't it better to wrap the fields with 'TCEforms' in ```Classes/Form/AbstractFormField.php```. So we have no problems with naming conflicts. see #1061

NamelessCoder:
@monosize The general idea is to have the components generate TCA only - then transform that to whichever implementation is wanted. Part of the idea is described in https://github.com/FluidTYPO3/flux_capacitor.

The decision stands between always adding TCEforms wrapping from the component itself and then stripping that from every other implementation that works in TCA - or to only add the wrapper when it is required, e.g. when the implementation is a FlexForm. I have opted for the second one.

There's also a tangent to this in that FormEngine may remove this wrapper at some point. Currently there are more than a few objects which exist in two versions; one which works in TCA and one which is nearly identical but only works in FlexForms. Ideally, those would be one component each which works in both places.

PR #1048, opened 2015-12-26
[BUGFIX] Rename FluxService injection and variable

xf-:
Same function, other argument type (includes extended) causes a exception

If we name the fluxService fluxService and not configurationService it should work in PHP7.

### DIFF
--- a/Classes/Provider/AbstractProvider.php
+++ b/Classes/Provider/AbstractProvider.php
@@ -138,7 +138,7 @@ class AbstractProvider implements ProviderInterface {
 	/**
 	 * @var FluxService
 	 */
-	protected $configurationService;
+	protected $fluxService;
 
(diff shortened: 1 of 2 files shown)


kersten:
I still got a problem with ```FluidTYPO3\Flux\Controller\AbstractFluxController``` I renamed it like you but then I get an error like this: https://travis-ci.org/FluidTYPO3/fluidtypo3-testing/jobs/73767004

```Could not analyse class: "FluidTYPO3\Flux\ViewHelpers\Field\CustomViewHelper" maybe not loaded or no autoloader?```

NamelessCoder:
This change is no longer required. We've not abandoned renaming the injected properties - it's just not something we want to be doing without first deprecating it.

PR #1051, opened 2015-12-28
[BUGFIX] Update ListViewHelper.php

kersten:
Hi,

I get the following error after I try to open the Page Configuration:

``` PHP Runtime Deprecation Notice: Non-static method FluidTYPO3\Flux\ViewHelpers\Wizard\ListViewHelper::getComponent() should not be called statically in /var/www/webroot/typo3conf/ext/flux/Classes/ViewHelpers/AbstractFormViewHelper.php line 46```

I don't know if this is the correct fix, but every other viewhelper ```getComponent``` is static too

### DIFF
--- a/Classes/ViewHelpers/Wizard/ListViewHelper.php
+++ b/Classes/ViewHelpers/Wizard/ListViewHelper.php
@@ -33,7 +33,7 @@ class ListViewHelper extends AbstractWizardViewHelper {
 	 * @param array $arguments
 	 * @return ListWizard
 	 */
-	public function getComponent(RenderingContextInterface $renderingContext, array $arguments) {
+	public static function getComponent(RenderingContextInterface $renderingContext, array $arguments) {
 		/** @var ListWizard $component */
 		$component = static::getPreparedComponent('ListWizard', $renderingContext, $arguments);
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by cedricziel** [comment]
That's no error, it's a deprecation notice. Your exception handler is not correctly configured. You need to set a higher bound for fatal errors in the install tool if it fatals your page modue.

**Comment by cedricziel** [comment]
Seems to be a legit fix though. Please amend, add the `[BUGFIX]` prefix and force-push to meet the CGL requirements.

**Comment by kersten** [comment]
@cedricziel Yes its not an error. I recently updated to PHP7 and the standard was showing me that. Sou it would be better to fix though, right?

I try to force push this.

**Comment by cedricziel** [comment]
Not the title of the pull request, but the title of the commit :)

Let's sharpen your git skills:

```sh
# in the source directory on your local box
git checkout patch-1
# this opens an editor. Add the prefix to the commit message and save & close
git commit --amend
# origin is the remote of your personal fork here
git push -f origin patch-1
```

Thanks in advance!

kersten:
Ok, thanks for teaching ;) I hope I got things right now.

cedricziel:
Very good - thank you :)

PR #1061, opened 2016-01-08
[BUGFIX] Data Structure ERROR: No ['ROOT']['el'] element found in ...

monosize:
flexform definition.

Only wrapping with 'TCEforms' if inside a 'config' definition exists a 'type' definition.
Prevend wrapping with 'TCEforms' if  'config' is used as fieldname or sheetname.

### DIFF
--- a/Classes/Backend/DynamicFlexForm.php
+++ b/Classes/Backend/DynamicFlexForm.php
@@ -100,7 +100,7 @@ class DynamicFlexForm {
 				$dataStructure[$index] = $this->patchTceformsWrapper($subStructure, $index);
 			}
 		}
-		if (isset($dataStructure['config']) && $parentIndex !== 'TCEforms') {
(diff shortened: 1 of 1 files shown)


Classes/Backend/DynamicFlexForm.php
```diff
@@ -100,7 +100,7 @@ protected function patchTceformsWrapper(array $dataStructure, $parentIndex = NUL
 				$dataStructure[$index] = $this->patchTceformsWrapper($subStructure, $index);
 			}
 		}
-		if (isset($dataStructure['config']) && $parentIndex !== 'TCEforms') {
+		if (isset($dataStructure['config']) && isset($dataStructure['config']['type']) && $parentIndex !== 'TCEforms') {
```
NamelessCoder: The first `isset($dataStructure['config'])` can then be removed since `isset($dataStructure['config']['type'])` will also safely return FALSE if `$dataStructure['config']` does not exist.

PR #1074, opened 2016-02-12
[BUGFIX] workspaces/nested elements - resolved the problem with incorrect pid …

mjankiewicz:
ENV:
- TYPO3 6.2.17
- FLUX 7.2.3
- FLUIDPAGES 3.3.1
- FLUIDCONTENT 4.3.3

I have the nested structure of elements workspaces_nested_el_1. When I try add new elements (for ex: text + image) I get blank page workspaces_nested_el_2. Both elements were created in workspace, so I have 2 records: workspaces_nested_el_3

Wrapper: http://pastebin.com/EAN20M3b
Columns: http://pastebin.com/r1UXaS5Z

The problem is with access to page where I want to add new elements. Exactly here typo3/sysext/backend/Classes/Controller/ContentElement/NewContentElementController.php: 140 / 

if ($this->id && $this->access) {
 //...
}

I found that the source my problems could be link to "create new element". Flux generates wrong links: workspaces_nested_el_4 and TYPO3 can't check correct access to page (in my case with pid = 6).

Links are generated by 2 methods: getNewLink and getNewLinkLegacy: typo3conf/ext/flux/Classes/View/PreviewView.php - executed by renderGrid(..) and next methods. Flux retrieves the correct record but doesn't check it if the record is in workspace. 
I've added additional condition which checks if the record is in workspace mode (if is, returns the base record).

What do you think ? Have anybody had similar problems with workspaces and nested elements ?

### DIFF
--- a/Classes/View/PreviewView.php
+++ b/Classes/View/PreviewView.php
@@ -250,6 +250,10 @@ class PreviewView {
 		$content = '';
 		if (TRUE === $grid->hasChildren()) {
 			$workspaceVersionOfRow = $this->workspacesAwareRecordService->getSingle('tt_content', '*', $row['uid']);
+			if ($workspaceVersionOfRow['pid'] == -1 && $workspaceVersionOfRow['t3ver_oid']) {
+				$orgRecord = BackendUtility::getRecord('tt_content', $workspaceVersionOfRow['t3ver_oid'], '*', '', FALSE);
+				$workspaceVersionOfRow['pid'] = $orgRecord['pid'];
+			}
 			$content = $this->drawGrid($workspaceVersionOfRow, $grid, $form);
 
 			$options = $this->getPreviewOptions($form);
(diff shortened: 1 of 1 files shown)


benjaminrau:
@NamelessCoder Maybe your time allows to have a look at this and maybe you can tell us if there are other known issues with workspaces and nested content areas. Thanks a lot in advance.

mjankiewicz:
The problem can be with incorrect value tx_flux_parent
incorrect_tx_flux_parent

NamelessCoder:
Hi guys - thanks for your work on this! I have manually picked and pushed and made some small adjustments (cosmetics). I'm not up to date on which issues still exist with workspaces (and versions) but would really appreciate if you could go through our open issues and see which ones you maybe can solve ;)

...and it is *perfectly* okay to do it in smaller steps like this!

PR #1075, opened 2016-02-15
[TASK] Check invalid characters in names

tomahi:
Please see the attached file. When flux columns contain hyphens or pipes in their name you get problems when moving content elements in the backend. E.g. if you paste an element to "column-2" it gets saved as "column" in the database. The reason is that the parameters for this action are seperated with hyphens and pipes. One possible solution would be to urldecode and encode the parameters. Another would be to give the developer at least a hint, so that he can correct this. I implemented a check for hyphens and pipes in the name and if so an exception will be thrown in frontend.

[ColThree.txt](https://github.com/FluidTYPO3/flux/files/130292/ColThree.txt)

### DIFF
--- a/Classes/Form/AbstractFormComponent.php
+++ b/Classes/Form/AbstractFormComponent.php
@@ -24,6 +24,7 @@ use TYPO3\CMS\Extbase\Object\ObjectManagerInterface;
 use TYPO3\CMS\Extbase\Reflection\ObjectAccess;
 use TYPO3\CMS\Extbase\Utility\LocalizationUtility;
 
+
 /**
  * AbstractFormComponent
  */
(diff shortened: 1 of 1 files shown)


NamelessCoder:
Thanks for your pull request, Steffen. Although the use case is valid, we should not restrict the names globally in this way - not all components cause trouble if using an invalid name. The ones that do cause issues mainly cause issues because of the current DOM formatting which there is a plan to change; from exploding a special DOM ID attribute to storing such configuration in a `data-tca` or similar property. It's still a semi-nebulous plan but there *is* a plan.

That being said we would face issues on LTS with the DOM ID, but only for a specific list of components:

* Forms
* Sheets
* Content areas

Possibly also in:

* Section
* Object (when inside section)

We currently dispatch a log warning in `Form::setId` since an invalid name here could potentially break TCA completely and could do the same in the `setName` method of those specific components. They would have to implement a `setName` method since that's currently inherited, but that would be okay. It must then also log this issue in both FE and BE mode (e.g. no check for `TYPO3_MODE`).

But what we don't want is an exception thrown at users at a global level. We only want this to be a "notice" type messaging and only in the components that cause trouble with these DOM ID parsing cases. A pull request implementing that in the mentioned methods would be appreciated! Coding style looks spot on - except for the merge commit. If you do the same style and avoid the merge commit in the next pull request it's an easy merge ;)

PR #1076, opened 2016-02-15
TASK] Change expected exception message from NULL to empty string

mjankiewicz:
ENV:
- PHPUnit 5.2.5
- PHP 5.6.11

I have 23 errors when run phpunit on clean repo: "FAILURES!
Tests: 1702, Assertions: 3653, Errors: 23." All errors refers to incorrect expected value.


NamelessCoder:
Please amend commit message; should read `[TASK] Change expected exception message from NULL to empty string`.

NamelessCoder:
Sorry to hijack this one - replaced by https://github.com/FluidTYPO3/flux/pull/1079 since I needed to do some other updates for tests as well.

For future cases when we ask to amend the commit message, what we're really asking is to do `git commit --amend` locally then adjust the commit message, and then `git push --force` the result to your pull request branch (I see you're already using one of those which is great!). Sadly the "edit title" button doesn't update the commit even though it could do so when there's just a single commit, imho (but that's a story for a Github feature request). And never be afraid to ask for help ;)

PR #1091, opened 2016-03-03
Update TemplatePaths.php

parser11:
Core: Error handler (FE): PHP Warning: rtrim() expects parameter 1 to be string, array given in /home/www/doc/9007/*/typo3conf/ext/flux/Classes/View/TemplatePaths.php line 308

$paths[self::CONFIG_LAYOUTROOTPATH] is an array and not correct pushed on $layoutRootPaths.

### DIFF
--- a/Classes/View/TemplatePaths.php
+++ b/Classes/View/TemplatePaths.php
@@ -344,14 +344,14 @@ class TemplatePaths {
 			}
 		}
 		// last appended if set: the legacy singular paths configuration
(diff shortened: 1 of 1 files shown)


NamelessCoder:
We need to remove this legacy support instead ;)

NamelessCoder:
Supports removed in https://github.com/FluidTYPO3/flux/commit/843f52fbea4bd8b290ddb4b7aae19985a2976fb3 - thanks for your contribution!

PR #1100, opened 2016-03-14
Wrong order of parameters in compile function

tpinne:
The code returned from the compile function had a wrong order of parameters.


cedricziel:
I propose to deprecate the viewhelper in favor of using vhs.

NamelessCoder:
I agree, we can deprecate this one - originally it was there because we needed a way to access variables using dynamic names in the preview template. Since that's no longer a Fluid template we don't have to include this ViewHelper, nor do we have to set VHS as a dependency for Flux.

But we do need this issue fixed (for those users that decided to use this one, which shouldn't be many). The patch itself is good but the commit message needs a `[BUGFIX]` prefix in order to be accepted.

bjo3rnf:
Manually picked. Thanks for contributing.

PR #1122, opened 2016-04-20
[BUGFIX] Fix getTranslateCsvItems for Select field

x3ro:
This broke since the generated language path was no longer correct (`$this->name` vs. `$this->getPath()`), as well as the change in `resolveLocalLanguageValueOfLabel` to no longer actually translate the resolved identifiers. Previously, it was checked whether the `$resolvedLabel` was translated or not, and if not, it fell back to using the `$itemName`. Since `resolveLocalLanguageValueOfLabel` now always returns a string starting w/ `LLL:`, this check has been dropped and the label will be empty if no translation is found.

### DIFF
--- a/Classes/Form/Field/Select.php
+++ b/Classes/Form/Field/Select.php
@@ -116,10 +116,7 @@ class Select extends AbstractMultiValueFormField {
 				}
 			} else {
 				foreach ($itemNames as $itemName) {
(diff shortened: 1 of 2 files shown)


NamelessCoder:
Hey Lucas - please rebase or merge development into this branch and I think we're ready to go! :)

x3ro:
Done

PR #1123, opened 2016-04-20
[BUGFIX] The label names are generated wrong.

monosize:
Remove 'form' from the label name, otherwise existing labels are no longer recognized.

see comments in https://github.com/FluidTYPO3/flux/commit/41cb134fe6a2ce2bac05fd04152ab5c392ceaba9


NamelessCoder:
Thanks!

PR #1126, opened 2016-04-22
[DOC] Better document Form\Option\GroupViewHelper

x3ro:


NamelessCoder:
Hi Lucas,

The `group` option is a generic option and relates to any extension, not just fluidcontent. We avoid documenting behaviors of fluidcontent and other extensions as part of flux for that reason. The same is true about the `sorting` and `icon` options.

Extending the documentation is perfectly fine but it should stay generic. In this case something like "Groups Form instances into logical groups that can be used by extensions using Flux Forms. See documentation of the extension you are writing a template for to know if and how groups are supported".

x3ro:
I see, but that does make relevant documentation rather hard to discover. I basically found the information in question by going through the code. I'm having a hard time figuring out where I would put this documentation instead.

NamelessCoder:
> I'm having a hard time figuring out where I would put this documentation instead.

We do, too, so you're not alone ;)

One of the problems is that we still don't have properly segmented documentation that explains Flux Forms separately which we could link to from a much shorter "Flux Forms in Fluidcontent" documentation. That's where I would put this.

Our challenge is and has been for years to find a docs structure that is good for humans *and* lets us avoid repeating ourselves and spreading for example Form usage instructions into multiple chapters. It's hard because the features of Flux are almost always best illustrated by either flluidcontent or fluidpages, but documenting Flux from within *those* carries a lot of problems concerning which version of Flux a given documentation relates to, since some major versions of fluidcontent and fluidpages support multiple versions of Flux with differences in available features.

We'd be very happy for assistance in that area but know from experience that it is definitely not a small task - and is one that requires knowing a lot about the internal stuff as well as the user API stuff. We don't have a budget so we can't pay for that, but I always offer that if anyone wants to assist with that then I don't mind teaching what needs to be known. It's easy enough to explain to expert users - but very hard (at least for me) to explain to someone just starting out with TYPO3 and FluidTYPO3. We like to think FluidTYPO3 is easier for beginners than equivalent with bare TYPO3 but docs is definitely our biggest problem for attracting and guiding our beginners. Indeed, things could be a lot better.

If you want to jump in on that part of the project you would be most welcome. Please come see us on IRC / the support chat on the site if you do!

NamelessCoder:
Picked and adjusted as https://github.com/FluidTYPO3/flux/commit/8f6aef65b4ad1f7f8fc37e87d17039ce4a218fc1

PR #1131, opened 2016-04-28
[DOC] flux:form.option.icon documentation

cweiske:


Classes/ViewHelpers/Form/Option/IconViewHelper.php
```diff
@@ -12,7 +12,20 @@
 use FluidTYPO3\Flux\ViewHelpers\Form\OptionViewHelper;
 
 /**
- * Form icon option ViewHelper
+ * ### Content element icon
```
NamelessCoder: This is actually also the page template icon, TCA icon, backend module icon, etc. - it's a generic setting so we should avoid documenting this way how that generic setting gets used in other extensions. There's a similar issue described in https://github.com/FluidTYPO3/flux/pull/1126 with some background.

cweiske: I know, but I think it's better have partial documentation than none.

NamelessCoder:
Picked and adjusted as https://github.com/FluidTYPO3/flux/commit/eeaa725359945b4d951c002005581a83d2c4cf84

PR #1136, opened 2016-04-29
[BUGFIX] Wrong positioning during dragging since TYPO3 7.6.6

monosize:
Wrong distance of the element to the mouse pointer while dragging.

artboard 1

artboard 2

### DIFF
--- a/Resources/Public/js/DragDrop.js
+++ b/Resources/Public/js/DragDrop.js
@@ -16,7 +16,13 @@ requirejs.undef("TYPO3/CMS/Backend/LayoutModule/DragDrop");
  * based on jQuery UI
  */
 define(['jquery', 'jquery-ui/sortable'], function ($) {
+    'use strict';
 
+    /**
+     *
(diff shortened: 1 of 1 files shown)


Resources/Public/js/DragDrop.js
```diff
@@ -33,6 +39,9 @@ define(['jquery', 'jquery-ui/sortable'], function ($) {
      * initializes Drag+Drop for all content elements on the page
      */
     DragDrop.initialize = function() {
+        // setting the style of all divs with data-language-uid to 'postion:relative' in order to prevend during dragging a wrong positioning
```
NamelessCoder: Should the last part of this comment read `in order to prevent dragging to a wrong positioning`? Looks like a word got lost or there's an extra one. And mind the typo in `prevent` :)

Overall: we don't really need the parameter annotations in JS but of course they don't hurt. We see this D&D JS patch as hopefully temporary so as long as it works, that's Good Enough (TM) for us ;)

NamelessCoder:
Manually picked - sorry about the delay.

PR #1138, opened 2016-05-03
[BUGFIX] remove disfunctional shortcut icon

franzkugelmann:
For content elements with CType='shortcut' flux tries to render an extra icon with link to edit the referenced element. The link however does not work is not necessary anymore, because TYPO3 already places an icon with context menu there, where you can jump to edit the referenced element.


NamelessCoder:
Manually picked, commit message adjusted and corresponding unit test removed - https://github.com/FluidTYPO3/flux/commit/c51ca738a054d51f4b7fe75a38a39ee321f5d681. Thanks Franz! :)

PR #1146, opened 2016-05-25
Fix rendering of target content's title and uid

arsrobota:
In createShortcutIcon(), I shortened value of $row['records'] using substr() as this value was formed like 'tt_conten_[uid]'.


cedricziel:
Hi Alexis,

thank you for the contribution! Please correct the following issues first:

1. Please drop c0cd177ac68c2d68cee4b7889b505945a93fe57c from the pull request. `ext_conf_template.txt` provides default a constants-editor-like editing experience for the most basic settings through the extension manager.
2. Please amend f5133c4 and use the prefix `[BUGFIX] ` then force push.
3. Please check travis. Seems you broke the build

Cheers,
Cedric

NamelessCoder:
Hi @arsrobota - we lean towards completely removing this shortcut icon, and leaving it up to the TYPO3 core to decide if such an icon should at some point be added. If you want, you can create an icon or link or button (any HTML goes) in your HTML with a manually constructed URL that points to the target in just the way you like; to edit, view in page module etc.

NamelessCoder:
Shortcut icon removed in https://github.com/FluidTYPO3/flux/commit/c51ca738a054d51f4b7fe75a38a39ee321f5d681 - core provides icon with context menu that has "jump to" option.

PR #1158, opened 2016-06-23
Added boolval() to dataType transformation

geldmacher:
This is very handy if you like to build up an json configuration object with {v:format.json.encode()}.

Example:
```
<flux:field.checkbox name="slick.autoplay" label="Autoplay" default="0" transform="bool" />
{v:format.json.encode(value:'{settings.slick}')}
```


NamelessCoder:
Merged regardless of code style warnings (we are converting to PSR-2 at this time). Adjusted code examples to avoid objectaccessor-in-textnode (see https://vimeo.com/167666466).

PR #976, opened 2015-10-27
Bugfixes for pasting elements with translations

BenjaminBeck:
Please do not merge - This is still work in progress.

Any comments, help or tips are welcome..

Related to #974

PR #1163, opened 2016-07-19
Fixing & handling issues caused by wrong handling of tx_flux_parent when creating translations

### LINKED ISSUE #1125: again: colPos=18181 after moving a translated content-element
TYPO3: 6.2.21
FLUX: 7.2.3

Pages-modul, Columns-view of all languages:
If I move a translated content-element around (arrow up  or arrow down), the content-element disappears, because it's colPos is set to 18181 .

dneuge:
The issue with content disappearing due to being assigned `colPos=18181` by error (#1125) was a subsequent fault caused by assigning a random `tx_flux_parent` reference when creating translations for elements not being assigned to a column of a container FCE.

This pull request contains a number of changes adressing the root cause and (hopefully all) effects:
- `ContentService::initializeRecordByNewAndOldAndLanguageUids` now checks for `tx_flux_parent` being 0 before using it as `t3_origuid` (which resulted in random IDs being used)
  - We added a few conditions to the query (`CType`, `pid`, `sys_language_uid`) for better result sanity and to remove the unnecessary foreach-loop (why loop if we are querying the database anyway).
  - `$translatedParents` has been renamed to `$copiedParents`, see below
- When moving or otherwise editing content elements `tx_flux_column` is now mandatory as well as `tx_flux_parent` (`ContentService::moveRecord` and `ContentService::affectRecordByRequestParameters`). This prevents records having been translated prior to the above fix to be moved to `colPos=18181` when there is no container column to show them in (just what was described in #1125).
- The update script now removes all `tx_flux_parent` where `tx_flux_column` has not been set, thus removing the erroneous parent reference caused by the translation bug.
- As content may already have disappeared from backend (and frontend!) the update script now checks for any (non-deleted) records which are assigned to `colPos=18181` without `tx_flux_column` (the translation bug) or `tx_flux_parent` (which will have been removed by the update script to clean up the effects of that bug). If any such content is found, the user now sees a listing of all inaccessible records with the option to edit them. Instructions on how to resolve these issues are included with that list.

As we don't use workspaces (never worked for us except on the most basic websites...) we didn't test if any of these changes will interfere with workspace handling. However, it seems unlikely as we didn't change anything in regard to workspaces.

While working on these fixes I found a few more issues which need separate handling (out of scope for this pull request):
- `ContentService::initializeRecordByNewAndOldAndLanguageUids` searches for copies using `t3_origuid` when looking for a container to register translated elements on in the target language. It uses the first non-deleted container it can find. This does not seem to be the correct way for a few reasons:
  - Copies are just that: Copies. There is no actual link to their original and thus the copied container may not be the correct one to insert the translation to. IMHO content "translated in copy mode" should not attempt to register to any container at all. However, this appears to have been a commonly requested feature in the past, so this may not be simply removed. I could imagine providing an option to disable that behaviour instead.
  - Real translations (linked by `l18n_parent`) are only found because they also set `t3_origuid`. It would be better to search for a real translation first and only continue searching for copies if none could be found.
- Furthermore, when creating translations:
  - There seem to be no checks if `tx_flux_column` actually exists for the `tx_flux_parent` to be assigned. This may lead to another issue of disappearing content (not checked).
  - There seem to be no check for the found `tx_flux_parent` to be the same FCE as the translation original. This as well may lead to disappearing content.
- In general, content may disappear from BE if inaccessible/wrong `tx_flux_parent`, `tx_flux_column`  FCEs or CTypes are being used. Such issues could be caused by backend editors. Since proper handling for this issue introduces a way to find and list elements affected by #1125 it may make sense to add more checks for other cases of disappearing content and include them as well to that list (together with instructions/explanations).

Pulling resolves #1125 and maybe other issues as well (check #1125 for referenced issues).

Pull request #976 will have to be revised as it already attempted to fix the translation bug and thus conflicts with our commit. Our fix seems to be more complete, however, and should be deployable without any ill side-effects, I hope (please review the changes, just in case, thanks :) ).

### DIFF
--- a/Classes/Service/ContentService.php
+++ b/Classes/Service/ContentService.php
@@ -345,24 +345,34 @@ class ContentService implements SingletonInterface
                 // look for the translated version of the parent record indicated
                 // in this new, translated record. Below, we adjust the parent UID
                 // so it has the UID of the translated parent if one exists.
-                $translatedParents = (array) $this->workspacesAwareRecordService->get(
-                    'tt_content',
-                    'uid,sys_language_uid',
-                    "t3_origuid = '" . $oldRecord['tx_flux_parent'] . "'" . BackendUtility::deleteClause('tt_content')
-                );
-                foreach ($translatedParents as $translatedParent) {
-                    if ($translatedParent['sys_language_uid'] == $newLanguageUid) {
-                        // set $translatedParent to the right language ($newLanguageUid):
-                        break;
+                unset($copiedParent);
+                if (0 < (integer) $oldRecord['tx_flux_parent']) {
+                    // Search for a container element which has been copied from
+                    // the container referred to by the original element
+                    // (oldRecord) and exists in the target language.
+                    // Note that this cannot guarantee the correct container
+                    // to be found and used (using copies for translations
+                    // leads to ambiguous results by principle).
+                    // This is the old method which used to implement a fix for
+                    // #816 and #796.
+                    $copiedParents = (array) $this->workspacesAwareRecordService->get(
+                        'tt_content',
+                        'uid,sys_language_uid',
+                        "t3_origuid = '" . (integer) $oldRecord['tx_flux_parent'] . "' AND sys_language_uid = '" . (integer) $newLanguageUid . "' " . BackendUtility::deleteClause('tt_content')
+                    );
+
+                    if (count($copiedParents) > 0) {
+                        $copiedParent = $copiedParents[0];
                     }
-                    unset($translatedParent);
                 }
+
                 $sortbyFieldName = true === isset($GLOBALS['TCA']['tt_content']['ctrl']['sortby']) ?
                     $GLOBALS['TCA']['tt_content']['ctrl']['sortby'] : 'sorting';
                 $overrideValues = [
                     $sortbyFieldName => $tceMain->resorting('tt_content', $row['pid'], $sortbyFieldName, $oldUid),
-                    'tx_flux_parent' => $translatedParent ? $translatedParent['uid'] : $oldRecord['tx_flux_parent']
+                    'tx_flux_parent' => $copiedParent ? $copiedParent['uid'] : $oldRecord['tx_flux_parent']
                 ];
+
                 $this->updateRecordInDatabase($overrideValues, $newUid);
             }
         }
--- a/Classes/Service/ContentService.php
+++ b/Classes/Service/ContentService.php
@@ -64,7 +64,7 @@ class ContentService implements SingletonInterface
         unset($id, $tceMain);
         if (false === empty($parameters['overrideVals']['tt_content']['tx_flux_parent'])) {
             $row['tx_flux_parent'] = (integer) $parameters['overrideVals']['tt_content']['tx_flux_parent'];
-            if (0 < $row['tx_flux_parent']) {
+            if (0 < $row['tx_flux_parent'] && false === empty($row['tx_flux_column'])) {
                 $row['colPos'] = self::COLPOS_FLUXCONTENT;
             }
         }
@@ -269,7 +269,7 @@ class ContentService implements SingletonInterface
             $row['tx_flux_parent'] = null;
             $row['tx_flux_column'] = null;
         }
-        if (0 < $row['tx_flux_parent']) {
+        if (0 < $row['tx_flux_parent'] && false === empty($row['tx_flux_column'])) {
             $row['colPos'] = self::COLPOS_FLUXCONTENT;
         }
         $this->updateRecordInDatabase($row);
--- a/Classes/Service/ContentService.php
+++ b/Classes/Service/ContentService.php
@@ -358,7 +358,7 @@ class ContentService implements SingletonInterface
                     $copiedParents = (array) $this->workspacesAwareRecordService->get(
                         'tt_content',
                         'uid,sys_language_uid',
-                        "t3_origuid = '" . (integer) $oldRecord['tx_flux_parent'] . "' AND sys_language_uid = '" . (integer) $newLanguageUid . "' " . BackendUtility::deleteClause('tt_content')
+                        "t3_origuid = '" . (integer) $oldRecord['tx_flux_parent'] . "' AND pid = '" . (integer) $row['pid'] . "' AND CType = 'fluidcontent_content' AND sys_language_uid = '" . (integer) $newLanguageUid . "' " . BackendUtility::deleteClause('tt_content')
                     );
 
                     if (count($copiedParents) > 0) {
--- a/Classes/Service/ContentService.php
+++ b/Classes/Service/ContentService.php
@@ -358,7 +358,11 @@ class ContentService implements SingletonInterface
                     $copiedParents = (array) $this->workspacesAwareRecordService->get(
                         'tt_content',
                         'uid,sys_language_uid',
-                        "t3_origuid = '" . (integer) $oldRecord['tx_flux_parent'] . "' AND pid = '" . (integer) $row['pid'] . "' AND CType = 'fluidcontent_content' AND sys_language_uid = '" . (integer) $newLanguageUid . "' " . BackendUtility::deleteClause('tt_content')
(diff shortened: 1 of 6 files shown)


Tests/Unit/Service/ContentServiceTest.php
```diff
@@ -315,6 +345,61 @@ public function getLanguageInitializationTestValues()
...
+    }
+
+    /**
+     * @return array all expected results from database which are equivalent to 0 when casted to integers
+     */
+    public function getIntegerEquivalentsOfZero() {
+        return [
+            [0],
+            [''],
+            [NULL]
```
NamelessCoder: I think we may need a test for `'0'` here (if equivalent to zero is what PHP would consider zero on `==`).

Tests/Unit/Service/ContentServiceTest.php
```diff
@@ -393,4 +478,92 @@ public function moveRecordWithPositiveRelativeToLoadsRelativeCopiesValuesSetsCol
...
+        $colPosFluxContent = 18181;
+
+        return [
+            // cases where colPos may change due to relative record using different colPos (starting with colPos 2, colPos 1 on relative record, neither flux parent or column name is set)
+            [['colPos' => 2], 0 - MiscellaneousUtility::UNIQUE_INTEGER_OVERHEAD - 1, [], [0, ''], ['', ''], 2], // using session
+            [['colPos' => 2], 123, ['tx_flux_column' => '', 'tx_flux_parent' => 0, 'colPos' => 1], [], ['', 'colpos-1-page-unused-unused-top-0-'], 1], // using parameters
+            [['colPos' => 2], -23, ['tx_flux_column' => '', 'tx_flux_parent' => 0, 'colPos' => 1], [], ['', ''], 1], // "inserting new element after"
+            [['colPos' => 2], 123, ['tx_flux_column' => '', 'tx_flux_parent' => 0, 'colPos' => 1], [], ['', ''], 2], // "first position in colPos"
+            [['colPos' => 2], 0, ['tx_flux_column' => '', 'tx_flux_parent' => 0, 'colPos' => 1], [0, ''], ['', ''], 2], // unmatched inner cases
+            [['colPos' => 2], 'x123', ['tx_flux_column' => '', 'tx_flux_parent' => 0, 'colPos' => 1], [], ['', ''], 2], // not changing for gridelements
```
NamelessCoder: Please remove reference to gridelements (will show up in code searches) and rename the test set to something like `not changing on unrecognized colPos`. Small note: if you use the test set description as key, it will show up if/when the test fails with that specific test set - which would be very nice since you added all these gorgeous test cases!

NamelessCoder:
Looks pretty damn good so far. Just the few notes from me at this time (will do a proper review and test in the coming days).

Classes/Service/ContentService.php
```diff
@@ -345,24 +345,38 @@ protected function initializeRecordByNewAndOldAndLanguageUids(
...
+                    // to be found and used (using copies for translations
+                    // leads to ambiguous results by principle).
+                    // This is the old method which used to implement a fix for
+                    // #816 and #796.
+                    $copiedParents = (array) $this->workspacesAwareRecordService->get(
+                        'tt_content',
+                        'uid,sys_language_uid',
+                        "t3_origuid = '" . (integer) $oldRecord['tx_flux_parent'] . "' AND " .
+                            "pid = '" . (integer) $row['pid'] . "' AND " .
+                            "CType = 'fluidcontent_content' AND " .
```
dneuge: Hm, maybe I'm confusing things but... While the content that is being translated may be of any `CType`, this is about searching for the container to put it into. Since that container's `uid` needs to go into `tx_flux_parent`. I thought the only `$copiedParents` that make any sense are `fluidcontent` (= FCEs, hopefully ones that actually have a container column of the matching name `tx_flux_column` which has been copied from the to-be-translated record).

For example, I could translate a record with CType `news` which is inside a container, but a CType `news` would not define flux content columns and thus is an invalid container (i.e. `tx_flux_parent`) record?

You're right about not hard-coding extension-related things - if someone would fork `fluidcontent` and rename it `fluidcontent2`, the above code would fail. Is there any better (and still easy) way to do a basic check if the chosen records are capable of being used as containers or should that extra check simply be removed?

NamelessCoder: It's not super obvious but you could in theory turn a `news` plugin into a content container. Use case example could be wanting to put unique content elements which show as default if no news exist (basically: anything you an imagine that can use a relation to a content record and could benefit from editing that relation as a nested content area). Such a conversion can even be done without touching a single line of code or template from the `news` extension (short version: register Provider for CType/list_type, define Grid in form). It is even possible to do this for non-content records; allowing them to contain a Grid of related content - however, such non-content records cannot be edited in the page module by default. Flux was built with the intention that *any* record can become a container (but **child** content is always from `tt_content` since this is the type of relation Flux was designed for). Hopefully that explains why there are no CType restrictions in Flux itself ;)

> Is there any better (and still easy) way to do a basic check if the chosen records are capable of being used as containers or should that extra check simply be removed?

I think we would need to remove it. There *is* a way but it isn't easy - you have to use the Flux API to filter records. The way that would be done is you would detect a Provider for each `$copiedParents` record and ask that Provider (if one can be resolved) for a Grid, then check if that Grid contains at least one row and one column. Looking at it without having tested, leaving out the CType clause should still work just as well.

dneuge: Thanks for the explanation. :)

I now agree we should remove the CType check introduced in 9ffecce064 and re-formatted in db9177ea. With that commit, I also introduced a condition for the container to be on the same page - that one should stay in as a sanity check since the "translated" container of a to-be-translated record should be on the same page or is there something else I missed?

dneuge:
Just a quick question about the best way to apply the changes from our little code review: In less complex pull requests (and without logged code review on GitHub) I `--ammend`ed the commits and used `push --force` since that appears to be the preferred way for Git. If I do that, would we loose this code review? In general, I feel more comfortable if I do not rewrite commit history and would prefer to add a new commit fixing the issues with reference to this discussion.

However... As you said it's a problem if "the other extension" shows up on code searches. In that case, is it actually better to rewrite history for this pull request or is it okay to submit the changes in just another commit?

BTW the extension is also currently being mentioned in `moveRecord` and other places (otherwise I wouldn't have named it on the test cases). To avoid the code search problem, these mentions should be removed as well (maybe not with this PR)?

NamelessCoder:
> Just a quick question about the best way to apply the changes from our little code review

For this one it's preferable to split it into multiple pull requests. The bad thing about that is the inability to just keep adding commits which is very easy, and you often have to rebase or merge `development` branch - the good thing though is much more important to users: individual PRs allow us to link to a pull request from the change log to include all the discussion leading up to the solution and when we merge, we can edit the PR cover message and it becomes the commit message.

And when a PR contains both features, bugfixes and tasks it's pretty much mandatory to split those up based on type ;)

When we merge pull requests we squash and merge, combining all commits into one. You can therefore create individual branches for the changes and make as many commits to those as you please (although we prefer you keep it fairly simple because we have to review all of it).

Not sure if this is useful to you, but on several complex projects I've had good success with creating a single working branch with a single commit, then continuously amending that single commit (and publishing it for review). Once it passes functional review I then undo that commit but keep changes, and begin to stage and commit individual lines. I use the stash function after each commit so I only carry the unstaged lines over to other branches. Not always possible (less likely to work the more stuff you do in one session, obviously) but when it is, it's a smooth workflow imho.

> To avoid the code search problem, these mentions should be removed as well (maybe not with this PR)?

Absolutely, and we welcome any PRs to improve code comments just as much as commits that add features / fix bugs. You'd be helping other contributors ;)

dneuge:
Just a quick update: We didn't find time to follow the comments from above, last year. Since I've changed jobs early this year and I'm no longer working with TYPO3 I don't currently have any installation to test changes to the PR in production. The issue appears to be still relevant (see latest comment on #1125), so I would be grateful if someone else could take over.

NamelessCoder:
No longer required - we've managed to remove the parts that were causing issues, now the translation strategy works completely with core features and most of the overrides are gone.

Still: thank you for your work!

Issue #1165, opened 2016-07-21
[TYPO3 7.6.0] Child elements are duplicated when using a variable amount of sections.


bouche:
[FLUX 7.3.0]
I am using a content element with a variable amount of tabs, as described in #1141, and I followed the advice from @NamelessCoder to use the id-Property with a prefix as field name. Like so:

```
<f:section name="Configuration">
	<flux:form id="contenttabs" label="{f:translate(key: 'contenttabs-fce-label')}"
		description="{f:translate(key: 'contenttabs-fce-description')}" options="{ group: 'FCE' }">

		<flux:form.section name="tab_container" label="{f:translate(key: 'contenttabs-container-label')}">
			<flux:form.object name="tab_element" label="{f:translate(key: 'contenttabs-element-label')}">
				<flux:field.input name="tab_title" label="{f:translate(key: 'contenttabs-tabs-title')}" />
				<flux:field.checkbox name="tab_hide" label="{f:translate(key: 'contenttabs-tabs-hide')}" />
			</flux:form.object>
		</flux:form.section>

		<flux:grid>
			<f:for each="{tab_container}" as="tab" iteration="iteration">
				<flux:grid.row>
					<flux:grid.column name="contenttab{tab.tab_element.id}" label="Tab {iteration.cycle}: {tab.tab_element.tab_title} {f:if(condition: tab.tab_element.tab_hide, then:'[hidden]')}" />
				</flux:grid.row>
			</f:for>
		</flux:grid>
	</flux:form>
</f:section>
```

This, however, leads to another issue: When such an element is created, and content is added to one section, without saving the parent element in the first place, it will be cloned to all available sections when saving the parent element. Here is an animation, which shows process and the unexpected clones at the end:

![content-tabs-bug](https://cloud.githubusercontent.com/assets/769708/17025850/13796258-4f5e-11e6-827e-e2676e4a5cab.gif)

The reason for this behavior is the missing id-properties of the sections. They are not available before the Record has been postProcessed, and this only happens on "update", not on "new" (`AbstractProvider::postProcessRecord`, and ultimately `MiscellaneousUtility::cleanFlexFormXml`).

I tried working around that issue, by somehow forcing an update of the element, but this would require additional action from the user. So my approach was, to simply do the postProcessing also on "new"-operations. This works pretty well, and I couldn't find any drawbacks so far. 

Is there anything that I have not considered here, which argues against this approach? In any case, I will provide a pull request for this change shortly.

PR #1166, opened 2016-07-21
[BUGFIX] PostProcess records on creation

bouche:
Execute the PostProcessing of records also when they are newly created. This is to ensure that the property 'id' of section objects is available right after creating a new content element.

Close: #1165


NamelessCoder:
Should be appropriate and should not cause ill effects. But please do fix the coding style issues reported at https://travis-ci.org/FluidTYPO3/flux/jobs/146409625#L329 (you can either amend and force-push or just add new commits, history will be squashed when merging).

Issue #1174, opened 2016-08-25
Deprecation code


FaraFara:
PHP-7.0.10

Deprecation notice:
PHP Runtime Deprecation Notice: Non-static method 
FluidTYPO3\Flux\Form\AbstractFormComponent::getObjectManager() should not be called statically in 
/typo3conf/ext/flux/Classes/Form/AbstractFormComponent.php line 101

PR #1175, opened 2016-08-25
Retrieve ObjectManager directly through GeneralUtility

cedricziel:
This is just a quick shot to see if we could retrieve the object directly through the global factory.

Ref: #1174


NamelessCoder:
Manually picked - works just fine that way ;)

PR #1180, opened 2016-09-22
Changed internal_type of the file field to file_reference: fixes #1171

### LINKED ISSUE #1171: Images are deleted when publishing workspaces
I have the problem that Images get removed from the Server when i publish a new Content Element.
The error occurs under the following conditions
- The Element is made with flux
- It is new in the Workspace
- I select a file
Bevor publishing, all seems fine. After live Publish, the image file is deleted from the server, but still referenced in the Content Element.

The error occurs **not** when
- I don't user flux on that element
- I don't use workspaces
- I'm only editing the element

CptCompiler:


NamelessCoder:
Sorry - we do not accept pull requests to the `master` branch, only `development` or `legacy`.

Issue #1129, opened 2016-04-27
Pasting an element w/ "gridelements" installed moves the original element to the new location


x3ro:
When Flux and [gridelements](https://typo3.org/extensions/repository/view/gridelements) are both installed, copying and then pasting an element will move the original element to the new location as well. This behavior appears to be caused by Flux' `ContentProvider`, more specifically by the [postProcessCommand method](https://github.com/FluidTYPO3/flux/blob/744e53c/Classes/Provider/ContentProvider.php#L125).

The `$command` parameter passed to the method is `copy`, both with and without `gridelements` installed. However, when using gridelements, the "callbackCommand" passed as `CB` in the query string does not exist, and thus the [`else` branch which moves the record](https://github.com/FluidTYPO3/flux/blob/744e53c/Classes/Provider/ContentProvider.php#L135) is executed. Here's a comparison between the requests when pasting w/ and w/o gridelements:

**Without gridelements:**
~~~
route=/record/commit
&token=0500c027bc9bfd71f90c0875a37c2cc935c3b634
&vC=9af99724f3
&prErr=1
&uPT=1
&CB[paste]=tt_content|28
&CB[pad]=normal&redirect=/typo3/index.php?M=web_layout
&moduleToken=[...]id=28
&CB[update][colPos]=0
&CB[update][sys_language_uid]=0
~~~

**With gridelements:**
~~~
ajaxID=/ajax/record/process
&ajaxToken=[...]
&cmd[tt_content][190][copy][action]=paste
&cmd[tt_content][190][copy][target]=-184
&cmd[tt_content][190][copy][update][colPos]=0
&cmd[tt_content][190][copy][update][tx_gridelements_container]=NaN
&cmd[tt_content][190][copy][update][tx_gridelements_columns]=false
&cmd[tt_content][190][copy][update][sys_language_uid]=0
~~~

To be honest, I have very little knowledge of what's going on there. To me, it looks like gridelements is simply using a different interface to the same functionality, and Flux' `postProcessCommand` is only considering the one that uses `CB`.

Is this something that should be fixed in Flux? Or is this a bug in gridelements? I'd build a PR for this, but I simply don't know what the best way of fixing this would be.

x3ro:
Transcript from IRC by @NamelessCoder on 2016-05-11 starting at 20:55

[...]

PR #1183, opened 2016-10-06
[BUGFIX] Don't move records when copying

neufeind:
Fixes: #1129


NamelessCoder:
Looks like some unit test or tests need to have their 1) expectations or 2) mocked input commands changed! Other than that, this should be fine.

neufeind:
WARNING | [x] Usage of ELSE IF is discouraged; use ELSEIF instead.
I'll take care.

NamelessCoder:
> WARNING | [x] Usage of ELSE IF is discouraged; use ELSEIF instead.
> I'll take care.

Oh, if that's all, that's easy enough :) ready to merge when you're done.

Issue #1176, opened 2016-08-31
Translated record gets random tx_flux_parent and may cause data loss


dmitryd:
Related issues: #1125, #1163, #1120

The code in `ContentService::initializeRecordByNewAndOldAndLanguageUids` selects a random content element as a parent for translated content:

``` php
protected function initializeRecordByNewAndOldAndLanguageUids(
    $row,
    $newUid,
    $oldUid,
    $newLanguageUid,
    $languageFieldName,
    DataHandler $tceMain
) {
    if (0 < $newUid && 0 < $oldUid && 0 < $newLanguageUid) {
        $oldRecord = $this->loadRecordFromDatabase($oldUid);
        if ($oldRecord[$languageFieldName] !== $newLanguageUid && $oldRecord['pid'] === $row['pid']) {
            // look for the translated version of the parent record indicated
            // in this new, translated record. Below, we adjust the parent UID
            // so it has the UID of the translated parent if one exists.
            $translatedParents = (array) $this->workspacesAwareRecordService->get(
                'tt_content',
                'uid,sys_language_uid',
                "t3_origuid = '" . $oldRecord['tx_flux_parent'] . "'" . BackendUtility::deleteClause('tt_content')
            );
            foreach ($translatedParents as $translatedParent) {
                if ($translatedParent['sys_language_uid'] == $newLanguageUid) {
                    // set $translatedParent to the right language ($newLanguageUid):
                    break;
                }
                unset($translatedParent);
            }
            $sortbyFieldName = true === isset($GLOBALS['TCA']['tt_content']['ctrl']['sortby']) ?
                $GLOBALS['TCA']['tt_content']['ctrl']['sortby'] : 'sorting';
            $overrideValues = [
                $sortbyFieldName => $tceMain->resorting('tt_content', $row['pid'], $sortbyFieldName, $oldUid),
                'tx_flux_parent' => $translatedParent ? $translatedParent['uid'] : $oldRecord['tx_flux_parent']
            ];
            $this->updateRecordInDatabase($overrideValues, $newUid);
        }
    }
}
```

That for each cycle only checks the language. Since `tx_flux_parent` is zero, it selects all content elements. So any content elements with a proper `sys_language_uid` field will be selected as a parent for the translated content element, which is obviously wrong.

[...]

PR #1188, opened 2016-10-13
Dont assign random flux parent to translations

pixelbrackets:
This is a bugfix for #1176

While translating records the newly created record currently
always gets assigned a flux parent record even if the origin
had no parent record. When searching for the nonexisting
parent record the database will return a rather random record.

So any translated record without a flux parent got this
random record assigned. If this random record is deleted by an editor
TYPO3 will check for any related records and delete these as well.
This may result in the deletion of all previously translated records.

Test for the existence of a parent record value in the origin record
to prevent the wrong relation.

### DIFF
--- a/Classes/Service/ContentService.php
+++ b/Classes/Service/ContentService.php
@@ -13,6 +13,7 @@ use TYPO3\CMS\Backend\Utility\BackendUtility;
 use TYPO3\CMS\Core\DataHandling\DataHandler;
 use TYPO3\CMS\Core\SingletonInterface;
 use TYPO3\CMS\Core\Utility\GeneralUtility;
+use TYPO3\CMS\Core\Utility\MathUtility;
 
 /**
  * Flux FlexForm integration Service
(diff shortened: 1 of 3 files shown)


NamelessCoder:
Hi Dan, would you mind rebasing this and resolving the conflicts - and check if the patch is still required? I think the conflicts stem from another patch which aims to solve the same issue.

pixelbrackets:
Hey Claus, I have rebased the branch and resolved any conflicts. 

The patch witch may solve the same issue is probably PR #1163 (still pending).

Classes/Service/ContentService.php
```diff
@@ -308,11 +309,14 @@ protected function initializeRecordByNewAndOldAndLanguageUids(
...
     ) {
         if (0 < $newUid && 0 < $oldUid && 0 < $newLanguageUid) {
+            // Get the origin record of the current record to find out if it has any
+            // parameters we need to adapt in the current record.
             $oldRecord = $this->loadRecordFromDatabase($oldUid);
-            if ($oldRecord[$languageFieldName] !== $newLanguageUid && $oldRecord['pid'] === $row['pid']) {
-                // look for the translated version of the parent record indicated
-                // in this new, translated record. Below, we adjust the parent UID
-                // so it has the UID of the translated parent if one exists.
+            if ($oldRecord[$languageFieldName] !== $newLanguageUid && $oldRecord['pid'] === $row['pid'] &&
```
cedricziel: Can we split this over multiple lines please?

pixelbrackets: Of course. I changed the line and rebased the branch.

NamelessCoder:
Manually picked and fixed for CGL - thanks!

PR #1190, opened 2016-10-14
[BUGFIX] Function generateUniqueIntegerForFluxArea

### LINKED ISSUE #1172: Drag&drop does not work on specific column/row combination
When moving a content element with d&d into row2/column2 of a fluid-content grid the change does get persisted correctly. 

The current view is updated properly and the content element gets placed in row2/column2. No javascript error visible and also the AJAX request for the move operation shows no error.
But after a reload the content element is placed inside row3/column1.

The interesting fact ist, that all other areas (row3/column2; row2/column1;..) do work! Only this one area does not work. I suspect a edge-case error in the javascript but had not the time to look closer at this.

It is also not important from which area the element is moved into row2/column2 - the behaviour is always the same.

Versions:
flux: development
fluid-content: development
TYPO3: 7.6.10

Can anyone confirm this bug?

screenshot 2016-08-18 10 24 30

BenjaminBeck:
Fixes #1172 where some areas got the same integer-id

### DIFF
--- a/Classes/Utility/MiscellaneousUtility.php
+++ b/Classes/Utility/MiscellaneousUtility.php
@@ -42,8 +42,11 @@ class MiscellaneousUtility
      */
     public static function generateUniqueIntegerForFluxArea($contentElementUid, $areaName)
     {
-        $integers = array_map('ord', str_split($areaName));
-        $integers[] = $contentElementUid;
+        foreach (str_split($areaName) as $index => $letter) {
+            $position = $index + 1;
+            $integers[] = $position*ord($letter);
+        }
+        $integers[] = (int) $contentElementUid;
         $integers[] = self::UNIQUE_INTEGER_OVERHEAD;
         return 0 - array_sum($integers);
     }
--- a/Tests/Unit/Utility/MiscellaneousUtilityTest.php
+++ b/Tests/Unit/Utility/MiscellaneousUtilityTest.php
@@ -166,16 +166,29 @@ class MiscellaneousUtilityTest extends AbstractTestCase
         $this->assertEquals($expected, $result);
     }
 
+    public function testGenerateUniqueIntegerForFluxAreaDoesGenerateUniqueValues() {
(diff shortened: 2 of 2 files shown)


BenjaminBeck:
Don´t merge yet, i just found a error that i still need to fix..

BenjaminBeck:
It´s fixed and can be merged.

BenjaminBeck:
One additional note:
This fix does decrease the probability of duplicated integers but it is not bulletproofed.
A perfect transformation ( that can also be calculated backward ) of e.g. "row3column3" to a nummeric value would be as big as "114 111 119 51 99 111 108 117 109 110 51" ( https://cryptii.com/text/decimal ).
Sadly my math is not good enough to estimate the probability of duplicates with this implementation.
This should be considered, because once merged those duplicated integer values will happen very seldom and that will make it very hard to debug ..

BenjaminBeck:
.. maybe another way of implementing the column to integer mapping could fix that problem:
If we write a mapping-array from the column-name to a random unique integer (that is not generated from the column-name) in the session variable..
That integer will then change with every new session - but i don´t see a problem with that..
Any opinions?

NamelessCoder:
> If we write a mapping-array from the column-name to a random unique integer (that is not  generated from the column-name) in the session variable..
> That integer will then change with every new session - but i don´t see a problem with that..
> Any opinions?

Off the top of my head, this sounds acceptable, if previews are not cached and get generated fresh for each individual user (you know about this pitfall I'm sure; getting this session-based value cached anywhere would lead to ill effects). We already use this in one other place so I'm not opposed to using session storage, as long as it is kept to the edge cases like this one.

NamelessCoder:
Hi Benjamin,

We're getting close to doing a release (shortly after 15th) - do you have time to look at bit more at this one before then?

NamelessCoder:
Replaced by https://github.com/FluidTYPO3/flux/pull/1268

PR #1195, opened 2016-10-19
[BUGFIX] Avoid issues with changes in DatabaseConnection

NamelessCoder:
Some methods now return different values and need
conditions to avoid PHP warnings. Other issues exist
beneath this one so it is not the only bugfix needed.


Classes/View/PreviewView.php
```diff
@@ -326,11 +326,13 @@ protected function drawGridColumn(array $row, Column $column)
         $dblist = $this->getInitializedPageLayoutView($row);
         $this->configurePageLayoutViewForLanguageMode($dblist);
         $records = $this->getRecords($dblist, $row, $columnName);
-
         $content = '';
-        foreach ($records as $record) {
-            $content .= $this->drawRecord($row, $column, $record, $dblist);
+        if ($records) {
```
xf-: test type like is_array?

PR #1196, opened 2016-10-20
[BUGFIX]: wrong sorting of records in ws

### LINKED ISSUE #1185: [BUG][TYPO3-7.6.11] Changing the order of content elements in a workspace also changes the order in LIVE workspace
## Actual Behavior

The order of content elements in a LIVE workspace changes if you change the order of content elements in a different workspace. This happens with default typo3 content elements in a default 4 column page layout without any dependencies on the flux extension. As soon as I deactivate the flux extension, everything works again as expected.
## Expected Behavior

The order of content elements in a LIVE workspace should not depend on changes in a different workspace.
## Extensions

workspaces, 7.6.0
flux, current development version (commit a477b8f)
## Steps to reproduce

1.) Setup a new typo3 7.6.11
2.) Install the extensions above
3.) Create a page (typo3 default 4 column layout)
4.) Create a new template on the first page with the following typoscript setup:

`page = PAGE`
`page.10 < styles.content.get`

5.) Add **Content Elements (fluid_styled_content)** as static extension to the template
6.) Create 3 **Header Only** content elements in the **normal column**
7.) Create a new workspace for the site and set your current admin user as an owner
8.) Reload the backend. The workspace is now available in the top menu.
9.) Switch to the newly created workspace
10.) Move the first header content element by dragging & dropping it between the second and third header
11.) Switch to the LIVE workspace (or load the LIVE frontend)
## Other information

OS: OS X
MAMP (4.0.5) Stack
PHP Version: 5.6.25
Apache: 2.2.31

bauermar:
This Bugfix is only saving ws records if you work in a workspace copy. 

Second part of bugfix is in vhs (https://github.com/FluidTYPO3/vhs/pull/1141)

(resolves #1185)

### DIFF
--- a/Classes/Service/RecordService.php
+++ b/Classes/Service/RecordService.php
@@ -54,7 +54,11 @@ class RecordService implements SingletonInterface
     public function update($table, array $record)
     {
         $connection = $this->getDatabaseConnection();
-        return $connection->exec_UPDATEquery($table, "uid = '" . intval($record['uid']) . "'", $record);
+        if ($GLOBALS['BE_USER']->workspace > 0 && intval($record['t3ver_wsid']) != $GLOBALS['BE_USER']->workspace) {
+            return false;
+        } else {
+            return $connection->exec_UPDATEquery($table, "uid = '" . intval($record['uid']) . "'", $record);
+        }
     }
 
     /**


xf-:
looks good

Classes/Service/RecordService.php
```diff
@@ -54,7 +54,11 @@ public function getSingle($table, $fields, $uid)
     public function update($table, array $record)
     {
         $connection = $this->getDatabaseConnection();
-        return $connection->exec_UPDATEquery($table, "uid = '" . intval($record['uid']) . "'", $record);
+        if ($GLOBALS['BE_USER']->workspace > 0 && intval($record['t3ver_wsid']) != $GLOBALS['BE_USER']->workspace) {
```
NamelessCoder: Couple of things: the WorkspaceAwareRecordService is the one that is supposed to handle record retrieval in places where we need to respect workspaces. And checks are required to check if the `BE_USER` global is initialised - which it isn't if no user is logged in.

NamelessCoder:
Solved through recent sponsored work on workspaces support.

PR #1206, opened 2016-11-06
[TASK] Increase default cache lifetime to 30 days

NamelessCoder:
This allows various cached Flux data sets to remain
cached for 30 days without needing special care
from the site administrator.


NamelessCoder:
CGL (indents, shorthand array) to be fixed in another PR.

PR #1222, opened 2016-11-27
[TASK] Fix build and enable multiple TYPO3 version testing

NamelessCoder:


cedricziel:
I really don't like the current approach in terms of handling the composer plugin by having a full blown typo3 installation in the src directory. Why not go the subdirectory route I mentioned yesterday? That'll keep the working directory clean and hide the mess.

NamelessCoder:
I'm no fan of it either, but it is the lesser of two evils. We either live with this one folder and two symlinks (and ignore them in git) - or we:

* Have to create a build directory anyway
* Have to relocate the vendor dir there
* Have to fake symlinks to the project root
* Needing to override TYPO3_PATH_WEB
* Needing a custom phpunit command

Ideally, TYPO3 would simply create a "web" dir, create symlinks and leave everything else in place, including leaving TYPO3 sources in the vendor dir. But we're not quite there yet.

NamelessCoder:
Assuming this build succeeds: this one is ready!

cedricziel:
In fact it's quite easy. I'll craft a PR for vhs when I find the time. Thx for your explanation.

NamelessCoder:
Sounds good Cedric! As long as we preserve the relatively easy `./vendor/bin/phpunit` and it runs everything without needing ENV or custom `-c` parameter, then I'm fine with whichever solution you come up with :)

PR #1230, opened 2016-11-29
[TASK] Remove hhvm from build config

cedricziel:
Nobody uses it since the advent of PHP 7


cedricziel:
Thx @bjo3rnf !

NamelessCoder:
Agreed, out with the old!

PR #1231, opened 2016-11-29
[BUGFIX] Create array explicitly when applying defaults

cedricziel:
PHP 7.1 doesn't allow creating array lazily / on-demand.

Fixes one of the PHP 7.1 builds

Contains a second commit that transforms the soup that was the WizardItemsHookSubscriberTest
into a more readable dish.


cedricziel:
Thank you @bjo3rnf !

Issue #1228, opened 2016-11-29
Call to a member function resolveConfigurationProviders() on null


wazum:
After the recent update (dev-development) I get an error after pages are in the cache (works one time after clearing the cache):

Call to a member function resolveConfigurationProviders() on null 
Error thrown in file
/srv/www/client/typo3conf/ext/flux/Classes/ViewHelpers/Form/DataViewHelper.php in line 110.
82 FluidTYPO3\Flux\ViewHelpers\Form\DataViewHelper::renderStatic(array, Closure, TYPO3\CMS\Fluid\Core\Rendering\RenderingContext)
/srv/www/client/typo3temp/Cache/Code/fluid_template/MyTemplate_Content_partial_Content_SomeFluidContent_8966be7ef7ad090acc4055e607607feb886797eb.php: …

This line makes problems:

$providers = static::$configurationService->resolveConfigurationProviders($table, $field, $record);

Any hints on how to resolve this issue?
TYPO3 7.6, all flu* extensions updated;

cedricziel:
Thank you for reporting. I found the issue and will prepare a fix ASAP.

cedricziel:
@wazum I would like you to test whether the current development branch state fixes the issue. If not, feel free to re-open the case.

Thank you for bringing this to our attention! :)

wazum:
Sorry, I don't know how to re-open this issue … too bad it still occurs … do you need a full stack trace (via pm)?

cedricziel:
Nope, thank you. @NamelessCoder brought another issue to my attention.

wazum:
Morning, any news on that? Which commit is not already affected from this bug, so I can go back for now? Tia

cedricziel:
Moin,

I'm not entirely sure about that but you'd do me a favor if you tried.

Thx!

wazum:
The bug occurs after removing typo3-cms-fluid-gap (after commit #965781feec0b943542dd6e11e2e29120f22a3311)

cedricziel:
Do you run the most recent version of TYPO3 7.6?

wazum:
7.6.14, all the Trait Classes from Fluid are there,
and all the flu* and vhs are updated as well (development)

Part of the stack trace
```Uncaught TYPO3 Exception
Call to a member function resolveConfigurationProviders() on null 

Error thrown in file
/var/www/clientxy/typo3conf/ext/flux/Classes/ViewHelpers/Form/DataViewHelper.php in line 110.

[...]

PR #1234, opened 2016-12-05
[BUGFIX] Fix DataViewHelper when injectors don't execute

NamelessCoder:
In fully compiled templates the injection methods will not
execute since a VH class instance is never created. The
classes used are now wrapped in a getter method to create
instances as needed.

NB: inject annotation removed, injection method preserved
in order to allow replacing instances from the outside in a
way that still works with uncompiled rendering.

Close: #1228

### DIFF
--- a/Classes/ViewHelpers/Form/DataViewHelper.php
+++ b/Classes/ViewHelpers/Form/DataViewHelper.php
@@ -26,14 +26,12 @@ class DataViewHelper extends AbstractViewHelper implements CompilableInterface
     use CompileWithRenderStatic;
 
     /**
-     * @var \FluidTYPO3\Flux\Service\FluxService
-     * @inject
+     * @var FluxService
      */
     protected static $configurationService;
 
     /**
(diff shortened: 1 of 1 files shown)


cedricziel:
Please do me a favor and remove the annotations while you're there :D

NamelessCoder:
Mergetool user fail ^^

NamelessCoder:
31ut8opgysl _sx342_

Me for breaking it.

wazum:
please import TYPO3\CMS\…\GeneralUtility and ObjectManager first to use it … otherwise you get the next error:

Class 'FluidTYPO3\Flux\ViewHelpers\Form\GeneralUtility' not found 

;-)

NamelessCoder:
More merge tool fails. Jeez.

cedricziel:
I think we need more coffee. -.-

NamelessCoder:
Cedric I *know* we do! My machine is already warming up, only had one cup today :)

PR #1209, opened 2016-11-07
[FEATURE] Outlet Forms

mneuhaus:
Testfile for this behavior: https://gist.github.com/mneuhaus/9a8cff9f5b47ad8872eebae2808c54e7

PR #1237, opened 2016-12-06
[TASK] Add php-cs-fixer configuration

cedricziel:
Requires php-cs-fixer on the developers machine.

Example: "Reformat all code according to TYPO3 CMS CGL":

```
php-cs-fixer fix --config-file .php_cs
```

Example: "Show a diff and fail the build if not formatted correctly":

```
php-cs-fixer fix --config-file .php_cs --diff --dry-run
```

### DIFF
--- a/.php_cs
+++ b/.php_cs
@@ -1,0 +1,80 @@
+<?php
+/*
+ * This file is part of the TYPO3 CMS project.
+ *
+ * It is free software; you can redistribute it and/or modify it under
+ * the terms of the GNU General Public License, either version 2
+ * of the License, or any later version.
+ *
+ * For the full copyright and license information, please read the
+ * LICENSE.txt file that was distributed with this source code.
+ *
+ * The TYPO3 project - inspiring people to share!
+ */
+
+/**
+ * This file represents the configuration for Code Sniffing PSR-2-related
+ * automatic checks of coding guidelines
+ * Install @fabpot's great php-cs-fixer tool via
+ *
+ *  $ composer global require friendsofphp/php-cs-fixer
+ *
+ * And then simply run
+ *
+ *  $ php-cs-fixer fix --config-file .php_cs
+ *
+ * inside the TYPO3 directory. Warning: This may take up to 10 minutes.
+ *
+ * For more information read:
+ * 	 http://www.php-fig.org/psr/psr-2/
+ * 	 http://cs.sensiolabs.org
+ */
+
+if (PHP_SAPI !== 'cli') {
+    die('This script supports command line usage only. Please check your command.');
+}
+// Define in which folders to search and which folders to exclude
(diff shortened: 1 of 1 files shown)


cedricziel:
We should really add this to make formatting as in #1209 easier out of the box.

.php_cs
```diff
@@ -0,0 +1,80 @@
...
+        'phpdoc_scalar',
+        'no_empty_lines_after_phpdocs',
+        'short_array_syntax',
+        'array_element_white_space_after_comma',
+        'function_typehint_space',
+        'hash_to_slash_comment',
+        'join_function',
+        'lowercase_cast',
+        'namespace_no_leading_whitespace',
+        'native_function_casing',
```
NamelessCoder: I think this one conflicts with several of our hook subscriber method names.

.php_cs
```diff
@@ -0,0 +1,80 @@
...
+        'no_empty_statement',
+        'self_accessor',
+        'short_bool_cast',
```
NamelessCoder: This one should be dropped, if there are others forcing `(int)` vs. `(integer)` that too must be relaxed. And if there is one restricting to `(boolean)$variable` vs. `(boolean) $variable` that too should be relaxed.

cedricziel: Please edit the file according to your needs, edits are enabled - yet we should be en par with the cores' cgl here. What do you think?

cedricziel: Unfortunately we cannot configure this in .editorconfig :-/

IMHO the CGL should cover almost everything. This config does - and it's in active use. Vanilla PSR-2 doesnt cover additional cases we would like.

My proposition stands: Let's use the standard everybody uses so we don't have to think about it anymore.

NamelessCoder:
Have to be honest - I don't like this, for multiple reasons:

* We previously converted to PSR-2 and say we use PSR-2, but this would switch us to "more than PSR-2"
* Switching to this standard imposes more requirements than PSR-2.
* We use `phpcs` / `phpcbf` which are already included as dependencies and work with `--standard=PSR2` - neither one supports a configuration file.
* The configuration encourages a global composer package install and so no longer is declared as `--dev` dependency.
* Except for the TYPO3-specific modifications to PSR-2, our existing tools can both sniff and fix PSR-2 violations (and we can add that to CI, no objections there).
* Adding this configuration means we have to maintain it and likely would mean adding it to all repositories.

We've been in the custom ruleset world before and it always ends up causing more work; maintaining the rules, adjusting them for edge cases, keeping them up-to-date with multiple supported core versions and so on. This is the reason for previously surrendering to PSR-2 vanilla and I would like us to stay there unless there's a seriously good reason not to.

PSR-2 enforced in CI would be welcome though. `./vendor/bin/phpcs --standard=PSR2 Classes/` would be enough.

PR #1239, opened 2016-12-09
[FEATURE] Move emptyOption + items to multivalue base class

NamelessCoder:
Arguments emptyOption and items apply to more than simple
select - and Flux simple select cannot be used for relations.

Refactoring these arguments allows flux:field.relation to respect
the already added emptyOption argument.


Classes/Form/AbstractMultiValueFormField.php
```diff
@@ -203,10 +248,139 @@ public function getRenderType()
...
+        $this->items = $items;
+        return $this;
+    }
+
+    /**
+     * @return array
+     */
+    public function getItems()
+    {
+        $items = [];
```
mneuhaus: couldn't we refactor this to a bit more return-early, since we touch it anyway?

NamelessCoder: Looks like the way to do that would be to move the adding of emptyOption to where $items is used when returned from the function. The problem is, we're building $items in many different ways and then finalising that built array with emptyOption in the end.

Although it would perhaps be cleaner, I think it's safer to keep adding of emptyOption in getItems so this method always returns the "final" array of items.

PR #1241, opened 2016-12-09
[FEATURE] Add API to register template file as CType

NamelessCoder:
The feature will, in all its glorious simpliticy, allow any Flux
template to be used as a custom CType. The following code
is literally all that is required beside the template file:

```php
// your provider extension's ext_localconf.php

\FluidTYPO3\Flux\Core::registerTemplateAsContentType(
    'My.ExtensionName',
    \TYPO3\CMS\Core\Utility\ExtensionManagementUtility::extPath('extension_name', 'Resources/Private/Templates/Content/Columns.html')
);
```

The rest is entirely handled by Flux, including adding the
TCA and pageTSconfig to drive the template.

Custom controller classes can also be used, if they exist
in the extension scope. If the controller class also contains
an action matching the template basename, then that
action is registered as the action to call when rendering
the plugin (thus allowing the same controller to handle
multiple content types / templates):

```php
// Still in your extension's ext_localconf.php

\FluidTYPO3\Flux\Core::registerTemplateAsContentType(
    'My.ExtensionName',
    \TYPO3\CMS\Core\Utility\ExtensionManagementUtility::extPath('extension_name', 'Resources/Private/Templates/Content/Columns.html'),
    My\ExtensionName\Controller\SpecialController::class
);
```

And you can also drop-in a `ContentController` in your
extension's scope - and Flux will autodetect it. E.g. if
class `My\ExtensionName\Controller\ContentController`
exists then this class will be the one rendering your
template - also if the class does not contain a matching
action. This allows you to manipulate things like
initializeView, initializeAction etc.

Finally, when registering your template file a Provider
instance gets returned which you can further modify
with chained syntax to configure more specific things:

```php
\FluidTYPO3\Flux\Core::registerTemplateAsContentType(
    'My.ExtensionName',
    \TYPO3\CMS\Core\Utility\ExtensionManagementUtility::extPath('extension_name', 'Resources/Private/Templates/Content/Columns.html'),
    My\ExtensionName\Controller\SpecialController::class
    )
    ->setTemplatePaths(['partialRootPaths' => ['Other/Path']])
    ->setTemplateVariables([
        'specialVariable' => 'specialValue'
    ])
    ->setConfigurationSectionName('AlternativeConfiguration')
    ->getForm([])
    ->setOption('allowedContentTypes', 'textmedia')
    ->setOption('static', true);
```

Which can be utilised to modify nearly every aspect of
the Flux definition, for example giving them default
values based on EM configuration of your extension.


Classes/Backend/TableConfigurationPostProcessor.php
```diff
@@ -49,11 +54,114 @@ class TableConfigurationPostProcessor implements TableConfigurationPostProcessin
...
+                --div--;LLL:EXT:lang/locallang_tca.xlf:sys_category.tabs.category,categories, 
+                --div--;LLL:EXT:flux/Resources/Private/Language/locallang.xlf:tt_content.tabs.relation, tx_flux_parent, tx_flux_column, tx_flux_children
+            ';
+            $GLOBALS['TCA']['tt_content']['types'][$contentType]['showitem'] = $showItem;
+            $GLOBALS['TCA']['tt_content']['columns']['pi_flexform']['ds']['*,' . $contentType] = [];
+            ExtensionManagementUtility::addToAllTCAtypes('tt_content', 'pi_flexform', $contentType);
+        }
+    }
+
+    /**
```
mneuhaus: imho this method is a bit large, could refactored into a few smaller ones which would improve readability

Classes/Core.php
```diff
@@ -335,6 +354,100 @@ public static function registerFluidFlexFormTable(
...
+        );
+
+        static::$queuedContentTypeRegistrations[$fullContentType] = [
+            $providerExtensionName,
+            $fullContentType,
+            $provider,
+            $emulatedPluginName
+        ];
+
+        return $provider;
```
mneuhaus: same here, it's a pretty big method :)

Classes/Backend/TableConfigurationPostProcessor.php
NamelessCoder: Point taken - it is pretty big yes.

NamelessCoder:
Refactored to more sensible methods and wrapped in a ContentTypeBuilder Helper class.

PR #1247, opened 2016-12-14
[BUGFIX] Setting default value of RenderType to an empty String

monosize:
If In TYPO3 8.5 using an inline field with ```<flux:field.inline.fal>``` or ```<flux:field.inline>``` the output is a select box and not an inline container.
Setting the default value to an empty string is secure, cause the renderType argument of the select ViewHelper use "selectSingle" as default argument.


NamelessCoder:
@monosize Does `null` as default value cause errors? If not, I'd prefer removing the initialised value completely so it becomes implicit null.

monosize:
i will test it, please take also a look at #1248

PR #1251, opened 2016-12-21
[BUGFIX] Serialization of Closure is not allowed

monosize:
getHash thrown an exception during serialization of $this->variables.
Exception thrown in file Classes/View/ViewContext.php in line 249.

If using a flux:form:custom element then a \Closure object is stored in the array of $this->variables,
which throws an error on serialization.

Change serialize into json_encode.


ponyfleisch:
I'm affected by this as well.

NamelessCoder:
I think the reason for unit tests failing here, is that LTS will include objects in the serialised value _by default, regardless of which variables you pass manually_, whereas TYPO3 v8+ won't pass any objects by default. This, combined with `json_encode` only encoding public properties on objects (e.g. not getter method returns, not respecting `__sleep`) means _the cache now is capable of causing collisions because two different contexts can easily end up using the same cache ID, but contain very different variables_.

We can probably avoid the closure in `$this->variables` (for example, by moving it to the ViewHelperVariableContainer and/or removing it from the template variable provider/container after it is attached to the form.

ponyfleisch:
@NamelessCoder i contacted you in IRC regarding sponsoring the fix for this.

monosize:
@NamelessCoder Is it necessary to generate the hash with ```form``` and/or ```grid``` objects inside ```$this->variables```? There are only present in the backend during rendering the previewSection. So we can remove the ```form``` and/or the ```grid``` object from the ```$this->variables``` before serialiazing.

NamelessCoder:
Sounds like a reasonable solution, @monosize - given that the input variables generate the grid and form variables, unsetting those would be okay.

monosize:
@NamelessCoder ```getHash``` now removes the form and grid objects.
Earlier removing runs into an error because some ViewHelpers need the stored form object.
So we must removing the objects directly in ```getHash```.
The form and grid objects are only stored in the previewSection into $this->variables and so the removing of them will not affect other views.

NamelessCoder:
Thanks Frank :)

PR #1253, opened 2016-12-27
[BUGFIX] Fix child content display in BE workspace

NamelessCoder:
Fixes an issue where child content would be shown as
duplicated (same tt_content UID in multiple places).
The following symptoms are fixed:

* New child content shows as duplicated
* Edited child content shows as duplicated
* Published changes not shown in workspace
* "Show hidden" option not respected correctly

The fixes are achieved by:

* Adding the current PID restriction to SQL query
* Adding workspaces fields restrictions to SQL query
* Correcting the referencing of $moduleData from
   undefined class property to correct local variable.

Fix kindly sponsored by MSI Chicago - http://www.msichicago.org/

### DIFF
--- a/Classes/View/PreviewView.php
+++ b/Classes/View/PreviewView.php
@@ -503,16 +503,25 @@ class PreviewView
      */
     protected function getRecords(PageLayoutView $view, array $row, $area)
     {
+        $pidCondition = sprintf(' AND pid = %d ', (integer) $row['pid']);
+        $workspaceCondition = '';
+        $workspaceId = $this->getActiveWorkspaceId();
+        if ($workspaceId) {
+            $workspaceCondition = sprintf(' AND (t3ver_wsid = %d OR t3ver_wsid = 0)', $workspaceId);
+            $pidCondition = sprintf(' AND (pid = %d OR (pid = -1 AND t3ver_wsid > 0)) ', (integer) $row['pid']);
+        }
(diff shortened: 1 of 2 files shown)


Classes/View/PreviewView.php
```diff
@@ -503,16 +503,25 @@ protected function getCookie()
      */
     protected function getRecords(PageLayoutView $view, array $row, $area)
     {
+        $pidCondition = sprintf(' AND pid = %d ', (integer) $row['pid']);
+        $workspaceCondition = '';
+        $workspaceId = $this->getActiveWorkspaceId();
+        if ($workspaceId) {
+            $workspaceCondition = sprintf(' AND (t3ver_wsid = %d OR t3ver_wsid = 0)', $workspaceId);
+            $pidCondition = sprintf(' AND (pid = %d OR (pid = -1 AND t3ver_wsid > 0)) ', (integer) $row['pid']);
+        }
```
mneuhaus: isn't there something like enableFields to create this workspace condition stuff?

NamelessCoder: There is, but the ones I found unfortunately only generate the correct conditions for use in frontend. It is possible I missed something though! In any case: the above is approximately the same as those core methods would generate, with the exception of an unwanted "pid" restriction.

mneuhaus: ok, it was just a thought, to keep this as "close" to core as possible for future proofing :)

PR #1254, opened 2016-12-28
[FEATURE] Simplify Provider pattern and shipped Providers

NamelessCoder:
This change is both a refactoring and feature addition.

It refactors all Flux record manipulation away from the
shipped ContentProvider into the DataHandler hook
listener. This refactoring then means the following:

* ContentProvider is no longer needed, since the only
   purpose it has was to provide these record manipulation
   hook methods (and was used for convenience as parent
   class in fluidcontent, probably others).
* Method call tracking on Providers is no longer needed,
   since the methods that needed to be protected, no longer
   exist in Providers. The parent AbstractProvider no longer
   contains any logic in methods which get called multiple
   times from the DataHandler, so there is no need to avoid
   calling them.

Then, because the shipped ContentProvider is dropped, a
substitute solution must exist which lets you configure a
standard Provider with the attributes ContentProvider
previously provided as defaults. This means a change to
the ProviderInterface to allow setting controller name and
action name via API, which in turn allows a default Provider
to be used as a "ContentProvider". These methods also
mean that it is possible to "tell" a Provider which controller
name it logically relates to - where by default, the class
name of the Provider instance would be scanned to extract
for example "Content" from the class name "ContentProvider"
and use this as controller name.

Essentially: it is now possible to configure some previously
internal-only, override-only behaviors using public API.

Unfortunately though, this *is* a breaking change for any
extension subclassing ContentProvider. Migration is however
very easy: simply override the table- and field name properties
in your Provider so it targets field "tt_content.pi_flexform". If
you used the Provider for a plugin (e.g. not CType) you may
also need to override the "listType" value (but chances are,
you have already done so).


deerfish:
Guten Tag
 
Ich bin zwischen Weihnachten und Neujahr 2016 nur teilweise anwesend.
 
Die Nachricht wird nicht weiter geleitet! In dringenden Angelegenheiten erreichen Sie mich auf dem Mobile. Vielen Dank!

Sonnige Grüsse und und frohes und spannendes 2017 wünscht

Damian Parpan



Damian Parpan
appline
Hirschengraben 52
6003 Luzern
Switzerland

T: +4141 266 0034
M: +4179 303 4878

www.appline.ch

PR #1259, opened 2016-12-30
[BUGFIX] Fix moving relative to other element

NamelessCoder:
Adds a few further bug fixes:

* Switched to post-processing move hooks since these
   are the only ones to receive the necessary input values.
* Implemented necessary saving of record after post hooks,
   since record has already been saved by core.
* Turned colPos into a positive value to make the core read
   it as such.
* Corrected PreviewView to insert the target colPos value.
* Added version copy shadow columns configuration to make
   newly created versions inherit column positions.
* Fixed FE preview output to exclude versioned placeholders.

This should be the last required fix for drag and drop support
in the backend. Moving of both original records and placeholders
is now fully supported; however, bugs still remain in the publish
process (which reverts the colPos to an incorrect value).

Fix kindly sponsored by MSI Chicago - http://www.msichicago.org/

### DIFF
--- a/Classes/Utility/MiscellaneousUtility.php
+++ b/Classes/Utility/MiscellaneousUtility.php
@@ -9,6 +9,7 @@ namespace FluidTYPO3\Flux\Utility;
  */
 
 use FluidTYPO3\Flux\Form;
+use FluidTYPO3\Flux\Service\ContentService;
 use TYPO3\CMS\Core\Imaging\GraphicalFunctions;
 use TYPO3\CMS\Core\Imaging\Icon;
 use TYPO3\CMS\Core\Imaging\IconFactory;
@@ -21,8 +22,8 @@ use TYPO3\CMS\Core\Utility\GeneralUtility;
 class MiscellaneousUtility
 {
 
-    /** Overhead used by unique integer generation. Allows 10 billion records before collision */
-    const UNIQUE_INTEGER_OVERHEAD = 10000000000;
+    /** Overhead used by unique integer generation. */
+    const UNIQUE_INTEGER_OVERHEAD = ContentService::COLPOS_FLUXCONTENT;
 
     /**
      * @var array
@@ -44,7 +45,7 @@ class MiscellaneousUtility
         $integers = array_map('ord', str_split($areaName));
         $integers[] = $contentElementUid;
         $integers[] = self::UNIQUE_INTEGER_OVERHEAD;
-        return 0 - array_sum($integers);
+        return array_sum($integers);
     }
 
     /**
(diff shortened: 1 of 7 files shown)


Classes/Utility/MiscellaneousUtility.php
```diff
@@ -21,8 +22,8 @@
 class MiscellaneousUtility
 {
 
-    /** Overhead used by unique integer generation. Allows 10 billion records before collision */
-    const UNIQUE_INTEGER_OVERHEAD = 10000000000;
```
mneuhaus: what do you think about using ```PHP_INT_MAX``` here instead of a hard-coded number?

NamelessCoder: PHP_INT_MAX wouldn't really apply here since now, it's a positive number and the unique integer value would automatically become too large. Now it starts from 18181 and only applies to real colPos numbers, so it should be "safe enough". Knock on wood.

monosize:
@NamelessCoder drag & drop not working.

tested in TYPO3 7.6 master with a generated fluid_content element with a ```flux:grid``` creating 4 columns ( col1, col2, col3, col4 ). 
Moving an element from 'col3' to 'col1' (or any other movement)
The colPos has the value from the data-colPos ( target area ) and the tx_flux_column isn't changed to the target column ('col1')

is here is missing a call to ```\FluidTYPO3\Flux\Service\ContentService::moveRecord```?

Tested with inactive workspace extension.

NamelessCoder:
@monosize I just tested this a few minutes ago and I'm not able to reproduce any issues like this. Drag and drop in and out of, and between nested content areas, is working on my 7.6.15-dev with the updated Flux sources.

You are sure you've got the commit from this PR included and you've cleared all caches (from install tool), and that you flushed opcode caches and have destroyed your session (php session ID cookie)? The recent changes modify a class constant and changes the session storage, so properly flushing after updating requires flushing all of opcache, session and TYPO3 temp files.

monosize:
@NamelessCoder yes, flushing all caches and cookies. Still same problem.
creating new elements in flux:grid columns also not working. colPos is here correct but tx_flux_column is empty.

I will do a test with a fresh installation.

monosize:
@NamelessCoder Test with fresh TYPO3 Installation is ok. 
Now I'll checking my previous installation for incompatible code or extensions :-( 
Once I found something, I'll report it.

NamelessCoder:
Thanks @monosize - the one problem I found after testing, is similar to what you describe but only happens if 1) workspaces is active and 2) you edit in a workspace and 3) you use TYPO3 7.6 LTS. That remaining bug is one I'll work to fix - I'd definitely like to know if you find anything that doesn't match that description!

PR #1265, opened 2017-01-02
[BUGFIX] Adapt custom CType registration to work on LTS

NamelessCoder:
Class loading via ObjectManager had to be delayed
since the database connection is not initialized yet
on LTS when reading ext_localconf.php files.

Logic which creates class has been moved to the
table configuration post processing hook where the
CType is also registered.

### DIFF
--- a/Classes/Core.php
+++ b/Classes/Core.php
@@ -365,29 +365,15 @@ class Core
      * @param string $providerExtensionName Vendor.ExtensionName format of extension scope of the template file
      * @param string $templateFilename Absolute path to template file containing Flux definition, EXT:... allowed
      * @param array $variables Optional array of variables which are assigned when rendering the Flux definition
-     * @param null|string $section
-     * @param null|string $paths
-     * @return ProviderInterface
      */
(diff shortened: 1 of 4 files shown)


Classes/Core.php
```diff
@@ -365,30 +365,16 @@ public static function registerFluidFlexFormTable(
...
-     * @param null|string $paths
-     * @return ProviderInterface
      */
     public static function registerTemplateAsContentType(
         $providerExtensionName,
-        $templateFilename,
-        $variables = [],
-        $section = 'Configuration',
-        $paths = null
+        $templateFilename
```
cedricziel: You are changing public API surface here - has this been part of a stable release yet?

NamelessCoder: > has this been part of a stable release yet?

No, has only ever existed in `dev-development` - that's why I didn't care so much to preserve the API, I expect nobody is using it yet.

cedricziel: Thx for the clarification :)

PR #1273, opened 2017-01-09
[BUGFIX] Fix FlexForm empty nodes removal

romm:
This PR changes the way Flux cleans a FlexForm. One of the things done by this cleaning function was to delete all the empty `sheet` elements, because it makes TYPO3 throw an exception in some cases.

But there was a problem in the way this feature was handled: looping on the elements and removing elements at the same time could cause nodes to be skipped. For instance, if three `sheet` nodes were empty in a row, the second one was skipped because the first one was removed and the looping pointer was not correctly updated.

The implemented solution to this (big) issue is to first loop on all `sheet` nodes, and store the one which _should_ be deleted first, _then_ loop again on this list and remove the nodes one by one.

One unit test has also been added to check that the solution works the desired way.

### DIFF
--- a/Classes/Utility/MiscellaneousUtility.php
+++ b/Classes/Utility/MiscellaneousUtility.php
@@ -195,11 +195,18 @@ class MiscellaneousUtility
             }
         }
         // Remove all sheets that no longer contain any fields.
+        $nodesToBeRemoved = [];
         foreach ($dom->getElementsByTagName('sheet') as $sheetNode) {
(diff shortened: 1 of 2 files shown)


cedricziel:
Thank you for your contribution, @romm ! :tada:

NamelessCoder:
Yep, thanks! Didn't know this was your first PR :)

romm:
First PR on Flux, indeed. :+1: 

I'm glad to give something to this wonderful tool! Keep up the good work guys!

PR #1276, opened 2017-01-09
[REMOVAL] Drop support for PHP 5.x

NamelessCoder:
PHP 7 branch is only supported/maintained branch.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,24 +2,19 @@ language: php
 
 sudo: false
 
-php:
-  - "5.5"
-  - "5.6"
+env: COVERAGE="YES" TYPO3_VERSION=">=7.6.13, <7.6.99"
 
-env: TYPO3_VERSION=">=7.6.13, <7.6.99"
+php:
+  - "7.0"
 
 matrix:
   include:
     - php: "7.0"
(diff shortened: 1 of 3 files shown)


EdRoxter:
For a quick question: Does this mean that Flux 8.0.0 will throw warnings or errors on PHP 5.6 (i.e. not work at all), or just that you don't offer support for this kind of setup/won't fix future possible errors that would occur in this environment?
I had a quick PHPCompatibility run with PHP_CodeSniffer for PHP5.6 with Flux 8.0.0 and it didn't throw anything.
I'm asking because I'm running a few systems on Debian 8 (Jessie) which had a version freeze on PHP 5.6, and the next version (9, Codename Stretch) which includes PHP7 will only be released in summer, presumably.

NamelessCoder:
Known failures with php7 include warnings/errors from an array sorting function which you will see when Flux attempts to sort objects (provider instances, when sorting by priority), which you cannot avoid. Your best option is to upgrade php (or run with multiple versions, applying 7.0+ only to your typo3 sites.

7.0+ will also give you a drastic performance increase so it is recommended in general. TYPO3 LTS also supports it and benefits from it.

bjo3rnf:
@EdRoxter a little offtopic but you can get PHP 7 for Debian here: https://www.dotdeb.org/

PR #1283, opened 2017-01-13
[BUGFIX] Fixes accessing the "options.Fluidcontent.sorting"

stespot:
This fixes the error "The property "options.Fluidcontent.sorting" on the subject was not accessible" caused by Fluidcontent. Tested on Typo3 7.6.15 with Flux 8.0.0 and Fluidcontent 4.4.1.


NamelessCoder:
Merged; broken tests are not related and will be fixed separately. Thanks!

PR #1285, opened 2017-01-16
[BUGFIX] Write caches correctly in DynamicFlexForm

cedricziel:
The cache key generation was bogus and thus never effective.

There's still a bunch of incorrect returns which I'll remove in a followup.

### DIFF
--- a/Classes/Backend/DynamicFlexForm.php
+++ b/Classes/Backend/DynamicFlexForm.php
@@ -122,8 +122,9 @@ class DynamicFlexForm
             if ($form->getOption(Form::OPTION_STATIC)) {
                 $identifier['staticIdentity'] = $formId;
                 $cache = $this->getCache();
-                if ($cache->has('datatructure-' . $formId)) {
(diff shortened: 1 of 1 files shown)


NamelessCoder:
Hmm - as far as I can read, this doesn't change neither the cache key nor the logic used to get/set it?

Classes/Backend/DynamicFlexForm.php
```diff
@@ -122,8 +122,9 @@ public function getDataStructureIdentifierPreProcess(array $tca, $tableName, $fi
             if ($form->getOption(Form::OPTION_STATIC)) {
                 $identifier['staticIdentity'] = $formId;
                 $cache = $this->getCache();
-                if ($cache->has('datatructure-' . $formId)) {
```
cedricziel: `datatructure` vs `datastructure` :)

NamelessCoder: Ach-haaa! Lightbulb.png :)

PR #1286, opened 2017-01-16
[BUGFIX] Fix inconsistent returns

cedricziel:
Also simplifies cache retrieval with a class local cache instance.

### DIFF
--- a/Classes/Backend/DynamicFlexForm.php
+++ b/Classes/Backend/DynamicFlexForm.php
@@ -191,7 +191,7 @@ class DynamicFlexForm
     public function getFlexFormDS_postProcessDS(&$dataStructArray, $conf, &$row, $table, $fieldName)
     {
         // @codingStandardsIgnoreEnd This comment ends CGL exemption to ensure only the signature is exempted!
-
+        $cache = $this->getCache();
         if (empty($fieldName) === true) {
             // Cast NULL if an empty but not-NULL field name was passed. This has significance to the Flux internals in
             // respect to which ConfigurationProvider(s) are returned.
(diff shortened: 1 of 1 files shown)


Classes/Backend/DynamicFlexForm.php
```diff
@@ -323,7 +320,10 @@ protected function recursivelyEvaluateClosures(array $dataStructureArray, array
      */
     protected function getCache()
     {
-        return GeneralUtility::makeInstance(CacheManager::class)->getCache('flux');
+        if ($this->cache !== null) {
```
NamelessCoder: This happens once per field and at most a couple of times; cachemanager already caches the cache so we don't really win anything by this.

Classes/Backend/DynamicFlexForm.php
```diff
@@ -78,9 +78,6 @@ public function injectRecordService(WorkspacesAwareRecordService $recordService)
         $this->recordService = $recordService;
     }
 
-    /**
```
NamelessCoder: Leave this in please - methods without doc comments not allowed.

Classes/Backend/DynamicFlexForm.php
cedricziel: It does, but we save a bunch of calls in GeneralUtility as well. Your choice - I'd be happy to remove it if you think it's better without a local cache.

NamelessCoder: convert it to a static memoizer in the function itself and we've got a deal ;)

cedricziel: I don't think I get what you want to memoize here. The values or a single cache instance on a static property?

cedricziel: First encounter with a scoped static field. Thx for fixing it :)

PR #1290, opened 2017-01-16
[TASK] Add toString method to allow displaying name in fluidpages message

cedricziel:
PageService::getAvailablePageTemplateFiles emits a message which implicitly casts
the ViewContext to string.

### DIFF
--- a/Classes/View/ViewContext.php
+++ b/Classes/View/ViewContext.php
@@ -79,6 +79,14 @@ class ViewContext
     }
 
     /**
+     * To string.
+     */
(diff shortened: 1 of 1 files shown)


NamelessCoder:
Let's avoid casting it to string instead - since the string representation has no other purpose. We can report the extension name and template file in message.

NamelessCoder:
Thanks btw for spending the entire day on these ;)

cedricziel:
Alright, I'm sure I will find some time for it tomorrow.

Issue #1280, opened 2017-01-10
The label of the columns from flux:grid.colum isn't displayed anymore.


monosize:
Does it have a special reason, why the label was removed in the gridcolumn template? 

see https://github.com/FluidTYPO3/flux/commit/d30df5e653c99ea7c37b5f47448df3ad16b47726#diff-340348ed7628cabfd93e78251097e199L65

and 

https://github.com/FluidTYPO3/flux/commit/d30df5e653c99ea7c37b5f47448df3ad16b47726#diff-340348ed7628cabfd93e78251097e199L765

<flux:grid.column name="col1" **label="Labelname"**/>

It was a very useful feature to label the columns for the users.

PR #1295, opened 2017-01-19
[BUGFIX] Show label of flux:grid.column again

monosize:
Using the same design from the page column labels.

columns

#1280


NamelessCoder:
Happy to restore this - but would like to preserve the 10px margin left and right of columns (at least until TYPO3 itself changes the margins of the page module itself). The added background shading is okay.

monosize:
now with margin 10px left and right.

PR #1297, opened 2017-01-20
[BUGFIX] Fix design for collapsing columns

monosize:
Displaying the toggler in the header of the element.
Right next to the edit / hide / delete / buttons.

Bugfix in fluxCollapse.js
Array.isArray() is used to check whether the cookie variable is of type "array" and to avoid javascript errors.
Optimizing Cookie handling; remove "null" items from cookie array.

toggler

Tested in TYPO3 7.6.15


NamelessCoder:
Looks really good and thanks for the cookie optimisation! I've replaced the icons with list-expand / list-collapse and cleaned up the duplication in content icon hook subscriber. And I tested it with TYPO3 master with also shows it correctly.

PR #1299, opened 2017-01-21
[FOLLOWUP][BUGFIX] Fix design for collapsing columns (#1297)

monosize:
The toggler was saved in the cache, after reoload the page layout, the state was wrong.

### DIFF
--- a/Classes/Hooks/ContentIconHookSubscriber.php
+++ b/Classes/Hooks/ContentIconHookSubscriber.php
@@ -37,7 +37,7 @@ class ContentIconHookSubscriber
         'iconWrapper' => '</div><span class="t3-icon t3-icon-empty t3-icon-empty-empty fluidcontent-icon">%s</span><div class="fluidcontent-hack">',
         'gridToggle' => '</div><div class="fluidcontent-toggler">
                             <div class="btn-group btn-group-sm" role="group">
-                            <a class="btn btn-default %s" title="%s" data-toggler-uid="%s">%s</a> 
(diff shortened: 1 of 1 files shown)


NamelessCoder:
Looks like it's my fault this happened. The reason is I wanted to avoid (re-)processing Providers on every element - that's why the icon cache exists after all. If we do this second round of processing, the benefit of the cache is almost completely lost.

Is there another way we could handle this, e.g. adding the expanded/collapsed state to the cache identifier so toggling the state rebuilds / uses other entry?

monosize:
I will think about this.

monosize:
Here the new solution.

NamelessCoder:
I took care of it meanwhile, in https://github.com/FluidTYPO3/flux/commit/9eca81187dae361e89eb78db3724d1e997a57de0 - didn't mean to hijack, but thought I'd better clean up my own mistake ;)

Solution causes a few additional method calls for each record but solves the problem with the wrong icon being retrieved.

NamelessCoder:
Thank you so much for all the time you spend on Flux contributions, Frank! If you want to officially join the team just say the word ;)

monosize:
@NamelessCoder That would be great to join the officially team! Yes I want ;-)

NamelessCoder:
Welcome to the team Frank :) you should now have access to manage issues, merge/close pull requests and perform edits in pull requests that are not your own, for all our repositories. You already know nearly all of our work flow but if you've got questions, email me or catch me on IRC and I'll help you out.

If in doubt about anything don't hesitate to ask ;)

cedricziel:
Welcome to the team and thank you for the contributions!

PR #1301, opened 2017-01-21
[BUGFIX] Remove unnecessary implements entries

monosize:


Classes/ViewHelpers/Pipe/AbstractPipeViewHelper.php
```diff
@@ -20,7 +20,7 @@
 /**
  * Class AbstractPipeViewHelper
  */
-abstract class AbstractPipeViewHelper extends AbstractFormViewHelper implements CompilableInterface
```
NamelessCoder: CompilableInterface is actually required until we drop 7.6 LTS support. The reason is that we use the CompileWith... traits which allow cross compatible compiling support, but on 7.6 LTS the CompilableInterface is still checked before a ViewHelper is allowed to compile.

monosize: ok, I have removed this from the PR

PR #1304, opened 2017-01-30
[TASK] Update DragDrop.js

monosize:
Using the new DragDrop.js from TYPO3 8, which have changed from the jQuery UI sortable to droppable.
Copying the TYPO3 8 css for drag and drop.
Better drag and drop feeling.
Worked also in TYPO3 7.6 ;-)

The copy feature during drag and drop (holding the cmd-key) in TYPO3 8 is disabled in TYPO3 7.6.
This is done with a check if the localized text for 'dragdrop.copy.message' exists.

Fixed a z-index bug during dragging an element outside from a grid column.
Hide the dragged element in the column. This is useful for an element which contains a flux grid.

Tested with in TYPO3 7.6 in Chrome and Firefox.
Tests with TYPO3 8 and Internet Explorer are not done. Maybe somebody else can test this patch in IE and TYPO3 8, especially the hiding of the dragged element in the column.

### DIFF
--- a/Classes/Backend/TceMain.php
+++ b/Classes/Backend/TceMain.php
@@ -149,9 +149,10 @@ class TceMain
      * @param string $id The records id (if any)
      * @param string $relativeTo Filled if command is relative to another element
      * @param DataHandler $reference Reference to the parent object (TCEmain)
+     * @param array $pasteUpdate Extended paste command
      * @return void
      */
-    public function processCmdmap_preProcess(&$command, $table, $id, &$relativeTo, &$reference)
+    public function processCmdmap_preProcess(&$command, $table, $id, &$relativeTo, &$reference, $pasteUpdate)
     {
         $record = $this->resolveRecordForOperation($table, $id);
         $properties = [];
@@ -159,6 +160,9 @@ class TceMain
         if (!empty($clipboardCommand['paste']) && strpos($clipboardCommand['paste'], 'tt_content|') === 0) {
             $properties = (array) $clipboardCommand['update'];
             $clipboardCommand = GeneralUtility::trimExplode('|', $clipboardCommand['paste']);
+        } elseif ($pasteUpdate !== false) {
+            $copyCommand = $this->getCopyPasteCommand();
+            $clipboardCommand = $copyCommand[$table][abs($id)];
         }
 
         // We only want to process clipboard commands, since these do not trigger the moveRecord hooks below
@@ -246,9 +250,11 @@ class TceMain
      * @param string $id The records id (if any)
      * @param string $relativeTo Filled if command is relative to another element
      * @param DataHandler $reference Reference to the parent object (TCEmain)
+     * @param array $pasteUpdate Extended paste command
+     * @param array $pasteDataMap Reference to the additional data, which is inserted to the record after processCmdmap_postProcess
      * @return void
      */
-    public function processCmdmap_postProcess(&$command, $table, $id, &$relativeTo, &$reference)
+    public function processCmdmap_postProcess(&$command, $table, $id, &$relativeTo, &$reference, $pasteUpdate, &$pasteDataMap)
     {
         $record = $this->resolveRecordForOperation($table, $id);
 
@@ -262,6 +268,16 @@ class TceMain
             if (!empty($clipboardCommand['paste']) && strpos($clipboardCommand['paste'], 'tt_content|') === 0) {
                 $properties = (array) $clipboardCommand['update'];
                 $clipboardCommand = GeneralUtility::trimExplode('|', $clipboardCommand['paste']);
+            } elseif ($pasteUpdate !== false) {
+                $copyCommand = $this->getCopyPasteCommand();
+                $clipboardCommand = $copyCommand[$table][abs($id)];
+                if ((integer) $pasteDataMap[$table][$reference->copyMappingArray[$table][$id]]['colPos'] > ContentService::COLPOS_FLUXCONTENT) {
+                    // remove colPos from pasteDataMap if greater than 18181
+                    // to avoid merging the colPos after finished the processCmdmap_postProcess
+                    // in the DataHandler
+                    // if colPos < 18181 then the target is outside of a flux:grid
+                    unset ($pasteDataMap[$table][$reference->copyMappingArray[$table][$id]]['colPos']);
+                }
             }
 
             // We only want to process clipboard commands, since these do not trigger the moveRecord hooks below
@@ -782,6 +798,15 @@ class TceMain
     protected function getClipboardCommand()
     {
         $command = GeneralUtility::_GET('CB');
+        return (array) $command;
+    }
+
+    /**
+     * @return array
+     */
+    protected function getCopyPasteCommand()
+    {
+        $command = GeneralUtility::_GET('cmd');
         return (array) $command;
     }
 
--- a/Resources/Public/css/grid.7.6.css
+++ b/Resources/Public/css/grid.7.6.css
@@ -1,0 +1,151 @@
+.flux-grid { width: 100%; table-layout: fixed}
+.flux-grid td { vertical-align: top; }
+.flux-grid-hidden { display: none; }
+.t3-page-ce-dragitem:hover .t3-page-ce-wrapper-new-ce { cursor: move; }
+
+.flux-grid tr:first-child:last-child td:first-child:last-child {
+	width: 100%;
+}
+.flux-grid tr:first-child:last-child td {
+	min-width: 174px;
+}
+@media (max-width: 750px) {
+	.flux-grid tr:first-child:last-child td {
+		min-width: 50%;
+	}
+}
+@media (max-width: 550px) {
+	.flux-grid tr:first-child:last-child td {
+		min-width: 100%;
+	}
+}
+.t3-page-lang-column .flux-grid td {
+	min-width: 100% !important;
+}
+.flux-grid-column > .t3-grid-cell {
+	margin-left: 10px;
+}
+.flux-grid-column:last-child > .t3-grid-cell {
+	margin-right: 10px;
+}
+
+/* copied from TYPO3 v8.6.0 */
+.t3-page-ce.active-drag {
+	z-index: 4500;
+}
+.t3-page-ce * {
+	border-spacing: 0;
+}
+.t3-page-ce > .t3-page-ce {
+	margin-left: 0;
+	margin-right: 0;
+}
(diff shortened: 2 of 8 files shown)


monosize:
I have found a bug. Please wait for an updated PR.

monosize:
No bug with this PR. I run into problems with activated xdebug plugin in PHP.

monosize:
update grid.css

monosize:
Tested in IE with TYPO3 7.6

NamelessCoder:
Will do a bit of testing on this when I get home.

NamelessCoder:
Overall I like the change. Just two minor things:

The video below shows some jumpiness when using colspan to control the width of columns. Detaching an element causes the container to change widths for the columns.

I'm not completely sure I like the enlarged dropzone effect. Perhaps it would be better if the *actual* drop zone was enlarged on mouse over, so once you enter it, it gets much easier to drop the element correctly (in particular this "hitting the spot" is an issue on my laptop which has a trackpad so a larger drop zone would be neat).

Maybe you have a better idea for that ;)

NamelessCoder:
jumpy

NamelessCoder:
Oh, forgot to say: the "ctrl to copy" should perhaps not be placed as text but rather some visual change when you hold down ctrl/cmd. Like in the page tree where the drag-n-drop with ctrl shows a different icon.

monosize:
A html table is used for the columns and the column width is set by the widest unbreakable content in the cells. I think we must change the css class .flux-grid to
```css
.flux-grid {
    width: 100%;
    table-layout:fixed;
}
```
table-layout:fixed will make the magic ;-)

I will test it.

NamelessCoder:
Much better, no more jumpiness 👍

monosize:
`\FluidTYPO3\Flux\Backend\TceMain`is extended to handle the `extended paste command` which contains data for the copy via holding the ctrl key during drag and drop.

tested in TYPO3v8.6-dev

monosize:
the`extended paste command`also exist in TYPO3 7.6 and the changes in `\FluidTYPO3\Flux\Backend\TceMain`are compatible with TYPO3 7.6

monosize:
Work's fine in TYPO3 7.6 ;-) with no text info for holding the ctrl key

monosize:
One bug found: during drag and drop with ctrl key.
Copy an element to an empty column: element is copied but flux_parent is not set.
I will check this...

monosize:
Drag and drop with copy on first position  in a column now works.
* compatible for flux:grid columns and standard columns
* new data attribute for a flux:grid column 'data-colpos-pastetop=""' to transmit the correct values for the column
  the format is: `PAGEUID-paste-0-CONTENTUID-AREANAME-FLUXCOLPOS`
* in DragDrop.js the data-colpos-pastetop attribute is add to the ajax call.

other changes see comments in the source.

monosize:
I forgot: the changes weren't tested in workspaces.

NamelessCoder:
I'll give them a run through tomorrow - and I'll also test it with workspaces. Got the proper setup for that running, as you can probably see from my recent commits :)

monosize:
Update branch to latest commits https://github.com/FluidTYPO3/flux/commit/281e9dcdbfa430860078843243ec9205263f41e1 and https://github.com/FluidTYPO3/flux/commit/e1fac362884b7a8bf159586556ce758f30d37dac

monosize:
Bug Fix: Copy element during drag and drop to first position in flux:column and the column isn't empty.
The sortnumber was wrong and we got every time the default sorting-number `256`.
The call to `$tceMain->getSortNumber` was changed to fetch the lowest sortnumber of the page.

monosize:
todo: split grid.css into a 7.6 and 8.6 Versions and use CompatibilityRegistry

monosize:
PR is from my side now ready for testing

Classes/View/PreviewView.php
```diff
@@ -66,7 +66,7 @@ class PreviewView
         'gridColumn' => '<td class="flux-grid-column" colspan="%s" rowspan="%s" style="%s">
                             <div class="t3-grid-cell">
                                 <div class="t3-page-column-header"><div class="t3-page-column-header-label">%s</div></div>
-                                <div data-colpos="%s" class="t3js-sortable t3js-sortable-lang t3js-sortable-lang-%s
+                                <div data-colpos="%s" data-colpos-pastetop="%s" class="t3js-sortable t3js-sortable-lang t3js-sortable-lang-%s
```
NamelessCoder: The "data-colPos" here should already contain a value that's an integer higher than ContentService::COLPOS_FLUXCONTENT which points to a session-stored array containing the target parent element and column name. That integer-higher-than-ContentService::COLPOS_FLUXCONTENT approach was chosen to avoid all of these customisations and instead use the default "data-colpos" attribute transferring.

monosize:
Using now the session-stored array instead of using the attribute data-colpos-pastetop.

NamelessCoder:
Hey Frank - I've been testing a bit and it actually looks like we are now compatible enough with the TYPO3 core markup that we can delete our DragAndDrop.js file (which I would much rather do than increase the size of our own override). On LTS a minor CSS fix is needed in grid.css (to prevent max-size of container element when being dragged and when dragging elements into it) but other than that, everything seems to work if I disable our drag-and-drop overrides.

I'll leave some other comments in the code about what is expected to happen in the code you changed - I think maybe the solutions you added, are intended to be handled by other functions and those functions contain bugs. If that's the case we should try to solve those bugs instead of adding new logic.

Classes/Backend/TceMain.php
```diff
@@ -149,16 +149,20 @@ public function processCmdmap_beforeStart(&$reference)
      * @param string $id The records id (if any)
      * @param string $relativeTo Filled if command is relative to another element
      * @param DataHandler $reference Reference to the parent object (TCEmain)
+     * @param array $pasteUpdate Extended paste command
      * @return void
      */
-    public function processCmdmap_preProcess(&$command, $table, $id, &$relativeTo, &$reference)
+    public function processCmdmap_preProcess(&$command, $table, $id, &$relativeTo, &$reference, $pasteUpdate)
```
NamelessCoder: After taking a look into the TYPO3 source code, I think we may be able to move the "postProcess" logic to "preProcess" and skip record updates, instead receiving $pasteUpdate by reference and filling it with the target information. I'll look into that!

Classes/Backend/TceMain.php
```diff
@@ -262,6 +268,16 @@ public function processCmdmap_postProcess(&$command, $table, $id, &$relativeTo,
...
                 $clipboardCommand = GeneralUtility::trimExplode('|', $clipboardCommand['paste']);
+            } elseif ($pasteUpdate !== false) {
+                $copyCommand = $this->getCopyPasteCommand();
+                $clipboardCommand = $copyCommand[$table][abs($id)];
+                if ((integer) $pasteDataMap[$table][$reference->copyMappingArray[$table][$id]]['colPos'] > ContentService::COLPOS_FLUXCONTENT) {
+                    // remove colPos from pasteDataMap if greater than 18181
+                    // to avoid merging the colPos after finished the processCmdmap_postProcess
+                    // in the DataHandler
+                    // if colPos < 18181 then the target is outside of a flux:grid
+                    unset ($pasteDataMap[$table][$reference->copyMappingArray[$table][$id]]['colPos']);
```
NamelessCoder: I'm still not sure we need this... but if we do: setting colPos to COLPOS_FLUXCONTENT (after we've caught and handled the `> COLPOS_FLUXCONTENT` case!) should be the right way to do it.

The change here makes sense as a separate pull request; assuming the problem isn't that our logic in ContentService is doing something wrong when pasting records?

Classes/Backend/TceMain.php
```diff
@@ -786,6 +802,15 @@ protected function getClipboardCommand()
     }
 
     /**
+     * @return array
+     */
+    protected function getCopyPasteCommand()
+    {
+        $command = GeneralUtility::_GET('cmd');
```
NamelessCoder: I'm still not convinced we need to read this command. In the method above where this method gets used, the existing case is there to catch a very specific operation: triggering when you do clipboard-paste commands. It must not trigger after drag-n-drop commands (the moveRecord_* methods should take care of that; they must trigger after drag-n-drop ends).

What I'm thinking is that maybe the moveRecord_* method to move-to-top-of-column is doing something wrong or missing some logic? In any case: those are the methods expected to handle this type of command!

Resources/Public/css/grid.7.6.css
```diff
@@ -0,0 +1,151 @@
...
+	min-width: 100% !important;
+}
+.flux-grid-column > .t3-grid-cell {
+	margin-left: 10px;
+}
+.flux-grid-column:last-child > .t3-grid-cell {
+	margin-right: 10px;
+}
+
+/* copied from TYPO3 v8.6.0 */
```
NamelessCoder: I would not override these CSS styles on LTS. Flux isn't supposed to homogenise the visual appearance on the two versions - our existing overrides are already vulnerable to changes and we can probably expect more of those! I'd like it if we aim to remove our overrides instead (grid.css is still needed though; and we need an additional patch to prevent height limiting and overflow-hidden on contains you are dragging elements into and which are being dragged).

And that should also mean we don't need the 7.6-specific CSS file - one less override to worry about ;)

Resources/Public/css/grid.css
```diff
@@ -1,4 +1,4 @@
-.flux-grid { width: 100%; }
+.flux-grid { width: 100%; table-layout:fixed;}
```
NamelessCoder: THANK YOU! Seriously, thank you so much for finding this one. I couldn't figure out why my 2/3 column sizing was rendering as 50/50. Excellent, please create this as a separate pull request.  :)

If you do, please add to css as well:

```
.t3-is-dragged .t3-page-ce-body {
	max-height: none;
}
```

That's the hight-limit-overflow-hidden prevention for LTS.

NamelessCoder:
Sorry to drag this one out (no pun intended, really). I think we're down to a couple of things that would be great as separate pull requests, I'd rather not merge the other things (the split CSS in particular).

Let's investigate if we really can drop the JS override! That would be a good step imho :)

NamelessCoder:
Hi again Frank - since I assume you're sleeping or out having fun, I took the liberty of pushing the CSS table fix and preparing https://github.com/FluidTYPO3/flux/pull/1337 which suggests removing our drag and drop JS override.

monosize:
can be closed if #1338 is approved.

NamelessCoder:
#1338 merged, closing this issue.

PR #1309, opened 2017-02-01
[BUGFIX] Resolve DS for FAL image

### LINKED ISSUE #1248: [BUG][TYPO3 8.5-dev] flux:field.inline.fal not working
Testing with PR #1247, which displays the correct container.

After selecting an image error 500 is displayed.

form the the sys_log 
```#1446996319: Configuration retrieved from FlexForm is incomplete or not of type "inline". | UnexpectedValueException thrown in file /typo3/sysext/backend/Classes/Controller/FormInlineAjaxController.php in line 921```

Testing with this configuration:
```html
  <flux:form id="testelement">
    <flux:field.inline.fal name="images" multiple="false" minItems="1" maxItems="1" showThumbs="1"
       allowedExtensions="jpg,jpeg,png"
       foreignMatchFields="{fieldname:'image'}"
       foreignUnique="uid_local"
      />
  </flux:form>
</f:section>```

With TYPO3 7.6 there a no problems.

### PR DESCRIPTION
**Pull request by BenjaminBeck**
Fixes #1248 if table uid available (record is saved)

### DIFF
--- a/Classes/Backend/DynamicFlexForm.php
+++ b/Classes/Backend/DynamicFlexForm.php
@@ -96,17 +96,30 @@ class DynamicFlexForm
      */
     public function getDataStructureIdentifierPreProcess(array $tca, $tableName, $fieldName, array $record)
     {
+        return [
+            'type' => 'flux',
+            'tableName' => $tableName,
+            'fieldName' => $fieldName,
+            'record' => $record
+        ];
+    }
+
+    /**
+     * @param array $identifier
+     * @return array
+     */
+    protected function resolveDataStructureByIdentifier(array $identifier)
+    {
+        $record = $identifier['record'];
+        if (!$record) {
+            return [];
+        }
+        $fieldName = $identifier['fieldName'];
         $dataStructArray = [];
         $providers = $this->configurationService->resolveConfigurationProviders($tableName, $fieldName, $record);
         if (count($providers) === 0) {
             return [];
         }
-        $identifier = [
-            'type' => 'flux',
-            'tableName' => $tableName,
-            'fieldName' => $fieldName,
-            'uid' => $record['uid']
-        ];
         foreach ($providers as $provider) {
             $form = $provider->getForm($record);
             if (!$form) {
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: Classes/Backend/DynamicFlexForm.php [hunk 1]
```diff
@@ -162,6 +162,10 @@ public function parseDataStructureByIdentifierPreProcess(array $identifier)
         } else {
             $dataSourceIdentity = $identifier['tableName'] . '.' . $identifier['fieldName'] . ':' . $identifier['uid'];
         }
+        if(!isset(static::$generatedDataSources[$dataSourceIdentity])) {
```
NamelessCoder: This should already be filled by the `getDataStructureIdentifierPreProcess` which gets called by FormEngine. Clearly the identifier was generated by the pre process method, because it contains the record UID. This means that the DS should already have been generated and stored in the internal cache.

Calling the pre-processing method from here is definitely not the way we want to go - the problem should be solved in the pre process method, if this method for some reason fails to generate a proper DS when the UID is provided. Such behaviour is definitely not expected, but I don't currently know what the cause might be. The pre-processing method should be called consistently so it is probably some flaw in the method's logic as implemented in this class...

BenjaminBeck: This is a screenshot with the current state:
screenshot 2017-02-01 14 42 06

This is a AJAx request that is being made when a new FAL image is added.
( Have a look at the call stack in the lower left corner )
The identifier was generated by ``FormInlineAjaxController::getFlexFormDataStructureIdentifierFromAjaxContext``.

I am not sure how to change it.
Can you give me a direction?

NamelessCoder: I see the issue - identifier is already provided so the hook method above isn't being called when in AJAX mode. It is correct that we need a way to generate the DS in this case.

I would choose to refactor the `getDataStructureIdentifierPreProcess` to separate the logic that generates the data source array, putting that in a protected method which we can call here. That way we don't cause a new identifier to be created and stored.

I can take care of this one later today if you don't want to mess with that refactoring ;)

BenjaminBeck: Thanks, i´ll leave the refactoring to you. I am a bit confused about the usage of the identifier array. Maybe this could become a small data storage class like the ``TemplatePaths`` .. ?

NamelessCoder: > Maybe this could become a small data storage class like the `TemplatePaths`

I think Christian Kuhn already considered something along those lines but opted against it in favor of just arrays. But I'll mention it to him again - unfortunately though that would require a breaking method signature change at this point.

NamelessCoder:
Just updating to say I haven't forgotten about this and will get around to it before releasing Flux 8.0.2 which will be in a few days, tops.

NamelessCoder:
Also solves DS resolving for *new* records which is quite an added bonus. I'm sure there are still edge cases it doesn't solve - but at least my tests on 8.6-dev with a simple FAL relation works correctly now both before and after saving the parent record.

PR #1311, opened 2017-02-05
[BUGFIX] Without krsort the template order is wrong

monosize:
Reverse by key to check the path with the highest number first.
Without the wrong templatepath was used (TYPO3 7.6)

### DIFF
--- a/Classes/View/TemplatePaths.php
+++ b/Classes/View/TemplatePaths.php
@@ -67,6 +67,11 @@ class TemplatePaths
     protected $partialRootPaths = [];
 
     /**
+     * @var boolean|null
+     */
+    protected static $needReverseOrder;
+
+    /**
      * @param string|NULL $packageOrPaths
      */
     public function __construct($packageOrPaths = null)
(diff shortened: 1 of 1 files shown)


NamelessCoder:
I checked on LTS before removing the `krsort` and the order was correct... but maybe I missed something in my test :/

In any case, this krsort must not happen for TYPO3v8+. There, highest priority paths must come last.

monosize:
I can insert a version check into the PR.
But I think we need for TYPO3v8+ a ksort to respect the right order of the keys - highest number last - because array_merge use the creation order of the array and doesn't respect the keys.

NamelessCoder:
@monosize if you have the time, please confirm this one on master. I think I see where I went wrong: I defined my paths in the expected order in TS, didn't think to define them in random order... this seems like a rather high priority fix but I'm a but tied up with other tasks at the moment!

monosize:
@NamelessCoder before I make an error. Is it ok to change the base from this PR to "master". Or is there another way you prefer.

NamelessCoder:
@monosize keep targeting `development` branch, master is only ever touched when doing releases.

monosize:
@NamelessCoder but how can I confirm this one on master, or did you mean something different? I need some support here. Excuse my inexperience ;-)

NamelessCoder:
I meant to confirm it on TYPO3's master branch, or just the most recent 8.5 release should be enough. Install that, apply your patch and debug to see if it needs `ksort()` like you suspect it might ;)

monosize:
yes, ksort is also needed.

NamelessCoder:
Side note, I expect we'll drop this class and some other ones too, once TYPO3v8 LTS becomes our minimum requirement :)

PR #1316, opened 2017-02-09
[BUGFIX] Add `tt_content` in list of tables affected by `new` command

romm:
This commit fixes a severe issue that would return a completely wrong page id for a given record upon creation.

This could lead to really severe issue. For instance the wrong TypoScript tree was loaded for a new content using FluidContent, only on creation.

I did only add `tt_content` to the list, but I really think we should consider modify this condition to be only checking the command type to be `new`, as every record just can't have a pid when it is not even created in database yet.


NamelessCoder:
I'll happily consider a second PR for the "only handle new command" case - though I'm not exactly sure off the top of my head whether I agree it's needed.

PR #1317, opened 2017-02-09
ensures GLOB_BRACE is actually available before using it, also

torvitas:
implements fallback for none GNU systems that do not support GLOB_BRACE


torvitas:
I thought about creating a unit test that tests the new lines. GLOB_BRACE is a predefined constant and there for cannot be mocked. To be able to test this condition I'd have to create a getter for GLOB_BRACE. Where would you want me to put that?

PR #1318, opened 2017-02-09
[BUGFIX] Set missing escaping behavior in ViewHelpers

monosize:


monosize:
Here is a call to `$renderChildrenClosure();`. I thought that the result is escaped.
see https://github.com/FluidTYPO3/flux/blob/development/Classes/ViewHelpers/Pipe/EmailViewHelper.php#L61 
or is this not relevant?

NamelessCoder:
> or is this not relevant?

The Flux form ViewHelpers are a bit different: with the exception of one ViewHelper, none of them actually uses the content returned from "rendering" children. I say "rendering" because the output isn't used - the variables that get generated are directly attached to the Form object being built and the output from $renderChildrenClosure() is discarded.

The exception is the "Custom" field type which allows you to define a chunk of HTML in the ViewHelper's tag content; that then gets serialized to be rendered later via FormEngine. All others will simply discard the tag content, but they will render it to cause child ViewHelpers to also be rendered.

So escaping has no effect one way or the other except in that one Custom field type :)

PR #1327, opened 2017-02-12
[BUGFIX] Creating new element which is not a fluidcontent_content element failed.

monosize:
create an element - for example the standard login form, you got an error.

#1478104554: Identifier {"type":"flux","tableName":"tt_content","fieldName":"pi_flexform","record":{"colPos":"0","sys_language_uid":"0","CType":"login","tx_flux_column":"","tx_flux_parent":"0","tx_fed_fcefile":"","pi_flexform":""}} could not be resolved (More information)

TYPO3\CMS\Core\Configuration\FlexForm\Exception\InvalidIdentifierException thrown in file
/typo3/sysext/core/Classes/Configuration/FlexForm/FlexFormTools.php in line 667.

To avoid this error, now it is checked if for CType exists a registered data structure. Then it is not necessary to create the custom identifier.

### DIFF
--- a/Classes/Backend/DynamicFlexForm.php
+++ b/Classes/Backend/DynamicFlexForm.php
@@ -106,6 +106,17 @@ class DynamicFlexForm
         if ((integer) $record['uid']) {
             $limitedRecordData = ['uid' => $record['uid']];
         } else {
+            if (is_array($tca['config']['ds']) && count($tca['config']['ds']) > 0) {
+                // check if for CType exists a registered data structure
+                // then we don't need to build the custom identifier
+                $searchFor = ',' . $record['CType'];
+                $searchForLen = -strlen($searchFor);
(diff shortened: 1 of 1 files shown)


NamelessCoder:
This solution could have conflicted with both our own implementations (we add a default TS) and implementations wishing to replace an existing FlexForm with a Flux based one.

I took the liberty of making https://github.com/FluidTYPO3/flux/commit/c90c02f7e126b986d611bf9bd59c21528002b166 which solves the issue by not returning a DS identifier if the record has no Providers. Since there now by default are no Providers for `tt_content` (there used to be, to handle the record copy/move operations, but those are now directly in the DataHandker hook subscriber), we can safely assume DS is not Flux when no Providers are returned.

NamelessCoder:
Thanks for bringing it to attention btw! I tested with normal content types but of course neglected to test with one using an old DS approach.

PR #1328, opened 2017-02-12
[BUGFIX] Pasting a element

monosize:
When pasting an element at a specific position with the paste button, then the element is always pasted at the top of the column.
Tested in TYPO3 7.6 and latest TYPO3 8


NamelessCoder:
Manually picked - good catch Frank :)

PR #1332, opened 2017-02-16
[BUGFIX] ContentTypeBuilder TCA and PageTS

monosize:
- addBoilerplateTableConfiguration : missing frame palette
- addPageTsConfig : missing addToList in  mod.wizards.newContentElement.wizardItems.GROUP.show
- initializeIfRequired : missing closing bracket


NamelessCoder:
Thanks Frank! Would you mind also taking a look if we need a condition on TYPO3 version in ContentTypeBuilder, to use different LLL files on LTS?

monosize:
I'll check it

PR #1333, opened 2017-02-16
[BUGFIX] ContentTypeBuilder LTS

monosize:
move showItem configuration into ext_localconf.php - using CompatibilityRegistry
convert arrays in ext_localconf.php into short arrays

### DIFF
--- a/Classes/Helper/ContentTypeBuilder.php
+++ b/Classes/Helper/ContentTypeBuilder.php
@@ -13,6 +13,7 @@ use FluidTYPO3\Flux\Controller\ContentController;
 use FluidTYPO3\Flux\Form;
 use FluidTYPO3\Flux\Provider\Provider;
 use FluidTYPO3\Flux\Provider\ProviderInterface;
+use FluidTYPO3\Flux\Utility\CompatibilityRegistry;
 use FluidTYPO3\Flux\Utility\ExtensionNamingUtility;
 use FluidTYPO3\Flux\Utility\MiscellaneousUtility;
 use TYPO3\CMS\Core\Imaging\IconProvider\BitmapIconProvider;
@@ -30,6 +31,8 @@ use TYPO3\CMS\Extbase\Utility\ExtensionUtility;
  */
 class ContentTypeBuilder
 {
+    const DEFAULT_SHOWITEM = 'defaultShowItem';
+
     /**
      * @param string $providerExtensionName
      * @param string $templateFilename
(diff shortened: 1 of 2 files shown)


monosize:
@NamelessCoder What I just think: can we put the call to CompatibilityRegistry for the DEFAULT_SHOWITEM into an if condition with TYPO3_MODE === 'BE' ?

monosize:
I have made an update which now checks for TYPO3_MODE === 'BE'

monosize:
Please wait: I'll better remove the TCA data from ext_localconf.php to reduce the memory consumption. It's better to use a compatibility class for the TCA data with CompatibilityRegistry.

monosize:
@NamelessCoder maybe we can combine the same TCA configuration in a separate class and can use it for fluidcontent and flux. It's redundant data.
What do you think? 
see https://github.com/FluidTYPO3/fluidcontent/issues/387

monosize:
I changed to a public static property for `$defaultShowItem` so we can use it also in fluidcontent for TCA in `Configuration/TCA/Overrides/tt_content.php`

monosize:
Or should we use `\FluidTYPO3\Flux\Helper\ContentTypeBuilder::addBoilerplateTableConfiguration` to create the TCA in fluidcontent?
see also https://github.com/FluidTYPO3/fluidcontent/issues/388

NamelessCoder:
If possible use addBoilerplateTableConfiguration - also if that means we nedd to add an argument for example.

monosize:
@NamelessCoder Hi Claus. 
the PR is now ready, but take a look at this
https://github.com/FluidTYPO3/flux/blob/development/Classes/Helper/ContentTypeBuilder.php#L204

```php
$GLOBALS['TCA']['tt_content']['columns']['pi_flexform']['ds']['*,' . $contentType] = [];
```
had no effect all the time, because it should be
```php
$GLOBALS['TCA']['tt_content']['columns']['pi_flexform']['config']['ds']['*,' . $contentType] = [];
```
but then the empty array is wrong

I have removed it and now `$GLOBALS['TCA']['tt_content']['columns']['pi_flexform']['config']['ds']['default']` is used for default dataStructure. 

```php
'default' => '
	<T3DataStructure>
	  <ROOT>
		<type>array</type>
		<el>
			<!-- Repeat an element like "xmlTitle" beneath for as many elements you like. Remember to name them uniquely  -->
		  <xmlTitle>
			<TCEforms>
				<label>The Title:</label>
				<config>
					<type>input</type>
					<size>48</size>
				</config>
			</TCEforms>
		  </xmlTitle>
		</el>
	  </ROOT>
	</T3DataStructure>
'
```

NamelessCoder:
You can skip the new class with static properties containing TCA (generally speaking, this pattern is not desired; metadata should not be configured in a class). Instead, use the CompatibilityRegistry in the TCA override file for tt_content and apply the TCA there as well. The result then gets cached which is fine in this case (better than not caching, really).

monosize:
Now the CompatibilityRegistry is used in TCA override file for tt_content.

PR #1334, opened 2017-02-17
[TASK] Remove CompatibilityRegistry in ext_localconf.php

monosize:
Flux minimal TYPO3 version is 7.6.13 and all entries with CompatibilityRegistry are for 7.6.0.
So we don't need them anymore.

remove CompatibilityRegistry::get in \FluidTYPO3\Flux\Backend\DynamicFlexForm and in \FluidTYPO3\Flux\Hooks\ContentIconHookSubscriber


NamelessCoder:
This isn't exactly clear from the API but: the version you use in the compatibility registry are minimum versions - so the 7.6.0 setting also applies to 7.6.14 and even 8.x if there is no 8.x-specific setting. It is sort of like a "closest minimum match" :)

monosize:
That's exactly what I mean ;-)
The current flux version restriction for TYPO3 is 7.6.13
Today we do not need to use CompatibilityRegistry anymore in the ext_localconf.php. Especially for the versions defined there because they are also used for 8+.

PR #1337, opened 2017-02-22
[REMOVAL] Discard DragnDrop.js override

NamelessCoder:
This patch removes the drag-and-drop integration
override from Flux, causing TYPO3's standard JS
to apply instead. Other aspects of Flux are now in
sync with TYPO3 so much that the override is no
longer needed.

Also prevents issues stemming from TYPO3 8.6
switching to from jQuery sortable to draggable.


NamelessCoder:
@monosize Could use your eyes on this one as well - looks like this brings the 8.6 drag and drop feel in sync with your suggested update to DragDrop.js, and drag and drop still works. Maybe you can spot a problem that I'm missing?

monosize:
@NamelessCoder drag and drop works. 
only drag and drop with ctrl key (copy) on 8.6 isn't working.

NamelessCoder:
@monosize Great! Matches my tests - and this could be where we need to take the added arguments for the data map commands into consideration. Let's start here if you're happy with this and I'll be happy to consider a new PR to fix the CTRL+drop ;)

monosize:
@NamelessCoder yes, that make sense ;-)

monosize:
@NamelessCoder we have yet only one visual problem when dragging elements outside a flux:grid element container to an element which is is below the current element.

And I'm afraid that we'll have to change something in JavaScript again. (e.g. moving the element at start of dragging to the end of the document, so it's in front of all elements)
dragdrop

NamelessCoder:
@monosize Ah, I did not encounter that combination while testing. Yep, we'll need to address that. Is there a chance we can do that with a tiny extra drag-start / drag-end listener which like you says, moves the element to bottom of DOM? Just so we don't end up with a full blown JS module override.

monosize:
@NamelessCoder this solves the problem with the hidden dragged element. see https://forge.typo3.org/issues/80024

PR #1338, opened 2017-02-22
[BUGFIX] Copying during drag and drop TYPO3v8.6

monosize:
setting the missing flux related data in processCmdmap_beforeStart


monosize:
@NamelessCoder That it is ;-)

NamelessCoder:
Nice, elegant solution!

NamelessCoder:
Works like a charm btw ;)

Issue #1305, opened 2017-01-31
"Paste as reference" not working


monosize:
The Button "Paste as reference in this position" makes now a copy of the element.

```\FluidTYPO3\Flux\Service\ContentService::pasteAfter ``` isn't call anymore.
Inside this method there is a call to ```\FluidTYPO3\Flux\Service\ContentService::createMappingArray``` which does the job with the subcommand 'reference'. Now this methods are unused code.

Formerly it was done in ```postProcessCommand```.
How can we insert this feature again? Shall we create a new ```InsertContentReferenceProvider``` and register it with ```    \FluidTYPO3\Flux\Core::registerConfigurationProvider('FluidTYPO3\Flux\Provider\InsertContentReferenceProvider')``` or is this a core feature, which can be placed in ```\FluidTYPO3\Flux\Provider\AbstractProvider:: postProcessCommand ``` ?

NamelessCoder:
The "paste as reference" feature is temporarily disabled and has to be recreated in a way that it supports workspaces. You would be *very* welcome to look into that!

No provider is needed; this is low-level enough that it should simply be put into the TceMain class. There is a caveat, though. The existing "pasteAfter" method's business logic cannot be reused, as it would not work with workspaces - and we need to ensure that the original "copy" operation does not happen and then manually create either a new record or a new versioned content record, depending on whether or not we are in a workspace. After this, ContentService's moveRecord method should be called on the record and the input arguments adjusted so action becomes "move".

When all this is done, "pasteAfter" and perhaps other methods as well, can be deleted.

fnagel:
> The Button "Paste as reference in this position" makes now a copy of the element.

Question: is there a way to disable the "reference" button as long it's only a duplicate of the copy button?

gessulatgessulat:
This this issue finally solved? Still can't refer a content element.

typecat:
@gessulatgessulat You can use a workaround: 
1. Right click on a CE and select "copy"
2. On the target page add the CE shortcut
3. There click the button "Insert 1 record(s) from the clipboard" (appears beneath the folder symbol)

NamelessCoder:
"Paste as reference" will not be brought back, sorry for the inconvenience. The tiny guide posted by @typecat is pretty much the perfect solution which has only one or two extra steps compared to having a dedicated button.

PR #1341, opened 2017-02-24
[BUGFIX] Paste into icon is shown twice in TYPO3v8.6

monosize:
since v8.6 the paste icons are inserted with javascript.

change the icon from "actions-document-paste-after" into "actions-document-paste-into" for TYPO3 version lower than 8.6

### DIFF
--- a/Classes/Backend/TceMain.php
+++ b/Classes/Backend/TceMain.php
@@ -256,9 +256,11 @@ class TceMain
      * @param string $id The records id (if any)
      * @param string $relativeTo Filled if command is relative to another element
      * @param DataHandler $reference Reference to the parent object (TCEmain)
+     * @param boolean $pasteUpdate
+     * @param array $pasteDataMap
      * @return void
      */
-    public function processCmdmap_postProcess(&$command, $table, $id, &$relativeTo, &$reference)
+    public function processCmdmap_postProcess(&$command, $table, $id, &$relativeTo, &$reference, $pasteUpdate, &$pasteDataMap)
     {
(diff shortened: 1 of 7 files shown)


monosize:
@NamelessCoder Hi Claus, we should insert the 'paste as reference' button in TYPO3v8.6+ also with javascript if #1305 is solved. So the button would shown on non flux:grid columns ;-)

NamelessCoder:
Hi Frank - I think we should use the CompatibilityRegistry for this one (define in localconf) and register for example a boolean variable named `PreviewViev::class . '->drawPasteIcon' that we set to `true` for `7.6.0` and change to `false` on `8.6.0` and above. Then simply read that variable directly in a condition in PreviewView so the icon is only written if the variable is true.

And yes to drawing our paste icons with JS to also render paste as reference in non-flux columns! :)

monosize:
Hi Claus, now it uses CompatibilityRegistry.

monosize:
@NamelessCoder Cut and Paste in empty flux:grid column doesn't work see https://github.com/FluidTYPO3/flux/pull/1362/commits/d8b5490737dbe1e64747ba8663de4f55e25a89bd#diff-1643673e7b3b4c4d282987bd9ccfd3b4R244 from my PR

monosize:
@NamelessCoder The problem is to insert the element at the beginning of the column ( with or without elements in the flux:grid column). Cut an element with the context menu and then click on the paste icon to insert it at the top of a flux:grid column.

monosize:
Bugfix is here #1363

PR #1348, opened 2017-02-28
[BUGFIX] translate content from other than default source results in internal server error

### LINKED ISSUE #1347: translate content from other than default source results in internal server error
This happens because flux expects the third argument to always be the parent record in default language in ContentService::fixPositionInLocalization.
https://github.com/FluidTYPO3/flux/blob/development/Classes/Service/ContentService.php#L431

If you chose to translate from an already localized record TYPO3 will pass this record to fixPositionInLocalization instead of the default one. Obviously BackendUtility::getRecordLocalization won't find the record using the uid of the localized record.

torvitas:
fixes #1347


monosize:
@NamelessCoder I have checked it with a modified version from #1331 and we need this PR. Thanks to @torvitas

PR #1355, opened 2017-03-01
Fixes invalid where-clause

s-diez:
WorkspacesAwareRecordService::get() expects a full valid where-clause, but BackendUtility::deleteClause('tt_content') returns 'AND deleted=0'.
This resulted in the following error when translating some content elements.
```
ERROR => "You have an error in your SQL syntax; check the manual that corresponds to y
      our MariaDB server version for the right syntax to use near 'AND tt_content.
      deleted=0' at line 1" (172 chars)
   lastBuiltQuery => "SELECT uid,sys_language_uid FROM tt_content WHERE  AND tt_content.deleted=0" (75 chars)
```
This bug was introduced in pull request #1256 with https://github.com/FluidTYPO3/flux/pull/1256/commits/6dcfa1dde68d646daf7bd38254ee148676040907#diff-1643673e7b3b4c4d282987bd9ccfd3b4R318


NamelessCoder:
You already opened a pull request I see, thanks! :)

NamelessCoder:
I'd be interested to know which TYPO3 version you use? There is a problem in 8.0 -> 8.4.0 (solved in 8.4.1) which mangles SQL conditions, so if your TYPO3 version is within that range you must solve this with a TYPO3 upgrade - as Flux is only compatible with 8.x branch starting from 8.4.1.

The SQL condition mangling noted above does not exist in 7.6 LTS.

s-diez:
7.6 LTS was used. I currently cannot check it again but will so over the week and respond again.

NamelessCoder:
Thanks @s-diez - if this is on LTS I'm sure we need the same fix for 8.x so it looks like your solution is good!

NamelessCoder:
Thanks for the pull request - and welcome to the list of contributors! :)

PR #1356, opened 2017-03-02
[TASK] Add autoloading information to ext_emconf

cedricziel:
This changeset adds psr-4 autoloading information to the extension definition file.

### DIFF
--- a/ext_emconf.php
+++ b/ext_emconf.php
@@ -21,22 +21,29 @@ $EM_CONF[$_EXTKEY] = array (
   'author_company' => '',
   'CGLcompliance' => '',
   'CGLcompliance_note' => '',
-  'constraints' => 
+  'constraints' =>
   array (
-    'depends' => 
+    'depends' =>
     array (
       'typo3' => '7.6.13-8.6.99',
       'php' => '7.0.0-7.1.99',
     ),
-    'conflicts' => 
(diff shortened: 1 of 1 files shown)


cedricziel:
Is there a reason we never mapped the tests with `autoload-dev`? I'll add those in both composer.json and ext_emconf as followups, I think.

NamelessCoder:
> Is there a reason we never mapped the tests with `autoload-dev`?

There is indeed - we do this in the unit test bootstrap.php file :)

cedricziel:
Is there any particular reason why we continue to do so? With composer environments, all required classes should be on the autoloader already.. Am I missing something?

NamelessCoder:
No, I don't think you're missing anything. We've got https://github.com/FluidTYPO3/flux/blob/development/Tests/bootstrap.php#L10 but this isn't needed anymore since we stopped replacing/preventing typo3/cms-composer-installers. And if we remove those runtime-added namespaces there will only be the extension-specific ones left - and then those may as well be defined in autoload-dev. It would be okay to do this, as pull requests that both remove the autoload chunk in test bootstrap and adds our autoload-dev section in composer.json :)

cedricziel:
Great, I'll put that one in the queue - thx for your time explaining it! :)

PR #1362, opened 2017-03-03
[BUGFIX] Cut and Paste element in empty flux:grid column

monosize:
The colPos was not correct. It was used the colPos > 18181
Reason:
- in moveRecord there is no areaName set in $parameters[1], so we must fetch the colPos from the $row record and use it to get the columnName from the session stored colPos values.
- command 'move' is using $pasteDataMap and there the colPos is stored. the colPos has a value > 18181 and must unset in $pasteDataMap


NamelessCoder:
I've integrated this fix in https://github.com/FluidTYPO3/flux/pull/1341 - seemed more appropriate to fix the bug at the same time as switching to the native paste button :)

PR #1371, opened 2017-03-12
[BUGFIX] RootPaths must sorted by key in reverse order

monosize:
For TYPO3v8+ the RootPaths must also sorted in reverse order.

### DIFF
--- a/Classes/View/TemplatePaths.php
+++ b/Classes/View/TemplatePaths.php
@@ -323,20 +323,25 @@ class TemplatePaths
 
     /**
      * Initalize rootPaths
-     * TYPO3 < 8.0 we need the rootpaths sorted reverse by key to check the path with the highest number first.
+     * TYPO3 < 8.0 we need the rootpaths sorted by key in reverse order to check the path with the highest number first.
      * TYPO3 >= 8.0 the highest priority paths must come last.
+     * internal special case for flux: the templateRootPaths must be alway sorted by key in reverse order
      *
      * @param mixed $paths
      * @return array
      */
-    protected function initializeRootPaths($paths)  {
+    protected function initializeRootPaths($paths, $reverseorder = false)  {
         $rootPaths = [];
         if (is_array($paths)) {
             // reverse order is needed for TYPO3 Verion < 8.0
             if (static::$needReverseOrder) {
                 krsort($paths, SORT_NUMERIC);
             } else {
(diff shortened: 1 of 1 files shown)


monosize:
Found some other issues - don't merge this PR

monosize:
PR Update: Special case for flux: the templateRootPaths must be alway be sorted by key in reverse order.

NamelessCoder:
This class doesn't have much expected life left in it - when we release the next version that's only compatible with v8 LTS and upwards, this and other classes will simply be removed because Fluid standalone now is capable of doing what Flux's ExposedTemplateView / TemplatePaths used to do.

grimmcreative:
Has this something todo with this Error:
Attempt to insert record on page '[root-level]' (0) where this table, tt_content, is not allowed (msg#1.1.11)

Get it by first Save Command on an ContentElement with flux:form.object
Second save command all good.

cedricziel:
Sooo.... the drop of v7 happened in development today. - with all due respect to the author, do we still need this PR?

NamelessCoder:
Shouldn't be necessary, this patch, since it only applied to v7 LTS. One less bug to worry about :)

monosize:
Hi @NamelessCoder, i got the problems with v8. I can't test this PR now, because a fast switching to https://github.com/FluidTYPO3/flux/commit/9a75549fae88946de6bf9efaa28a471aa55996ac on an existing v8 Project with fluidpages and fluidcontent ( all latest development) isn't possible, too many errors have occurred... I think we need here also updates for fluidpages and fluidcontent. I'm busy the next days, so i can't take a deeper look at the errors. You can check the correct template handling, if you create an extension with a fluidpage page template. Add then via TypoScript additional partials, layouts and templatepaths. create modified templates in the additional directories. This PR can be closed, when the modified templates are used.

NamelessCoder:
> I think we need here also updates for fluidpages and fluidcontent.

One is coming right up for Fluidpages - I don't think Fluidcontent needs any but we'll see.

> This PR can be closed, when the modified templates are used.

I have a testing setup for this one so I'll confirm it and close the PR if necessary.

Take care and don't overwork yourself Frank ;)

NamelessCoder:
Solved in TYPO3 which was the source of the problem, with https://review.typo3.org/#/c/53112/ - and IIRC is available in 8.7.2+ (and of course in 9 when it comes out).

PR #1373, opened 2017-03-13
Feature: Implement ItemsProcFunc for select types

steffenk:
This is a missing configuration for type select which already works out-of-the-box


steffenk:
Noone? This is really straight forward and adds already existing configuration.

deerfish:
Guten Tag
 
Ich bin ferienhalber in der Woche 12, vom 20.03. bis und mit 24.03.2017 abwesend.
 
Die Nachricht wird nicht weiter geleitet! In dringenden Angelegenheiten erreichen Sie mich zeitweise auf dem Mobile, oder hinterlassen Sie eine Nachricht, damit ich Sie zurückrufen kann. Vielen Dank!

Sonnige Grüsse

Damian Parpan



Damian Parpan
appline
Hirschengraben 52
6003 Luzern
Switzerland

T: +4141 266 0034
M: +4179 303 4878

www.appline.ch

NamelessCoder:
The reason none of the Flux components have the `itemsProcFunc` setting (or indeed any other class name / user function references) is conceptual. The reason for having an `itemsProcFunc` is wanting to dynamically add or generate select values - but every context that lets you create these Flux components is already dynamic and provides multiple ways to define the items of selector fields. So the purpose - the very reason for being - of `itemsProcFunc` does not really serve the purpose it should, when in a Flux context. For example, when you're in Fluid a ViewHelper makes more sense - and when you use the PHP components directly, setting the item values is a matter of calling a setter method.

The other perspective is that `itemsProcFunc` is considered legacy - `DataProviders` should be used, and creating/configuring one of those is even more difficult than making a user function. When this is the standard, the Fluid/PHP/TS alternative methods are all a lot simpler to use.

The alternatives that I'd suggest which are all more configurable and transparent than `itemsProcFunc`:

* Create a ViewHelper which generates the items array you need. Call this ViewHelper inline whenever you need the values. This solution has the added benefit that you can generate a list of those item values anywhere and for example check it in Fluid conditions.
* Assign your option values as TypoScript in the `{settings}` scope of your extension - this is a nice alternative to making a ViewHelper and is relevant if you don't need the options to actually be dynamic.
* Create TS overrides that target the Flux component by form ID / name and add your additional option values as an override for the `items` attribute of your selector.

There are other ways but those three are the most common ones; the latter two don't allow dynamic values (e.g. no TS objects get rendered, only the TS array is used) but the first is quite handy in general.

We'd rather not invite people to use these older methods and instead invite them to use ways that are much friendlier to change/override than a hardcoded user function. That's why this particular property was never added in the first place and why I hesitate to accept it now.

steffenk:
Hi Claus,

i see your point. On the other hand, the viewhelpers generate TCA, and itemsProcFunc is a valid TCA setting which exists since old days and will not be removed as it's also needed for pure TCA.
To be honest, the usage of VH is from the same simplicity as the userFunc used by itemsProcFunc, and the itemsProcFunc could be shared for TCA and VH. Using TS for this purpose does not seems to be "correct" imho, as i would overload TS for just some BE selects, and this would also given to FE but not needed for FE rendering. Also DataProviders for this usage is like "using a sledge-hammer to crack a nut".
I don't see that itemsProcFunc speaks against the general concept of FLUX - it's just a TYPO3 API function which does not cost anything.

pixelplant:
Hmm, but if I'm not mistaken a ViewHelper only generates strings, so if you use one for a select, how are you supposed to replace the items returned by an itemsProcFunc with a string returned by the viewhelper? Or is there some way to force a viewhelper into returning arrays?

NamelessCoder:
@pixelplant ViewHelpers return what you make them return, for example that's why most of the VHS iterator ViewHelpers can be chained.

pixelplant:
@NamelessCoder oh yes, you are right. Was a bit silly of me, as creating a simple ViewHelper to test it rendered my custom list just fine. Should have made the test before making the comment :)

NamelessCoder:
It's a fair question and now there's evidence for the Google record :)

pixelplant:
Ok, so turns out it's a neat solution but... I have quite a lot of items displayed in the dropdown. And it seems that my helper that generates the items gets called in both the "Main" and "Preview" sections, even though it is only defined in the "Configuration" section only, which makes my page load very slow in the Backend (all the time) and in the Frontend too (only once the cache needs to be regenerated).
Is there a way to have it get called only on the corresponding section?

NamelessCoder:
> Is there a way to have it get called only on the corresponding section?

Yes - make sure the templates compile, e.g. avoid use of `f:switch` and other viewhelpers which disable compiling. You can use this utility to check: https://github.com/NamelessCoder/typo3-cms-fluid-precompiler

pixelplant:
No switch. The template is quite basic (using it with fluidcontent)

```
{namespace v=FluidTYPO3\Vhs\ViewHelpers}
{namespace flux=FluidTYPO3\Flux\ViewHelpers}
{namespace mine=My\Custom\ViewHelpers}

	<f:layout name="Content" />

		<flux:form id="organisation">
			<flux:field.select name="organisation_emp" items="{mine:getItems()}" required="true"/>
		</flux:form>
	</f:section>

	<f:section name="Preview">
		<f:if condition="{organisation_emp}">
			<f:format.html>{organisation_emp}</f:format.html>
		</f:if>
	</f:section>

	<f:section name="Main">
		<f:format.html>{organisation_emp}</f:format.html>
	</f:section>

```

mine:getItems() just fetches the data from a repository and generates an array of items out of it. It is the call made to the repo that takes a lot of time since it holds a lot of items. As you can see there's no f:switch, nothing fancy, still my viewhelper get's called in all 3 sections.

NamelessCoder:
If `{mine:getItems()}` isn't compilable, none of the template is. You can check out the Flux or VHS ViewHelpers for how to make a VH easily compilable (it involves a Trait and an Interface).

NamelessCoder:
> the itemsProcFunc could be shared for TCA and VH

Sold :)

PR #1378, opened 2017-03-18
[REMOVAL] Drop 7.6 LTS support

NamelessCoder:
Remove 7.6 LTS support and several rarely used and redundant features.


NamelessCoder:
Looks like a good 1500+ LoC can be dropped - good start, far from done yet!

NamelessCoder:
This one is ready - and with the exception of five method calls, will run on TYPO3v9 (as it is in dev-master right now).

Classes/Controller/AbstractFluxController.php
```diff
@@ -12,14 +12,15 @@
...
-use FluidTYPO3\Flux\ViewHelpers\Outlet\FormViewHelper;
+use FluidTYPO3\Flux\ViewHelpers\FormViewHelper;
 use TYPO3\CMS\Core\Utility\GeneralUtility;
 use TYPO3\CMS\Extbase\Configuration\ConfigurationManagerInterface;
 use TYPO3\CMS\Extbase\Mvc\Controller\ActionController;
 use TYPO3\CMS\Extbase\Mvc\Exception\StopActionException;
 use TYPO3\CMS\Extbase\Mvc\View\ViewInterface;
 use TYPO3\CMS\Extbase\Mvc\Web\Response;
+use TYPO3Fluid\Fluid\Core\Rendering\RenderingContextInterface;
+use function get_class;
```
cedricziel: Since when do we import global functions (maybe we should - see the link below)?

FYI: https://veewee.github.io/blog/optimizing-php-performance-by-fq-function-calls/

cedricziel:
Yolo. :tada:

NamelessCoder:
Thanks buddy! You should be able to just plug-and-play this in any site as long as you're not using the TS or Domain Model annotation features, but of course I've already tested it extensively on 8.x and dev-master of TYPO3.

One note: `fluidpages` needs a PR before it becomes compatible because it used the TS feature. I'll prepare that PR soon.

Classes/Controller/AbstractFluxController.php
NamelessCoder: Yep, I may do a full conversion to import global functions, there are thousands of opcode calls to be saved here.

Issue #1380, opened 2017-03-21
Severe Translation Error: Translating content elements that contain a FAL image does lead to incorrect sys_file_reference entries - FAL duplicates for original/root FCE


DenisMir:
We got a pretty severe problem with the latest FluidTypo3 extension family. We are using the latest master versions of vhs, fluidcontent, fluidpages and flux. They are running on TYPO3 7.6.16 LTS (composer mode) running on PHP 7.0.16. 

When translating via copy mode from the default language to another language everything gets copied correctly but the content elements that contain a FAL image do lack the image reference in the new language. 

After digging through the typo3 tables I found the reason for it. 
While copying a content element containing a FAL image from the default language to the foreign language another `sys_file_reference` entry gets created but for the wrong `foreign_uid` and `sys_language_uid`. 
The `foreign_uid` used is the `uid` from the original (root) content element of the default language. The `sys_language_uid` is set to 0. (default)

This has two impacts:

1. The image reference is missing for the copied content element
2. Even worse the original element has now two references in the `sys_file_reference` table which leads to errors while opening in the backend and frontend depending on how you output them. 

I'm not sure if this is a bug in fluidcontent or flux but it creates `sys_file_reference` duplicates when not stopping your editors to translate content containing FAL images.

-- Update --

We have tested a bit more and got to the conclusion that when translating existing pages and its content elements the above mentioned error occurs. 
But the error does not occur when creating new pages and content elements and translating them.

-- Update --

I can now describe the problem even further after testing out more usecases. I can confirm that there is definitely a problem with the latest master versions of flux right here. 

For the following page structure a translation of FCEs containing FAL images does **work**: 

Sorting 256: 

FCE with FAL image inside

Sorting 512: 

FCE with a container area like this:

```
      <flux:grid>
        <flux:grid.row>
          <flux:grid.column name="content">
            <flux:form.variable name="allowedContentTypes" value="fluidcontent_content" />
            <flux:form.variable name="Fluidcontent" value="{allowedContentTypes: '{settings.multiColElements}'}" />
          </flux:grid.column>
        </flux:grid.row>
      </flux:grid>
```

For the following page structure a translation of FCEs containing FAL images does **fail**:

Sorting 256:

FCE with a container area like this:

```
      <flux:grid>
        <flux:grid.row>
          <flux:grid.column name="content">
            <flux:form.variable name="allowedContentTypes" value="fluidcontent_content" />
[...]

PR #1381, opened 2017-03-24
[BUGFIX] Prevent wrong cache key generation (#1380)

DenisMir:
This does prevent a broken cache key to be generated and used by the `AbstractProvider`. Without this bugfix wrong cache objects get used which leads to severe errors during localization and copying of multiple fluid content elements. (see #1380)

### DIFF
--- a/Classes/Provider/AbstractProvider.php
+++ b/Classes/Provider/AbstractProvider.php
@@ -319,11 +319,13 @@ class AbstractProvider implements ProviderInterface
         }
         $formName = 'form';
         $cacheKey = $this->getCacheKeyForStoredVariable($row, $formName);
-        $runtimeCache = $this->getRuntimeCache();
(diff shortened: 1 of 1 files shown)


NamelessCoder:
Thanks a lot for all the hard work identifying this issue, you are completely correct that the key was not unique enough! I created an alternative patch that makes this operation generic and allows values to be cached correctly with all type-specific facets of the record included. Also new records are allowed to use a cached DS (so skipping caching for no-UID records is not correct) and will be as unique as the input record is.

The cache key that Providers use now consists of the table, field, uid (or zero), variable name and the controller action reference (which consists of extension key and action name).

NB: Fluidpages also needs a patch to add the two template selection fields to list of fields which use default values. I will take care of that next.

See https://github.com/FluidTYPO3/flux/commit/666cce643a8b81682c57cac096b5e75c7f53b94a.

PR #1382, opened 2017-03-25
[BUFGIX] Fix an error when copying a plugin

BenjaminBeck:
This fixes a bug where a flexform provider can not be resolved
when a plugin is copied.


NamelessCoder:
Thanks Benjamin for the work identifying the issue. The only problem I would have had with this solution, is that it specifically only targets one field from one table - and the code is intended to work with any field from any table. I chose to solve this in a way that is generic and also solves another reported issue about cache identifiers not being unique. https://github.com/FluidTYPO3/flux/commit/666cce643a8b81682c57cac096b5e75c7f53b94a ensures that `list_type` is included in the partial record (regardless which table gets used).

PR #1383, opened 2017-03-25
[BUGFIX] Grid hover styles in preview

BenjaminBeck:
Apply hover css to nested content only if mouse is over ce.


NamelessCoder:
Excellent - thanks :)

PR #1385, opened 2017-03-28
7.4 BACKPORT: Add change from BUDG-1202 to function getAllTypoScript to improve abysmal backend perfomance.

sven-carstens-udg:
Displaying about 17 content elements in the backend lead to more than 6 million calls to removeDotsFromTs. Caching the prepared array for the backend will drastically reduce backend render time.

As not everybody has finished the move to PHP-7 with the newest flux version, this might prove handy to others.

### DIFF
--- a/Classes/Service/FluxService.php
+++ b/Classes/Service/FluxService.php
@@ -352,18 +352,37 @@ class FluxService implements SingletonInterface {
 	 *
 	 * @return array
 	 */
-	public function getAllTypoScript() {
-		if (!$this->configurationManager instanceof BackendConfigurationManager) {
-			$typoScript = (array) $this->configurationManager->getConfiguration(ConfigurationManagerInterface::CONFIGURATION_TYPE_FULL_TYPOSCRIPT);
-			$typoScript = GeneralUtility::removeDotsFromTS($typoScript);
-			return $typoScript;
-		} else {
-			$pageId = $this->configurationManager->getCurrentPageId();
-			if (FALSE === isset(self::$typoScript[$pageId])) {
-				self::$typoScript[$pageId] = (array) $this->configurationManager->getConfiguration(ConfigurationManagerInterface::CONFIGURATION_TYPE_FULL_TYPOSCRIPT);
-				self::$typoScript[$pageId] = GeneralUtility::removeDotsFromTS(self::$typoScript[$pageId]);
+	public function getAllTypoScript()
+	{
+		static $cache = [];
+		$pageId = $this->getCurrentPageId();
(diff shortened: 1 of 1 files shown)


NamelessCoder:
Hi Sven - I think you don't have the most recent changes from Flux, or you intended to make this for the `legacy` branch but created the fix on your local `development`. Depending on which one is true, you either need to:

* Pull the latest `dev-development` and test your use case again to confirm performance issues are gone.
* Check out `legacy` branch locally, make a new branch starting form that, do the backport and re-create or force-push to the pull request

The change is already in `dev-development` so I'm guessing you meant to target the `legacy` branch?

sven-carstens-udg:
Hi Claus,

I forked from the tag 7.4.0 as this is the version currently used by my project. Github does not offer tags for pull requests. So I don't know how to handle it properly inside of github.
Expected result for me would be to get a version tagged as 7.4.1.

CU Sven

NamelessCoder:
@sven-carstens-udg We don't release bug fixes for the 7.4 branches, only for the most recent release. You should be able to upgrade from Flux 7.4 to 8.1 without breaking stuff, and this version is still compatible with TYPO3 7.6. That should be your best option. It does mean you will need PHP7 but I'd recommend that in general. If for whatever reason you absolutely require that Flux 7.4 is your starting point and this fix gets applied, I suggest simply adding your own forked repository in the composer `repositories` section and use `fluidtypo3/flux:dev-backport/BUG-1202_cache_for_typoscript` to require the branch from this pull request. I strongly recommend you upgrade instead, though - there are plenty other improvements to gain. But if you're in a pinch you can use your fork to "fake" your own bug fix releases since there will be no further releases for the Flux 7.4 branch. Same advise goes for anyone else reading this and for some reason must continue to use Flux 7.4.

What you can expect to happen when we start requiring TYPO3 8.7 and above (starting from Flux 9.0): the `legacy` branch will be overwritten with the last tag that works on TYPO3 7.6. Then it enters the same maintenance priority as the `legacy` branch has now. That means we will completely drop even legacy support for TYPO3 6.2 and set 7.6 as our new legacy support. If you continue to use TYPO3 7.6 at that point, I recommend switching to the `legacy` branch.

So unfortunately I have to reject your pull request - but thank you for taking the time and don't hesitate to create pull requests for the `development` branch, or back-ports for the `legacy` branch if/when that time comes! Hopefully you find some inspiration above to either upgrade or create a (temporary) solution to apply this back-port to your own 7.4 branch.

sven-carstens-udg:
We are stuck at PHP-5.6 and can't upgrade to anything beyond 7.4.0

PR #1402, opened 2017-04-23
[TASK] Clean up tests

NamelessCoder:


NamelessCoder:
`dev-master` tests must be allowed to fail for a while - work is ongoing to remove and change the code that fails because it uses features that cannot be supported on both `7.6` and `9.0-dev` of TYPO3.

PR #1401, opened 2017-04-20
[TASK] Allow installation on TYPO3 CMS 8.7

cedricziel:

PR #1405, opened 2017-04-24
[BUGFIX] Allow installation in TYPO3 8

cedricziel:
Beware that I raised the constraint from `^8.4.1` to `8.7`, but I think it makes absolute sense.

Ref: https://github.com/FluidTYPO3/flux/pull/1401#issuecomment-296576474

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -27,7 +27,7 @@
     },
     "require": {
         "php": ">=7.0.0",
-        "typo3/cms": "^7.6"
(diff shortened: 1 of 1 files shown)


NamelessCoder:
I pushed a fix; we'll stick with the `^8.4.1` constraint as to not drop previous v8 version support until https://github.com/FluidTYPO3/flux/pull/1378 is merged. Sorry about the mess :/

Issue #1407, opened 2017-05-02
[BUG?][TYPO3-7.6.18] recursive correction of "sorting" field by  hook for "copy" results in corruption of sorting fields on multiple records


sypets:
Hello,
a severe problem occured where multiple tt_content records (couple of 1000) had the sorting field (automatically) changed. tstamp was not set and sys_log / sys_history records not created. This occurs when a tt_content text+image is inserted as shortcut (using the functionality from gridelements: "Insert as reference"). Multiple content elements that have nothing to do with this operation will be manipulated. The "sorting" value will be set to their t3_origuid counterparts. 

**Info**

TYPO3 version 7.6.18
compatibility6 installed
flux version 8.2.1
gridelements version 7.1.0

**Reproduce / Analysis**

tracked this down to an insert operation in the history:
_Record 'Reference' (tt_content:223581) was inserted on page 'xyz' (5278) (msg#1.1.10)_

record 223581 is of type shortcut. References a text / image content element with one image.

**Debug**

When the problem occurs, the recursive function **copySortingValueOfChildrenFromOriginalsToCopies** will set the "sorting" field to "sorting" value of its  t3_origuid counterpart. 

This apparently occurs because copySortingValueOfChild... was called with an array containg 0 as 'uid'.

This happens because $reference->copyMappingArray['tt_content'][$id] in processCmdmap_postProcess is empty. In fact $reference->copyMappingArray is empty (possibly due to the insert as reference?)

**Is this really a bug in flux or a result of a bug somewhere else**

- I am not sure, but I think especially the recursive function should make extra sure, that the input values are ok. 

The question is, should DataHandler->copyMappingArray[$table][$uid] always be set? If this should be the case, the bug must be tracked further to the original problem. For now however, an extra check in flux might be not such a bad idea. 

**Possible Fix** 

-  quit the copySortingValueOfChildrenFromOriginalsToCopies function immediately if !$parentRecord['uid'] (and use sys_log to log an error)
- processCmdmap_postProcess should check $reference->copyMappingArray['tt_content'][$id]

**Question**

Does the original problem in the core still exist for fixing the sort order?

[...]

PR #1408, opened 2017-05-05
[BUGFIX] Added check for input value (#1407)

sypets:
I tried to follow the excellent contribution guidelines best as possible (https://fluidtypo3.org/documentation/contributing/contribution-guide.html). composer install was successfull. However vendor/bin/make, vendor/bin/checkstyle and vendor/bin/runtests are missing. 

Ran phpunit manually, 100% pass.
phpcbf had a number of fixes on several files that I did not changed, so I abandoned that.

Hope this is a sufficient PR. If not I will do my best to improve it.

### DIFF
--- a/Classes/Backend/TceMain.php
+++ b/Classes/Backend/TceMain.php
@@ -612,6 +612,11 @@ class TceMain
      */
     protected function copySortingValueOfChildrenFromOriginalsToCopies(array $parentRecord)
     {
+        if (!$parentRecord['uid']) {
+            // make sure value is set (#1407)
+            return;
+        }
+
         $children = $this->recordService->get(
             'tt_content',
             'uid',


sypets:
I closed the initial issue, but I still think the check for input value makes sense, unless of course the entire function can be dropped for TYPO3 8.

NamelessCoder:
My apologies for the outdated information in the contribution guide, it really needs an update. At this point all we require is that you do `composer install` and run `./vendor/bin/phpunit` - or simply wait for Travis to do the testing.

Good job on the patch and not least the issue description in #1407, detailed reproduction steps are *greatly* appreciated. And welcome to the list of contributors! :)

sypets:
Ok. Thank you. I will try to work on a reproducible test case, if possible. Should we bother at all with TYPO3 7.6? Or only report problems (if any) on TYPO3 8?

NamelessCoder:
> Should we bother at all with TYPO3 7.6? Or only report problems (if any) on TYPO3 8?

Let's start with TYPO3 8. When Flux 9.0 gets released the `legacy` branch will be updated to use `8.2.1` as starting point and becomes the recommended "version" to use on TYPO3 7.6. We can then backport a patch to `legacy` so people with TYPO3 7.6 can get the fix as well.

PR #1215, opened 2016-11-14
[BUGFIX] Adapt to FlexForm language handling deprecation

mneuhaus:

PR #1410, opened 2017-05-09
[BUGFIX] Adapt to FlexForm language handling deprecation

achim71:
In addition to https://github.com/FluidTYPO3/flux/pull/1215 above changes are required.

I use customize fluidstyledcontent templates in my project and without that fix I can not access piflexform data from within them.


cedricziel:
I've removed the arguments altogether since they always match the default value. Thx for your contribution!

achim71:
Thanks for the quick response. Removed the arguments here as well and it works.

cedricziel:
Great first time contribution, thank you! - Keep the good stuff coming :)

PR #1419, opened 2017-06-02
[CLEANUP] Remove unused protected methods

romm:


NamelessCoder:
I tend to want to keep those (at least until the core provides a nested content solution natively). Although they are currently unused, there's good chance we end up needing them again. The only reason they're currently unused is that Workspaces-support would not work correctly when updating the data map - only by updating the record in post could they be manipulated enough to work.

But it *is* definitely good to see you're into the code enough to spot these unused methods ;)

romm:
Fair enough, thanks for taking the time to analyze the situation!

Issue #1388, opened 2017-04-03
[BUG][TYPO3-8.6.1] flux:field.text with CKEdit


patricsmn:
Hi,

first of all please excuse my poor english due to it not being my native language. I hope you can get what i am pointing at nonetheless.

With an Update to Version 8.6.1 and activating CKEdit as RTE the editor shows a pretty unwanted behavior.
On every save-action there are empty lines added around every line-end. This results in multiple <p>&nbsc;</p> with an exponential growth. This problem only occurs within flux-forms. When adding a rte-field with FlexForms everything works as intended.

Example with Flux (adding tons of empty lines):

```
<flux:field.text
    name="text"
    label="SomeLabel"
    required="true"
    enableRichText="true"
/>
```

Example with Flexform (working as intended):

```
<TCEforms>
    <label>SomeLabel</label>
    <config>
        <type>text</type>
        <cols>48</cols>
        <rows>4</rows>
        <enableRichtext>1</enableRichtext>
    </config>
    <defaultExtras>richtext[*]:rte_transform[]</defaultExtras>
</TCEforms>
```

I am not sure what transformations are applied to the flux-field-content but since it only occurs with the usage of flux i guess that the bug is not related to the core and i hope you have the time to have a look at it soon. 
Please tell me what additional information you need or what further steps i could take to better assist you.

Kind regards
Patric

edit: using flux 8.1.0 from ter

patricsmn:
This behavior is still showing up on typo3 8.7 with the latest development-branch of flux.

Sethorax:
I ran into the same problem today. I was using TYPO3 8.7.0 and latest flux release.

By looking at the TCA configuration for the default content elements of TYPO3 i noticed that there is additional configuration for content elements which use the RTE.
The TCA configuration for the type **text** for example shows the following configuration in `typo3/sysext/frontend/Configuration/TCA/tt_content.php:1354`:
```php
'columnsOverrides' => [
    'bodytext' => [
        'config' => [
            'enableRichtext' => true,
            'richtextConfiguration' => 'default'
        ]
    ]
]
```

I was able to fix the problem in flux by adding the above configuration to the `buildConfiguration()` function in `Classes/Form/Field/Text.php` like so:

```php
if (true === $this->getEnableRichText()) {
    $configuration['enableRichtext'] = true;
    $configuration['richtextConfiguration'] = 'default';
}
```

Just add it before `$configuration` is returned.

patricsmn:
[...]

PR #1422, opened 2017-06-13
#1421 use enableRichtext

steffenk:
Maybe defaultExtras can be dropped completely.

PR #1426, opened 2017-06-27
[FEATURE] Add richtextConfiguration option on Text fields

cedricziel:
Enables switching the CKEditor configuration based on the current field.

This option falls back to `default` and can be overriden through TypoScript:

```
plugin.tx_flux.settings.flexform.richtextConfiguration = MySite
```

Closes: #1388

### DIFF
--- a/Classes/Form/Field/Text.php
+++ b/Classes/Form/Field/Text.php
@@ -15,6 +15,7 @@ use FluidTYPO3\Flux\Form\FieldInterface;
  */
 class Text extends Input implements FieldInterface
 {
+    const DEFAULT_RTE_CONFIG = 'default';
 
     /**
      * @var integer
@@ -39,6 +40,11 @@ class Text extends Input implements FieldInterface
     /**
      * @var string
      */
+    protected $richtextConfiguration;
+
+    /**
+     * @var string
+     */
     protected $renderType = '';
 
     /**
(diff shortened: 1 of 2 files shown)


olegkarun:
@cedricziel Have no effect. Tried in my flux.form
<flux:field.text name="text" label="Text"  richtextConfiguration="Minimal" enableRichText="TRUE"/>
and in my TypoScript
plugin.tx_flux.settings.flexform.richtextConfiguration = Minimal

Could you create patch based on 8.2.1 for testing. Mb i miss something?

Classes/Form/Field/Text.php
```diff
@@ -173,4 +182,41 @@ public function setFormat($format)
     {
         $this->format = $format;
     }
+
+    /**
+     * @return string
+     */
+    public function getRichtextConfiguration()
+    {
+        return $this->richtextConfiguration;
```
NamelessCoder: Although it makes the overall experience a little less predictable, I think we should move the body of computeRichtextConfiguration() to this getter and drop computeRichtextConfiguration (along the lines of getItems() on AbstractMultiValueFormField). Then you always get the applied configuration also when you access the value separately from buildConfiguration(), like you do with select items.

Caching the resolved configuration in the internal property is completely okay and should help performance, e.g. you can call buildConfiguration() over and over without re-analysing the TS parameter.

NamelessCoder:
Does this pull request replace https://github.com/FluidTYPO3/flux/pull/1422?

cweiske:
I've adjusted the patch to @NamelessCoder's request: https://github.com/mogic-le/flux/tree/richtextconfiguration

cweiske:
I wonder why the config is in TypoScript and not PageTSConfig.

NamelessCoder:
Merged via other PR: https://github.com/FluidTYPO3/flux/pull/1438

PR #1430, opened 2017-06-30
Doc fixes for field.relation

cweiske:


NamelessCoder:
Thanks for taking time to improve these :)

PR #1431, opened 2017-07-07
[BUGFIX] Bind parameters to prepared statement

HenningCash:


cweiske:
Same as #1441, but has 2 commits instead of 1.

NamelessCoder:
#1441 is merged, it was first on my notifications list. Thanks to both of you for taking care of this one!

PR #1434, opened 2017-07-13
[BUGFIX] Use "enableRichtext" when rte_ckeditor is in use.

kraftb:
New TYPO3 version 8.x use rte_ckeditor extension by default instead
of rtehmlarea. For this extension a different TCA/Flex configuration
is required.

Close: #1421

PR #1438, opened 2017-07-24
[FEATURE] Add richtextConfiguration option on Text fields

cweiske:
Enables switching the CKEditor configuration based on the current field.

This option falls back to `default` and can be overriden through TypoScript:

```
plugin.tx_flux.settings.flexform.richtextConfiguration = MySite
```

Resolves: #1388

Open issues:
- [x] Should we fallback to PageTSConfig `RTE.default.preset`?
  - yes
- [x] Should we remove `defaultExtras` completely (param def + getter/setter)?
  - no, we keep that since it can be used for other things as well.
- [x] Keep defaultExtras when CKEditor is not used, see  #1434
- [x] I wonder why the config is in TypoScript and not PageTSConfig (was introduced that way in #1426).
  - we move to PageTSconfig
- [ ] Decide if it is possible to get the current page ID in the text object
- [ ] Which PageTSconfig path to use? @cweiske suggests `RTE.tx_flux.preset`, see https://docs.typo3.org/typo3cms/CoreApiReference/Rte/Transformations/Tsconfig/ for more info

### DIFF
--- a/Classes/Form/Field/Text.php
+++ b/Classes/Form/Field/Text.php
@@ -15,7 +15,6 @@ use FluidTYPO3\Flux\Form\FieldInterface;
  */
 class Text extends Input implements FieldInterface
 {
-
     /**
      * @var integer
      */
@@ -39,6 +38,11 @@ class Text extends Input implements FieldInterface
     /**
      * @var string
      */
+    protected $richtextConfiguration;
+
+    /**
+     * @var string
+     */
     protected $renderType = '';
 
     /**
@@ -55,18 +59,31 @@ class Text extends Input implements FieldInterface
         $configuration['rows'] = $this->getRows();
         $configuration['cols'] = $this->getColumns();
         $configuration['eval'] = $this->getValidate();
-        $defaultExtras = $this->getDefaultExtras();
-        if (true === $this->getEnableRichText() && true === empty($defaultExtras)) {
-            $typoScript = $this->getConfigurationService()->getAllTypoScript();
-            $configuration['defaultExtras'] = $typoScript['plugin']['tx_flux']['settings']['flexform']['rteDefaults'];
+        if (true === $this->getEnableRichText()) {
+            $configuration['enableRichtext'] = true;
+            if ($this->isCkEditorActive()) {
+                $configuration['richtextConfiguration'] = $this->getRichtextConfiguration();
+            } else {
+                //rtehtmlarea
+                $defaultExtras = $this->getDefaultExtras();
+                if (true === empty($defaultExtras)) {
+                    $typoScript = $this->getConfigurationService()->getAllTypoScript();
+                    $configuration['defaultExtras'] = $typoScript['plugin']['tx_flux']['settings']['flexform']['rteDefaults'];
+                } else {
+                    $configuration['defaultExtras'] = $defaultExtras;
+                }
+            }
         } else {
-            $configuration['defaultExtras'] = $defaultExtras;
+            $configuration['defaultExtras'] = $this->getDefaultExtras();
         }
         $renderType = $this->getRenderType();
         if (false === empty($renderType)) {
             $configuration['renderType'] = $renderType;
             $configuration['format'] = $this->getFormat();
         }
+
+        //var_dump($configuration);die();
+
         return $configuration;
     }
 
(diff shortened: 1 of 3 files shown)


cweiske:
This is v8 only, since it ignores `defaultExtras`.

cweiske:
do not merge yet.

cweiske:
Works fine now.

cweiske:
I wonder why the config is in TypoScript and not PageTSConfig (was introduced that way in #1426).

NamelessCoder:
Thanks for pulling this one together.

> I wonder why the config is in TypoScript and not PageTSConfig

For legacy reasons, the default RTE config was added to TS since TS settings automatically transfer to templates. However, the only feature that actually requires this (excluding for a while the option to override Flux forms using TS) is this RTE configuration which needed a default value that can be applied universally - or template editors would have been required to write a pretty extensive `defaultExtras` statement and would be very likely to write it hardcoded into templates. So the TS was a compromise between "yep, you'd normally use pageTSconfig" and "pageTSconfig isn't available in Fluid templates".

> Should we fallback to PageTSConfig `RTE.default.preset`?

YES! Please do add this as the default source of preset name. And if there is a constant in PHP anywhere in TYPO3 which contains that `default` name, please use that instead of the new class constant in this PR.

Since rtehtmlarea is deprecated and scheduled for removal, the only component which made TS sensible will be gone, and I'm in favor of moving to pageTSconfig for this RTE stuff (and using existing options whenever possible).

> Should we remove `defaultExtras` completely (param def + getter/setter)?

Not yet and possibly not at all. If I'm not mistaken, there are other things this can be used for than merely RTE configuration. But we do need the new property `richtextConfiguration` and getters/setters.

(see also code comments)

Classes/Form/Field/Text.php
```diff
@@ -15,6 +15,7 @@
  */
 class Text extends Input implements FieldInterface
 {
+    const DEFAULT_RTE_CONFIG = 'default';
```
NamelessCoder: If we can find a constant in TYPO3 itself that has this default value, let's reference that instead of creating a new class constant here.

Classes/Form/Field/Text.php
```diff
@@ -39,6 +40,11 @@ class Text extends Input implements FieldInterface
     /**
      * @var string
      */
+    protected $richtextConfiguration;
```
NamelessCoder: This property could have the constant as default value, this would make `getRichtextConfiguration()` a bit smaller and faster for default usages.

Classes/Form/Field/Text.php
```diff
@@ -55,12 +61,9 @@ public function buildConfiguration()
         $configuration['rows'] = $this->getRows();
         $configuration['cols'] = $this->getColumns();
         $configuration['eval'] = $this->getValidate();
-        $defaultExtras = $this->getDefaultExtras();
-        if (true === $this->getEnableRichText() && true === empty($defaultExtras)) {
-            $typoScript = $this->getConfigurationService()->getAllTypoScript();
-            $configuration['defaultExtras'] = $typoScript['plugin']['tx_flux']['settings']['flexform']['rteDefaults'];
-        } else {
-            $configuration['defaultExtras'] = $defaultExtras;
+        if (true === $this->getEnableRichText()) {
```
NamelessCoder: We need a condition here which checks if either ckeditor or rtehtmlarea is installed and does the appropriate thing depending on that. We cannot yet remove the existing behavior, but it is perfectly fine to only trigger it if rtehtmlarea is installed.

The current TS usage can be marked deprecated, iirc it's added in ext_localconf.php.

Classes/ViewHelpers/Field/TextViewHelper.php
```diff
@@ -65,6 +65,15 @@ public function initializeArguments()
...
             ''
         );
+        $this->registerArgument(
+            'richtextConfiguration',
+            'string',
+            'Specifies which configuration to use in combination with EXT:rte_ckeditor.' .
+            'The default value is \'default\', since that\'s the standard configuration the CMS uses. ' .
+            'More information: https://docs.typo3.org/typo3cms/TCAReference/ColumnsConfig/Properties/TextRichtextConfiugration.html',
+            false,
+            null
```
NamelessCoder: The default value here should be the value of the constant (either the newly added class constant or the one TYPO3 contains, if it contains one).

cweiske:
> getRichtextConfiguration(), setter and property should be named richTextConfiguration for consistency with other methods.

The TYPO3 core code uses `richtextConfiguration`, so this is used here, too. We should instead rename the other config option from `enableRichText` to `enableRichtext`.

Classes/Form/Field/Text.php
cweiske: No, we have to keep that empty there, as well as in the `registerArgument` call - otherwise we'd never know that we should fall back to the TS value.

Classes/ViewHelpers/Field/TextViewHelper.php
cweiske: No, we have to keep that empty - otherwise we'd never know that we should fall back to the TS value.

cweiske:
I implemented a crude method of retrieving the current page ID. Not sure how viable that is; at least it works here(TM).

NamelessCoder:
Picked and adjusted in https://github.com/FluidTYPO3/flux/commit/00100eee7ea4c6386ec0ab297a5c23fef3689962. I chose to add it initially without further lookups for other pageTSconfig locations (since the normal table+field didn't really seem to cut it and no pragma exists to target flexform fields). Also, `defaultExtras` is history (soft deprecated) and so is the strategy of using regular TS to store this information.

NamelessCoder:
Kindly provided in https://github.com/FluidTYPO3/flux/pull/1438

mmalisz:
@cweiske getPageId method will throw "Unable to determine page ID" exception if RTE is placed in section and whole CE is not saved.

Code example:
https://pastebin.com/2g9E62CK

Issue #1444, opened 2017-07-27
Default language locallang file


ogrosko:
Isn't better file `locallang_db.xlf` instead of `locallang.xlf`?

NamelessCoder:
Extbase/Fluid's default is locallang.xlf and Flux is tightly integrated with those - this is why we use that filename as the default. Same as if you had used `f:translate` in a BE Fluid template ;)

ogrosko:
What about to configure this path eg. in plugin `extConf` ?

PR #1445, opened 2017-08-04
Feature/1444/locallang config

ogrosko:
[FEATURE] - #1444 Default locallang configured by extension configuration

Path to flux:form locallang file is no more fixed as `DEFAULT_LANGUAGEFILE` constant but configured in extension configuration

### DIFF
--- a/Classes/Form/AbstractFormComponent.php
+++ b/Classes/Form/AbstractFormComponent.php
@@ -23,6 +23,7 @@ use TYPO3\CMS\Core\Utility\GeneralUtility;
 use TYPO3\CMS\Extbase\Object\ObjectManager;
 use TYPO3\CMS\Extbase\Object\ObjectManagerInterface;
 use TYPO3\CMS\Extbase\Reflection\ObjectAccess;
+use TYPO3\CMS\Extensionmanager\Utility\ConfigurationUtility;
 
 /**
  * AbstractFormComponent
@@ -63,7 +64,7 @@ abstract class AbstractFormComponent implements FormInterface
      *
      * @var string
      */
-    protected $localLanguageFileRelativePath = Form::DEFAULT_LANGUAGEFILE;
+    protected $localLanguageFileRelativePath = '';
 
     /**
      * @var string
@@ -355,6 +356,15 @@ abstract class AbstractFormComponent implements FormInterface
      */
     public function getLocalLanguageFileRelativePath()
     {
+        if (!$this->localLanguageFileRelativePath) {
+            /** @var ConfigurationUtility $configurationUtility */
+            $configurationUtility = $this->getObjectManager()->get(ConfigurationUtility::class);
+            $extensionConfiguration = $configurationUtility->getCurrentConfiguration('flux');
+            return $extensionConfiguration['defaultLocallangFile']['value']
+                ? $extensionConfiguration['defaultLocallangFile']['value']
+                : $extensionConfiguration['defaultLocallangFile']['default_value'];
+        }
+
         return $this->localLanguageFileRelativePath;
     }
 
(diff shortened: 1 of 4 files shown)


deerfish:
Guten Tag
 
Vom 4. August bis und mit 18. August darf ich ein paar Ferientage geniessen, und bin deshalb im Büro nicht erreichbar.
 
Die Nachrichten werden in dieser Zeit nicht gelesen! In dringenden Angelegenheiten erreichen Sie mich zeitweise auf dem Mobile. Ansonsten bitte ich Sie eine Sprachnachricht zu hinterlassen, damit ich Sie zurückrufen kann. Vielen Dank!

Sonnige Grüsse

Damian Parpan



Damian Parpan
appline
Hirschengraben 52
6003 Luzern
Switzerland

T: +4141 266 0034
M: +4179 303 4878

www.appline.ch

NamelessCoder:
@deerfish you may want to add a rule for your auto-reply so it doesn't post your email and phone numbers here every time you go on vacation ;) thanks!

ext_conf_template.txt
```diff
@@ -9,3 +9,6 @@ listNestedContent = 0
 
   # cat=basic/enable; type=boolean; label=Error handling: If enabled, all Flux controllers are permitted to handle their own errors in ways that are far more flexible than traditional error catching. Enable with care - if your plugin, content or page template collection does not contain an Error.html template, your site may risk breaking with an extremely basic and not very informative error á la "Template not found for action foobar" but when enabled and combined with proper error templates you can customise pretty much any error which might happen anywhere when a Flux controller subclass is in play. Best turned off on development sites unless you are developing exactly error templates as it greatly limits debug output!
 handleErrors = 0
+
+  # cat=basic; type=string; label=Default locallang file: Relative (from extension Flux) path to locallang file containing labels for the LLL values built by FluidTYPO3\Flux\Form\AbstractFormComponent class.
+defaultLocallangFile = /Resources/Private/Language/locallang_db.xlf
```
NamelessCoder: Major breaking change if you change the default here, so please don't do that.

Classes/Form/AbstractFormComponent.php
```diff
@@ -355,6 +356,15 @@ public function setLocalLanguageFileRelativePath($localLanguageFileRelativePath)
      */
     public function getLocalLanguageFileRelativePath()
     {
+        if (!$this->localLanguageFileRelativePath) {
```
NamelessCoder: This method gets called extremely often (every time any label is built) so retrieving extension configuration at this point is not a good solution in terms of performance.

NamelessCoder:
@ogrosko thanks for the PR - we'll need to attack this a bit differently to make it less invasive, not global but rather local to each extension, and using a method that's more efficient than extension configuration.

Please consider the following though:

* The `_db` suffix is not conventional in Fluid nor in Extbase. FluidTYPO3 recommends overriding it only when you absolutely need to reuse labels from a different extension - for all other cases the conventions are recommended (and you can use shorthand references like `LLL:label_name` to access specific labels in Flux components).
* There are several ways of configuring the LLL file path to use, both from Fluid and by overriding Flux form properties through TS, or by creating your own overrides for the ViewHelpers or Form components (which would let you completely change all aspects of the label generation to your needs). Using these methods would be preferable and can also affect forms from third party Flux-based extensions.
* On TYPO3v8 it is extremely easy to extend a Fluid namespace with your own namespace location, which makes Fluid use ViewHelpers from that namespace if they exist. This way you can replace specific parts of Flux - for example to change default values of ViewHelper arguments.

This is just to mention the other options that exist and are usable on every attribute rather than just this one ;)

ogrosko:
Yes you are right, I will check that and make some improvements

ogrosko:
Ok I solved this with registering custom flux namespace

@NamelessCoder thank you for your support

Issue #1449, opened 2017-08-18
Asset loading


steffenk:
Hi,

on fresh TYPO3 8.7.4, flux with development branch, i got several exceptions because resources where not found (In BE rendering Flux elements).

After debugging i found this to be the correct way to include them, you see i the patch.
I didn't checked it with fluid_content, but i'm happy to use flux as standalone.
If you want i can create a pull request.

Patch:
   ```
$pageRenderer->addCssFile(
-                $doc->backPath . GeneralUtility::getFileAbsFileName('EXT:flux/Resources/Public/css/grid.css')
+                'EXT:flux/Resources/Public/css/grid.css'
             );

protected function attachAssets()
     {
-        $GLOBALS['TBE_STYLES']['stylesheet'] = GeneralUtility::getFileAbsFileName('EXT:flux/Resources/Public/css/icon.css');
+        $GLOBALS['TBE_STYLES']['stylesheet'] = 'EXT:flux/Resources/Public/css/icon.css';
     }

```

steffenk:
Pull request: #[1450](https://github.com/FluidTYPO3/flux/pull/1450)

NamelessCoder:
Pull request merged, closing.

PR #1450, opened 2017-08-18
#1449 Use correct Paths for Assets

steffenk:


NamelessCoder:
Can you give an example of a setup where this fails? Both of these have been tested and should be working - but perhaps there is an edge case where the abs paths / references don't work?

steffenk:
i have this problem in several local installations, all with TYPO3 v8.7.4 and flux, branch development.
If i do not use this patch, i always get exception in BE when calling page module:

> file_get_contents(/Users/steffen/Development/web/Sites/typo3/dev8/web/typo3//Users/steffen/Development/web/Sites/typo3/dev8/web/typo3conf/ext/flux/Resources/Public/css/grid.css): failed to open stream: No such file or directory

NamelessCoder:
Thanks Steffen :)

PR #1452, opened 2017-08-18
[BUGFIX] Use correct limit in statement

steffenk:


Classes/Service/RecordService.php
```diff
@@ -45,7 +45,13 @@ public function get($table, $fields, $clause = null, $groupBy = null, $orderBy =
             $statement->where($clause);
         }
         if ($limit) {
-            $statement->setMaxResults($limit);
+            if (is_int($limit)) {
+                $statement->setMaxResults($limit);
+            } else {
+                $limit =  explode(',', $limit);
```
NamelessCoder: Are you sure this is as it is intended and not something you assimilated from the EXT:news method of defining sortings and limits? The problem seems more to be that we don't have an offset argument.

steffenk: Have a look into FluidTYPO3\Flux\ViewHelpers\Content\GetViewHelper,
line 156:

```
 $rows = static::getRecordService()->get('tt_content', '*', $conditions, '', $order, $offset . ',' . $limit);
```

Smove:
This fix works for me

Tomowok:
You can split the current $limit parameter into 2 parameters $limit and $offset to follow the Doctrine function "setMaxResults" and "setFirstResult" (https://api.typo3.org/typo3cms/current/html/class_t_y_p_o3_1_1_c_m_s_1_1_core_1_1_database_1_1_query_1_1_query_builder.html#a6a0ecd5d39d5315f3d84e395e26442e8)
So, you can add this in RecordService.php :

```
if ($offset) {
	$statement->setFirstResult($offset);
}
```

And, of course, adapt the parameters in function get in "RecordService.php", "WorkspacesAwareRecordService.php" and "GetViewHelper.php".

NamelessCoder:
Replaced by 76d787eefe84ed6d32b380d99a624a974ef35c50 - sorry about the hijack Steffen but the other patch seemed more appropriate now that we're close to a 9.0 release and can accept minor breaking changes. Thanks for your work all the same! :)

PR #1456, opened 2017-09-02
[FEATURE] Split ProviderInterface

NamelessCoder:
In an effort to reduce the number of methods required by most Provider implementations and make Provider resolving more efficient, the ProviderInterface is now split into multiple smaller interfaces which can be implemented individually and which will be possible to restrict to while resolving Providers. For example, a Provider can be made which is solely responsible for returning a Form instance, rendering a preview, triggering on DataHandler commands and so on. Or any combination of those interfaces.

Additionally it is now possible to demand one or more interfaces when attempting to resolve Providers via the ProviderResolver or FluxService. Any provider that does not implement all of the demanded interfaces will be excluded even before `trigger()` is called and as such, can be used to limit the number of Providers that must be asked if they apply to the given input criteria. To illustrate, if your purpose is to read both a Grid and Form from the Provider then both the GridProviderInterface and FormProviderInterface can be required and you won't have to verify the returned Provider's interface.


NamelessCoder:
Manually merged - thanks for the review @steffenk :)

PR #1464, opened 2017-09-15
[BUGFIX] Add offset in RecordService->get

franzkugelmann:
Curently elements in containers are not rendered if you use flux:content.get due to different behaviour in new QueryBuilder since TYPO3 8 due to wrong offset usage.

Just saw that there is already a pull request regarding this problem:
https://github.com/FluidTYPO3/flux/pull/1452/
This new pull request is a slightly different and IMO somewhat cleaner approach.

### DIFF
--- a/Classes/Service/RecordService.php
+++ b/Classes/Service/RecordService.php
@@ -29,9 +29,10 @@ class RecordService implements SingletonInterface
      * @param string $groupBy
      * @param string $orderBy
      * @param integer $limit
+     * @param integer $offset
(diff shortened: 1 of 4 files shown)


franzkugelmann:
Hm, it seems i would need to adapt some tests as well. Right now i don't get why the setMaxResults check fails.

NamelessCoder:
Manually merged in 76d787eefe84ed6d32b380d99a624a974ef35c50 - thanks for taking care of this one Franz!

PR #1469, opened 2017-10-19
[REMOVAL] Remove outdated 7.6-specific condition

jmverges:
https://github.com/FluidTYPO3/fluidpages/issues/368


Classes/Backend/TceMain.php
```diff
@@ -780,7 +780,8 @@ protected function getOriginalRecordUid($table, $uid)
             't3ver_move_id',
             sprintf(
                 ' %s t3ver_state = 3 AND deleted = 0 AND uid = %d',
-                version_compare(ExtensionManagementUtility::getExtensionVersion('workspaces'), '8.0.0', '<') ? 'AND' : '',
+                (ExtensionManagementUtility::isLoaded('workspaces') && version_compare(ExtensionManagementUtility::getExtensionVersion('workspaces'),
```
NamelessCoder: We can actually remove this now that we no longer support TYPO3 7.6 - the condition that gets generated is okay in all supported TYPO3 versions without adding the ` AND `.

PR #1473, opened 2017-10-30
[BUGFIX] find child records also with "insertRecord"

franzkugelmann:
If a fluidcontent element containing other content elements is added as "insertRecord" into a page, then currently the child records are not found anymore, because the pid is wrong. TYPO3 used by default the pid of the currently rendered page, but if you used "insertRecord" for your container element, then the pid for the childs records will be different, they always match the pid of the container element.


NamelessCoder:
Seems quite reasonable. You wouldn't be able to add child content from other pages anyway!

PR #1475, opened 2017-11-02
[TASK] Remove fluidcontent from composer suggestions

jdreesen:
fluidcontent is considered obsolete (see: https://github.com/FluidTYPO3/fluidcontent#obsolete) so there's no reason to suggest it as a dependency anymore.


NamelessCoder:
Good catch, this isn't exactly a place we look at often :) thanks for the patch and welcome to the list of contributors in this repo as well! 🎉

PR #1476, opened 2017-11-02
[BUGFIX] Brings back Inline and Relation fields expected behaviour

jmverges:
After https://github.com/FluidTYPO3/flux/commit/0bd5466f85b0ad9962803a396d7a9ac624c65cc4
field.inline and field.relation are not working as expected

### DIFF
--- a/Classes/Form/Field/Inline.php
+++ b/Classes/Form/Field/Inline.php
@@ -17,6 +17,11 @@ class Inline extends AbstractInlineFormField
 {
 
     /**
+     * @var string|null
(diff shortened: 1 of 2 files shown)


Classes/Form/Field/Inline.php
```diff
@@ -17,6 +17,11 @@ class Inline extends AbstractInlineFormField
 {
 
     /**
+     * @var null
```
NamelessCoder: `@var string|null` please - also for the other file.

PR #1479, opened 2017-11-07
[BUGFIX] Fixing error issue when editing page properties

zoranilic:
According to RenderType description:
https://docs.typo3.org/typo3cms/TCAReference/ColumnsConfig/Type/Group.html#rendertype-default

### DIFF
--- a/Classes/Form/Field/File.php
+++ b/Classes/Form/Field/File.php
@@ -53,6 +53,11 @@ class File extends AbstractMultiValueFormField
     protected $internalType = 'file_reference';
 
     /**
+     * @var string|null
+     */
+    protected $renderType = null;
+
+    /**
      * @return array
      */
     public function buildConfiguration()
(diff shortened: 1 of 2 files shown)


NamelessCoder:
Thanks for the patch @zoranilic - just for the record, it appears you're not responsible for the unit tests failing on dev-master. I'll have to fix those and re-run the tests. Change itself looks good :)

zoranilic:
One more place for renderingType to be set to NULL. It's "Fal" object.

NamelessCoder:
Oops! I forgot we had this pull request open - so I went and pushed a change to nullify renderType on the File component. Sorry to remove that from your patch but in order to resolve the conflict it had to be done.

NamelessCoder:
Thank you for the patch - and welcome to the list of contributors! :tada

zoranilic:
My pleasure. ;)

PR #1487, opened 2017-12-04
[BUGFIX] Use another alias for class "Object"

### LINKED ISSUE #1486: PHP 7.2 compatibility
I've tried to run Flux with PHP 7.2.0 but the following error occurred:

`Fatal error: Cannot use FluidTYPO3\Flux\Form\Container\Object as Object because 'Object' is a special class name in typo3conf/ext/flux/Classes/Form/AbstractFormComponent.php on line 15`

Lagerregal:
Fixes #1486

### DIFF
--- a/Classes/Form/AbstractFormComponent.php
+++ b/Classes/Form/AbstractFormComponent.php
@@ -12,7 +12,7 @@ use FluidTYPO3\Flux\Form;
 use FluidTYPO3\Flux\Form\Container\Column;
 use FluidTYPO3\Flux\Form\Container\Container;
 use FluidTYPO3\Flux\Form\Container\Grid;
-use FluidTYPO3\Flux\Form\Container\Object;
+use FluidTYPO3\Flux\Form\Container\ObjectContainer;
 use FluidTYPO3\Flux\Form\Container\Section;
 use FluidTYPO3\Flux\Form\Container\Sheet;
 use FluidTYPO3\Flux\Hooks\HookHandler;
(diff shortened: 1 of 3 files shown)


Classes/Form/AbstractFormComponent.php
```diff
@@ -12,7 +12,7 @@
 use FluidTYPO3\Flux\Form\Container\Column;
 use FluidTYPO3\Flux\Form\Container\Container;
 use FluidTYPO3\Flux\Form\Container\Grid;
-use FluidTYPO3\Flux\Form\Container\Object;
+use FluidTYPO3\Flux\Form\Container\ObjectContainer;
```
cedricziel: I would suggest using `Object as ObjectContainer` and we need the very same change in the test.

Lagerregal: oh, my fault... of course!

Lagerregal:
It seems that using aliases does not work in PHP 7.2 ... https://travis-ci.org/FluidTYPO3/flux/jobs/311287117#L589
We have to rename the class.

cedricziel:
Yep. And then class_alias it on php versions < 7.2. Seems doable. Go ahead :)

NamelessCoder:
Just restoring the diff to not add 7.2 tests. We will need to attack this one differently - by renaming the entire class.

NamelessCoder:
I took the liberty of doing the renaming patch since there were a few string references to `'Object'` that would not have been so easy to locate. Thanks for opening this and doing the research, @Lagerregal!

kthalmann:
Could you apply this fix in 8.2? I am still using flux 8.2.1 with typo3 7.6.32 and would like to use PHP 7.2

NamelessCoder:
@kthalmann 8.2 is no longer maintained, sorry. Our minimum supported TYPO3 LTS version is 8.7.

PR #1489, opened 2017-12-04
Wrong renderType "selectSingle" was set for FAL Selector, which broke…

warki:
In TYPO3 8.7 the FalViewHelper leads to an error, because a wrong Render Type has been set.


NamelessCoder:
Already taken care of in https://github.com/FluidTYPO3/flux/commit/854f5be1e1b2ce4814ea5a6f1c6fe94c2c034397 - when you create new patches, make sure you start from `development` branch (and it helps us a lot if you also test for the bug on `development` before submitting patches) ;)

Sorry to close your first Flux pull request - maybe even your first ever GitHub pull request! - right away like this, but the problem is already fixed. Don't let that discourage you from creating new pull requests in the future; this time it was fixed but the next one you make could be a critical fix!

PR #1492, opened 2017-12-06
Avoid appending of SubControllerAction

M-a-r-n-o:


NamelessCoder:
Hey @M-a-r-n-o - thanks for the patch! Can you give us a bit of background information why this is necessary? Normally you would not have anything in `$this->response` unless you put content there from a point in the controller where output should not happen (e.g. one of the initialize methods).

M-a-r-n-o:
Hey @NamelessCoder, 
My Problem was, that i made a function inside my PageController (extending \FluidTYPO3\Fluidpages\Controller\PageController) to assign some settings to the page template and there - to my suprise - the content rendered twice. I couldn't find any duplicate renderings inside my TS. So i tracked down the Problem to the "callSubControllerAction" function which seems to hold a previous rendering inside "$response". Maybe it is a problem with my setup or a "Core feature". But the patch seems to solve the issue without any side effects. 

Best wishes, 

P.s
My Setup:
Typo 8.7.8
PHP 7.1.2
Flux 8.2.1
Fluid Pages 4.1.0

jsegars:
I've run into the same issue @M-a-r-n-o described with duplicate rendering and the patch seemed to solve the issue with no side effects in my testing.

NamelessCoder:
Picked manually as https://github.com/FluidTYPO3/flux/commit/22766cdd527a9b1e40357b098723f5058b6b3478 - and I adjusted the tests so they test the additional instance creation.

Issue #1484, opened 2017-12-01
"Show hidden content elements" in Page module doesn't show hidden elements which are inside flux container content element


zoranilic:
I have container content element with grid columns used. When I add other flux content element inside in the grid it is visible in back-end preview. When I hide it with "hidden" property it disappears from back-end preview in Page module but is present in List module as hidden. This happens even if "Show hidden content elements" is checked. 
Furthermore, I can see, under brackets, how many elements are hidden on the page: "Show hidden content elements (1)".
This is for inner content elements either from typo3 core (text, form...) or made by flux inside container content element. 
I am not using "fluidcontent" extension any more.
I checked currently submitted issues but couldn't  find the same issue.

typo3: 8.7.8
flux: 8.2.1(development branch)
fluidpages: 4.1.0 (development branch)

NamelessCoder:
You can only activate listing of nested elements via the Flux extension settings in extension manager. Beware that this may cause sorting problems if you then sort tt_content records in list view!

zoranilic:
I think you didn't understand me right. Confirm if I'm wrong. 
Check the fallowing 2 screenshots. First is preview in Page module and the second in List module. In the list module you will see a hidden content element which is not displayed in Page module inside container below "Form" content element. It was displayed there when it is not hidden. This regards any content element inside container content element.

![page-module](https://user-images.githubusercontent.com/7780949/33504927-bb732be2-d6e9-11e7-9477-7a50546de15e.png)

![list-module](https://user-images.githubusercontent.com/7780949/33504930-bfa96618-d6e9-11e7-8437-1fdeca02740a.png)

zoranilic:
I will look into this as soon as I have some time.

Apen:
There is a fast fix for this.

typo3conf/ext/flux/Classes/Service/RecordService.php @ public function get()

```
...
        if ($offset) {
            $statement->setFirstResult($offset);
        }

        $moduleData = $GLOBALS['BE_USER']->getModuleData('web_layout', '');
        if (!empty($moduleData['tt_content_showHidden'])) {
            $statement->getRestrictions()->removeByType(\TYPO3\CMS\Core\Database\Query\Restriction\HiddenRestriction::class);
        }

        return $statement->execute()->fetchAll();
...
```

NamelessCoder:
Solved as part of https://github.com/FluidTYPO3/flux/pull/1563, will be included in release 9.0. Please see the PR for more details, among other things about how to use the migration script.

PR #1496, opened 2017-12-22
[BUGFIX] "Show hidden content elements" in Page module doesn't show hidden elements which are inside flux container content element #1484

Apen:


Classes/Service/RecordService.php
```diff
@@ -51,6 +51,13 @@ public function get($table, $fields, $clause = null, $groupBy = null, $orderBy =
         if ($offset) {
             $statement->setFirstResult($offset);
         }
+        
+        if (TYPO3_MODE === 'BE') {
```
NamelessCoder: IMHO this really doesn't belong in RecordService, which is supposed to be a generic service that doesn't depend on state. If there is no alternative, we'll need to reproduce this query in Preview instead of using RecordService.

Apen: Yes, it's not possible to edit the statement outside RecordService. Will see that soon.

NamelessCoder:

PR #1503, opened 2018-01-14
[BUGFIX] Using EmptyOption for icontables

monosize:
It is not yet possible to use the empty option with the argument showIconTable to display a symbol for the empty value.


Classes/Form/AbstractMultiValueFormField.php
```diff
@@ -79,7 +79,7 @@
      * If not-FALSE, adds one empty option/value pair to the generated selector
      * box and tries to use this property's value (cast to string) as label.
      *
-     * @var boolean|string
+     * @var boolean|mixed
```
xf-: mixed includes boolean

PR #1504, opened 2018-01-15
[BUGFIX] Set flux colPos for moved records only if correct

franzkugelmann:
It seems that \FluidTYPO3\Flux\Service\ContentService::moveRecord is surprisingly called with target pid in parameter $relativeTo if editor is cutting and inserting a content record via clipboard. Only on big pagetrees this leads to lost content elements, because colPos is set to 18181 then even for regular content elements.
Since the flux colPos only makes sense, if we have a tx_flux_parent, this pull request fixes that issue.

I am fully aware that this fix is likely not targeting the root cause of the problem (getting the pid for the parameter $relativeTo) but i was not able to fully grasp what should be going on here.
Thus i hope this helps at least as hotfix for people affected by this problem.

### DIFF
--- a/Classes/Service/ContentService.php
+++ b/Classes/Service/ContentService.php
@@ -153,12 +153,14 @@ class ContentService implements SingletonInterface
      */
     public function moveRecord(array &$row, &$relativeTo, $parameters, DataHandler $tceMain)
     {
+        // This condition matches colPos values which are above the constant which is a virtual, assumed limit. When
+        // the $relativeTo value is above this constant value it is likely a virtual value - but it may also be the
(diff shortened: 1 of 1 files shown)


NamelessCoder:
@franzkugelmann Thanks for opening this one - I've reworked the patch a bit so the logic instead is to not assign `colPos` in the various conditions except in the one where you move relative to a record. The last lines of code that execute in this function are:

```php
if (0 < $row['tx_flux_parent']) {
    $row['colPos'] = static::COLPOS_FLUXCONTENT;
}
```

So removing the assignment means the `colPos` should be adjusted as the final action and will only be based on whether `tx_flux_parent` is set.

Issue #1516, opened 2018-02-08
RTE.default.preset ignored


glucka:
This works:
<flux:field.text name="tt_content.bodytext" label="Text" enableRichText="1" richtextConfiguration="mycustomtheme"/>

But RTE.default.preset config does not.

rupasix:
This should help: #1438

spackmat:
I came across this problem with flux 9.0.1, where the `RTE.default.preset` is still ignored. [I added a comment to the code](https://github.com/FluidTYPO3/flux/pull/1538/files#r206082162) in #1538 with further information. The fix there doesn't work in my case.

PR #1518, opened 2018-02-12
[BUGFIX] PageTSConfig RTE.default.preset ignored

glucka:
Solved #1516

### DIFF
--- a/Classes/Form/Field/Text.php
+++ b/Classes/Form/Field/Text.php
@@ -8,11 +8,9 @@ namespace FluidTYPO3\Flux\Form\Field;
  * LICENSE.md file that was distributed with this source code.
  */
 
-use FluidTYPO3\Flux\Configuration\BackendConfigurationManager;
 use FluidTYPO3\Flux\Form;
 use FluidTYPO3\Flux\Form\FieldInterface;
 use TYPO3\CMS\Backend\Utility\BackendUtility;
-use TYPO3\CMS\Extbase\Configuration\ConfigurationManagerInterface;
 
 /**
  * Text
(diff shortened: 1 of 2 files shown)


NamelessCoder:
Hi @glucka - if I read this right, the change has no effect. The instance that is checked for no longer exists so the code that is removed, would not be executed anyway (except for the creation of objectmanager). Removing the code is good - but does it really solve a problem with pageTSconfig presets?

glucka:
Hi @NamelessCoder - the solution is in the FormViewHelper.php file.
Now $root->getOption('record')['pid'] is always exist and BackendUtility::getPagesTSconfig($pageUid) works too.

glucka:
hi @NamelessCoder do you need more detailed feedback?

Classes/Form/Field/Text.php
```diff
@@ -202,13 +200,8 @@ public function getRichtextConfiguration()
...
     {
-        $configurationManager = $this->getObjectManager()->get(ConfigurationManagerInterface::class);
-        if ($configurationManager instanceof BackendConfigurationManager) {
-            $pageUid = $configurationManager->getCurrentPageId();
-        } else {
-            $root = $this->getRoot();
-            $pageUid = $root instanceof Form ? $root->getOption('record')['pid'] ?? 0 : 0;
-        }
+        $root = $this->getRoot();
+        $pageUid = $root instanceof Form ? $root->getOption('record')['pid'] : 0;
```
pkerling: This has to be`uid` instead of `pid`, since we're already looking at a page. I have the Page TSConfig defined on page 1 (site root), and with `pid` it will look for the TSConfig on page 0 (tree root) when editing the site root page, which is wrong.

pkerling:
Works for me (in conjunction with the other fix I posted)

Classes/Form/Field/Text.php
pkerling: Wait, this also applies to flexforms in content elements, right? So it could be both depending on the context... :-)

glucka: Yes, it could be both depending on the context...
In most cases it is a content element, but yes it could be a Page Configuration as richtext too.

glucka:
hi @pkerling should i clean up my pull request to one commit?

pkerling:
Don't know why you're asking me, but I'd assume so :-) You might also want to follow the [contribution guide](https://fluidtypo3.org/documentation/contributing/contribution-guide.html) regarding the commit message.

glucka:
Thanks for the link

PR #1519, opened 2018-02-14
[BUGFIX] Set input field range independent of validators

pkerling:
Checking for 'int' in the validator list is needlessly
limiting. range works just as well e.g. with double2 fields.


NamelessCoder:
Merged w/ two of five build errors - they're not your fault and the three that matter are green. Thanks for the perfectly formed pull request @pkerling ;)

Issue #1522, opened 2018-02-22
flux:form.text should support placeholder attribute


luzat:
According to the TCA documentation placeholder is available for `type='text'` and `type='input'`, but only implemented for `input` in flux.

PR #1523, opened 2018-02-22
[FEATURE] Add placeholder attribute to TextViewHelper.

luzat:
flux:field.text gains support for placeholder, similar to flux:field.input.

### DIFF
--- a/Classes/Form/Field/Text.php
+++ b/Classes/Form/Field/Text.php
@@ -55,6 +55,11 @@ class Text extends Input implements FieldInterface
     protected $format;
 
     /**
+     * @var string
+     */
+    protected $placeholder;
+
+    /**
      * @return array
      */
     public function buildConfiguration()
@@ -63,6 +68,7 @@ class Text extends Input implements FieldInterface
         $configuration['rows'] = $this->getRows();
         $configuration['cols'] = $this->getColumns();
         $configuration['eval'] = $this->getValidate();
+        $configuration['placeholder'] = $this->getPlaceholder();
         if (true === $this->getEnableRichText()) {
             $configuration['enableRichtext'] = true;
             $configuration['richtextConfiguration'] = $this->getRichtextConfiguration();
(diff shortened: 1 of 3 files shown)


luzat:
Should close #1522

NamelessCoder:
Perfect patch, just the way we like it :) the only possible improvement would be to add a test case that uses the property, but that's not required.

Test failures are not your fault.

luzat:
I have looked into writing a test, but the best I came up with is very similar to the tests from `FalViewHelperTest.php`, which only check if the component has its placeholder correctly set. I have also tried `$output = $this->executeViewHelper($arguments);` to check the output, but that returned `null`. Does this require further setup and would it be preferable? I only later found the `Form` tests, such as `TextTest.php`, but am not sure what the preferred way to test this would be. Support in `$chainedProperties` could be tested. Not so much wondering just for this little change, but also for possible further contributions.

NamelessCoder:
@luzat Sorry about the delay, I've been meaning to write a proper explanation for these tests, but then other projects kept me away.

Using the `->executeViewHelper` method makes it easy to test ViewHelpers which actually return something - which isn't the case for this particular group of ViewHelpers. Instead of returning anything (they return null) they build a structure of PHP objects that is then kept in the ViewHelperVariableContainer. And making assertions on that isn't as easy... testing the product of `getComponent` when given various argument combinations is the right way to test these, so you picked the right option.

Generally speaking, full coverage requires a test both for the ViewHelper and the component it creates - so feel free to add `placeholder` to the arguments that get tested in `TextTest.php` (since it has no conditions it's fine to add in the default chain properties array to test the getter and setter behavior, https://github.com/FluidTYPO3/flux/blob/development/Tests/Unit/Form/Field/TextTest.php#L19). So the only way this test could be more explicit was if you added the property to `$chainProperties` in the component test class. That ensures the component is covered completely including the return type of the setter method.

And had this addition also included conditions that for example only assigned placeholder if default was not given (don't make that change, just an example) then it might be appropriate to add dedicated testing methods that use a `@dataProvider` to run multiple value combinations and compare to an expected result for each combination (lots of tests use this practice and it's not too hard to understand, just arrays becoming arguments).

If you run into more complex situations you need to write tests for you are always welcome to ask before starting. But this one you did here - almost perfect. And your comment demonstrates you're definitely able to understand the logic of the unit tests, maybe with an explanation or two to speed up the info digging ;)

luzat:
Thanks for the explanations. I will try to look into some more of the code the next weeks after some deadlines are over. Is there any action that I should take regarding this PR? Add the line to `$chainProperties` (and `InlineTest`, too)? Leave the other test in? Squash? Rebase? :-)

PR #1525, opened 2018-02-26
[BUGFIX] Fix FluxService->getTypoScriptByPath not working 

pkerling:
Problem should be obvious, basically `getTypoScriptByPath` will not work at all at the moment.


pkerling:
Travis errors look unrelated.

PR #1526, opened 2018-02-28
[BUGFIX] Actually apply template paths from TypoScript

pkerling:
Previously, template paths for the extension the template belonged to
that were set up in TypoScript would be ignored.
For page templates for example, only `fluidpages` and the default
directories of the extension were used.


NamelessCoder:
Fixed via https://github.com/FluidTYPO3/flux/commit/e1880b75a61614be8c9f7b283c9af97a1cbb641e by moving the extension scoping into initializeView and giving the rendering context a fresh template paths instance initialized with the provider extension key.

Issue #1530, opened 2018-03-07
gridelements + flux - after d&d colpos is set to null (0)


steffenk:
This issue drove me crazy, as i thought this is a bug in gridelements.

gridelements use colpos = -1 for their container elements. If i d&d an element inside gridelements container, there are a few hooks involved, so also the one from flux. After debugging, i found the error in FluidTYPO3\Flux\Service\ContentService->moveRecord
The line
`$relativeToRecord = $this->loadRecordFromDatabase($relativeUid);`
may return null, which leads to wrong settings e.g. 
`$row['colPos'] = $relativeToRecord['colPos'];`

The fix is easy, as there should be a condition surrounded:
`if (NULL !== $relativeToRecord) { ...`

I can provide a patch next days.

tim-wendel:
Thanks to steffenk for debugging and fixing this annoying bug.

But there's still an issue if element is dropped inside a nested gridelement. $relativeTo will then result in "-1" which will end up in loading a wrong $relativeToRecord and getting an inappropriate colPos from $relativeToRecord.

steffenk:
question: do we need setting colpos at all? Is there a valid value beside -1?

tim-wendel:
Yes. In fact, you need to set the colPos when moving an element outside of a gridelement.

steffenk:
but this is done by the core ;)

steffenk:
it's bizarre - i created a test case (see screenshot) with nested gridelements container and one flux container. Now started d&d inside/outside the containers, where i first fixed #1536 to be able to continue.
And right, if i drag a CE into the flux container, $relativeTo is -1 which leads to wrong handling. Still some fixes to go ....
![bildschirmfoto 2018-03-22 um 13 31 41](https://user-images.githubusercontent.com/234874/37773307-8a140b20-2ddd-11e8-8734-2a1741ecc554.png)

NamelessCoder:
Solved as part of https://github.com/FluidTYPO3/flux/pull/1563, will be included in release 9.0. Please see the PR for more details, among other things about how to use the migration script.

PR #1533, opened 2018-03-16
#1530: set colPos only if relativeToRecord exists

steffenk:
Patch for Issue #1530


steffenk:
I don't know why it's conflicting, i used latest development branch in fork.

wazum:
The conflict is because the file in development looks different then the one your fork is based on (I tried to merge it into my fork, which failed too). Please double-check using the right branch and latest updates.

steffenk:
hm - i do not get it. In my fork i synced with flux repo and created then a new branch for the patch.

Issue #1534, opened 2018-03-22
\FluidTYPO3\Flux\Service\ContentService::moveRecord still wrong when moving to first position


wazum:
`moveRecord` sets the colPos to `COLPOS_FLUXCONTENT` in any case, even if the element is not in a flux container when moving records to the first position of an area.
The solution is easy, just remove line 175 with
`$row['colPos'] = static::COLPOS_FLUXCONTENT;`
as setting the colPos (if required) is already handled further down with
```
if (0 < $row['tx_flux_parent']) {
     $row['colPos'] = static::COLPOS_FLUXCONTENT;
}
```

FailXSiegen:
Does not work for me. Tried Drag&Drop Element from flux parent to normal colPos and it vanishes, fixed wrong colPos entry, then tried Drag&Drop Element back into flux parent, vanishes again, fixed colPos again.

TYPO3 8.7.8
flux 8.2.1
no fluidcontent

wazum:
@FailXSiegen Flux 8.2.1 is nearly a year old(!) 
If you really want to use flux, use the development branch. 
Unfortunately there are too few stable releases here.

FailXSiegen:
@wazum Have you looked at the version number in the development branch? Look again please.

wazum:
@FailXSiegen 
So you used the latest development branch for testing?
For some reason you call it 8.2.1 (when I talk about a specific version I talk about a tag: https://github.com/FluidTYPO3/flux/tree/8.2.1 and not some string value that has no relevance)?
Because there's a huge difference between those two … ;-) 
Did you test with my pull request (diff) or simply deleted the line with `$row['colPos'] = static::COLPOS_FLUXCONTENT;`?

FailXSiegen:
@wazum 
I just pulled the latest development branch and tried your fix. Perhaps i mean another Drag/Drop feature, which is broken with this fix.
I deleted the line 175 as you suggest.
I have a flux parent element with 1 child. If i drag/drop the child element to a normal colPos, the colPos is still 18181. 
If I try to drag/drop an element into the parent with the fix, the colPos changes to 20784 an the parent has no visible relations.
If I do this same procedure without the fix, it still doesn't work, but the colPos is correct.

Perhaps you fixed another type of issue inside of flux parent elements and this might work inside of one but the drag/drop feature with different colPos doesn't work at all.

wazum:
[...]

PR #1535, opened 2018-03-22
[BUGFIX] Remove setting colPos when moving record to first position

wazum:
Fixes issue #1534


NamelessCoder:
Fixed via https://github.com/FluidTYPO3/flux/pull/1504 - thanks for opening the pull request @wazum and sorry this did not become your first contribution; it made more sense to me to combine this with the other colPos patch. Keep them coming ;)

Issue #1536, opened 2018-03-22
getOriginalRecordUid uses wrong SQL when workspaces is not installed


steffenk:
There is no test for installed workspaces, this leads to wrong sql:
`version_compare(ExtensionManagementUtility::getExtensionVersion('workspaces'), '8.0.0', '<') ? 'AND' : ''`

If workspaces is not installed, getExtensionVersion returns '', so version_compare fails and the true-part is used, which leads to SQL:
` WHERE AND t3ver_state = 3 AND deleted = 0 AND uid =...`
You can see this in BE while d&d elements inside containers, the ajax dies with exception and the frame is not refreshed.

steffenk:
it seemed that the development branch of my fork was not updated correct, so the issue is outdated. Closing this.

PR #1537, opened 2018-03-22
#1536: check for workspaces is installed

steffenk:


steffenk:
why the heck is there a conflict? I do not get it.

PR #1538, opened 2018-03-23
[BUGFIX] PageTSConfig RTE.default.preset ignored

glucka:
Solves: #1516

### DIFF
--- a/Classes/Form/Field/Text.php
+++ b/Classes/Form/Field/Text.php
@@ -8,11 +8,9 @@ namespace FluidTYPO3\Flux\Form\Field;
  * LICENSE.md file that was distributed with this source code.
  */
 
-use FluidTYPO3\Flux\Configuration\BackendConfigurationManager;
 use FluidTYPO3\Flux\Form;
 use FluidTYPO3\Flux\Form\FieldInterface;
 use TYPO3\CMS\Backend\Utility\BackendUtility;
-use TYPO3\CMS\Extbase\Configuration\ConfigurationManagerInterface;
 
 /**
  * Text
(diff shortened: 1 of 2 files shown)


NamelessCoder:
@glucka I'm not quite sure what the intention is with this patch? The "record" variable should be assigned when you resolve the Form instance via the API - which is why we haven't assigned it in the FormViewHelper. The following is not so good about this patch:

* The "record" is implicitly dynamic when resolving the form and should never be assigned via the ViewHelper. This will, for example, break if you combine it with "static" option to cache the form DS.
* The removal of using ConfigurationManager is technically correct to do, but there's no explanation about why it is done.

So there's a bit too much breaking to accept this patch as it is. I think the right question to ask here is: why, in your use case, is the `$pageUid` not the right value? This is as far as I can see the core of the problem you're trying to solve with this patch - but it should not be possible to resolve a Form without also having the "record" variable set within it when you use the API.

NamelessCoder:
@glucka I've changed the patch so that pageTSconfig will always be resolved even if `$pageUid` is zero, and removed the record assignment from the ViewHelper. This should cover more cases and I expect, yours too. But if you still experience problems when you test this next time, please open an issue that describes that "record is not available when reading Form" and if possible include a stack trace and your entire `flux:form` definition.

glucka:
Thank you for the detailed explanation!

Classes/Form/Field/Text.php
```diff
@@ -226,18 +225,10 @@ public function getRichtextConfiguration()
...
     {
-        $configurationManager = $this->getObjectManager()->get(ConfigurationManagerInterface::class);
-        if ($configurationManager instanceof BackendConfigurationManager) {
-            $pageUid = $configurationManager->getCurrentPageId();
-        } else {
-            $root = $this->getRoot();
-            $pageUid = $root instanceof Form ? $root->getOption('record')['pid'] ?? 0 : 0;
-        }
+        $root = $this->getRoot();
+        $pageUid = $root instanceof Form ? $root->getOption('record')['pid'] ?? 0 : 0;
```
spackmat: `$root->getOption('record')` is NULL with flux 9.0.1, dumping `$root` gives me a Form-object with no `record` within the options path. So `$pageUid` gets always `0`, and the `RTE.default.preset` of the first page doesn't get loaded.

Issue #1514, opened 2018-02-04
Drag and Drop in development-branch still broken


SchuBu:
Hey there,
creating an CE within a grid element is possible without problems. The tx_flux_column and tx_flux_parent column in database is correctly filled.

BUT:
Dragging within a grid element (e.q. from named "col1" into "col2") results in disappearing of the CE. The database columns tx_flux_column and tx_flux_parent are filled with "NULL". So there is still a problem.

I'm using the flux development branch. TYPO3 at Version 8.7.9

Any ideas how to fix it? Here a video of this bug:
https://youtu.be/gDqK5WiA77E

SchuBu:
I've just tested a bit around:
I only can reproduce the problem when I'm using this extensions directly from GitHub (dev branch):
* flux
* fluidpages
* fluidcontent
* vhs

If I take them from the TYPO Extension Repository there is a warning about the fluidpages needed TYPO3 Version (7.6 needed, I'm using 8.7.9) but if you ignore this, it's just working. The Drag & Drop thing also...

NamelessCoder:
Most likely a duplicate of https://github.com/FluidTYPO3/flux/issues/1502, please check and confirm.

SchuBu:
I don't think so ... it is a problem for all columns. every time you move a content element (also from within a gridelement to outside "normal" page). please watch video above. Thx.

NamelessCoder:
Did you check if the solution found in https://github.com/FluidTYPO3/fluidcontent/issues/418 applies to your case?

glucka:
I think this is not related to https://github.com/FluidTYPO3/fluidcontent/issues/418

typo3 8.7.10
latest dev flux
latest dev fluidpage
without fluidcontent

And i have the same problem.

SchuBu:
@NamelessCoder: this solution did not work for me. I think @glucka is right, it is not related to fluidcontent

NamelessCoder:
> "tx_flux_parent, tx_flux_column" exist in TCA

Yes, but are they also in TCA showitem for your specific content type? That's the thing that was solved in the issue I linked to: adding the fields specifically for your (custom?) CType values is required.

It is not related to fluidcontent. But the solution found in that issue applies to Flux as well.

glucka:
Yes they are also in TCA showitem of my custom CType

glucka:
This is the showitem of my custom content element:
[...]

PR #1545, opened 2018-04-16
[BUGFIX] Drag and Drop

glucka:
Solve #1514

### DIFF
--- a/Classes/Backend/TceMain.php
+++ b/Classes/Backend/TceMain.php
@@ -492,6 +492,7 @@ class TceMain
 
         // Move the immediate record, which may itself be a placeholder or an original record.
         $row['uid'] = $uid;
+        $row['colPos'] = $newColumnNumber;
         $target = [$table, $destPid];
         $this->contentService->moveRecord($row, $destPid, $target, $reference);
         $this->recordService->update($table, $row);
@@ -535,9 +536,10 @@ class TceMain
         // Following block takes care of updating the immediate record, be that a placeholder, an
         // original or a versioned copy.
         $moveData = $this->getMoveData();
-        if (!$moveData) {
-            return;
-        }
+
+        $resolveUid = $this->getOriginalRecordUid($table, $uid);
+        $updateFields['colPos'] = GeneralUtility::_GET('data')[$table][$resolveUid]['colPos'];
+
         $updateFields['uid'] = $uid;
 
         $this->contentService->moveRecord($updateFields, $origDestPid, $moveData, $reference);
--- a/Classes/Backend/TceMain.php
+++ b/Classes/Backend/TceMain.php
@@ -268,7 +268,7 @@ class TceMain
         $record = $this->resolveRecordForOperation($table, $id);
 
         if ($table === 'tt_content') {
-            if ('localize' === $command) {
+            if ('localize' === $command || 'move' === $command) {
                 $this->contentService->fixPositionInLocalization($id, $relativeTo, $record, $reference);
             }
 
--- a/Classes/Service/ContentService.php
+++ b/Classes/Service/ContentService.php
@@ -365,7 +365,6 @@ class ContentService implements SingletonInterface
      */
     public function fixPositionInLocalization($uid, $languageUid, &$sourceRecord, DataHandler $reference)
     {
-        $previousLocalizedRecordUid = $this->getPreviousLocalizedRecordUid($uid, $languageUid, $reference);
         if (!empty($sourceRecord['l18n_parent'])) {
             $defaultRecordUid = $sourceRecord['l18n_parent'];
         } else {
@@ -373,22 +372,15 @@ class ContentService implements SingletonInterface
         }
         $localizedRecord = BackendUtility::getRecordLocalization('tt_content', $defaultRecordUid, $languageUid);
         $sortingRow = $GLOBALS['TCA']['tt_content']['ctrl']['sortby'];
-        if (null === $previousLocalizedRecordUid) {
-            // moving to first position in tx_flux_column
-            $localizedRecord[0][$sortingRow] = $reference->getSortNumber(
-                'tt_content',
-                0,
-                $sourceRecord['pid']
-            );
-        } else {
-            $localizedRecord[0][$sortingRow] = $reference->resorting(
-                'tt_content',
-                $sourceRecord['pid'],
-                $sortingRow,
-                $previousLocalizedRecordUid
-            );
-        }
-        $this->updateRecordInDataMap($localizedRecord[0], null, $reference);
+
+        $localizedRecord[0][$sortingRow] = $sourceRecord[$sortingRow];
+        $localizedRecord[0]['colPos'] = $sourceRecord['colPos'];
+
+        $parentLocalizedRecord = BackendUtility::getRecordLocalization('tt_content', $sourceRecord['tx_flux_parent'], $languageUid);
+        $localizedRecord[0]['tx_flux_parent'] = $parentLocalizedRecord[0]['uid'];
+        $localizedRecord[0]['tx_flux_column'] = $sourceRecord['tx_flux_column'];
+
+        $this->recordService->update('tt_content', $localizedRecord[0]);
     }
 
     /**
--- a/Classes/Service/ContentService.php
+++ b/Classes/Service/ContentService.php
@@ -156,12 +156,11 @@ class ContentService implements SingletonInterface
         // This condition matches colPos values which are above the constant which is a virtual, assumed limit. When
         // the $relativeTo value is above this constant value it is likely a virtual value - but it may also be the
         // case for very large page trees where UID values naturally exceed this.
-        if (MiscellaneousUtility::UNIQUE_INTEGER_OVERHEAD < $relativeTo) {
+        if (MiscellaneousUtility::UNIQUE_INTEGER_OVERHEAD < $row['colPos']) {
             // Fake relative to value - we can get the target from a session variable
-            list ($parent, $column) = $this->getTargetAreaStoredInSession($relativeTo);
+            list ($parent, $column) = $this->getTargetAreaStoredInSession($row['colPos']);
             $row['tx_flux_parent'] = $parent;
             $row['tx_flux_column'] = $column;
-            $row['sorting'] = 0;
         } elseif (0 <= (integer) $relativeTo && false === empty($parameters[1])) {
             // Special case for clipboard commands only. This special case also requires a new
             // sorting value to re-sort after a possibly invalid sorting value is received.
--- a/Classes/Service/ContentService.php
+++ b/Classes/Service/ContentService.php
@@ -370,10 +370,26 @@ class ContentService implements SingletonInterface
         } else {
             $defaultRecordUid = $uid;
         }
-        $localizedRecord = BackendUtility::getRecordLocalization('tt_content', $defaultRecordUid, $languageUid);
-        $sortingRow = $GLOBALS['TCA']['tt_content']['ctrl']['sortby'];
 
-        $localizedRecord[0][$sortingRow] = $sourceRecord[$sortingRow];
+        $sortingRow = $GLOBALS['TCA']['tt_content']['ctrl']['sortby'];
+        $localizedRecord = BackendUtility::getRecordLocalization('tt_content', $defaultRecordUid, $languageUid);
+        $previousLocalizedRecordUid = $this->getPreviousLocalizedRecordUid($uid, $languageUid, $reference);
+        if (null === $previousLocalizedRecordUid) {
+            // moving to first position in tx_flux_column
+            $localizedRecord[0][$sortingRow] = $reference->getSortNumber(
+                'tt_content',
+                0,
+                $sourceRecord['pid']
+            );
+        } else {
+            $localizedRecord[0][$sortingRow] = $reference->resorting(
+                'tt_content',
+                $sourceRecord['pid'],
+                $sortingRow,
+                $previousLocalizedRecordUid
+            );
+        }
+
         $localizedRecord[0]['colPos'] = $sourceRecord['colPos'];
 
         $parentLocalizedRecord = BackendUtility::getRecordLocalization('tt_content', $sourceRecord['tx_flux_parent'], $languageUid);
@@ -381,6 +397,9 @@ class ContentService implements SingletonInterface
         $localizedRecord[0]['tx_flux_column'] = $sourceRecord['tx_flux_column'];
 
         $this->recordService->update('tt_content', $localizedRecord[0]);
+
+
+
     }
 
     /**
@@ -414,12 +433,25 @@ class ContentService implements SingletonInterface
             // Respect the colPos for content elements
             if ($table === 'tt_content') {
                 $where .= sprintf(
-                    ' AND colPos=%d AND tx_flux_column=\'%s\' AND tx_flux_parent=%d',
-                    (integer) $row['colPos'],
-                    $row['tx_flux_column'],
+                    ' AND colPos=%d',
+                    (integer) $row['colPos']
+                );
+            }
+            // Respect the tx_flux_column for content elements
+            if ($table === 'tt_content' && $row['tx_flux_column']) {
+                $where .= sprintf(
+                    ' AND tx_flux_column=\'%s\'',
+                    $row['tx_flux_column']
+                );
+            }
+            // Respect the tx_flux_parent for content elements
+            if ($table === 'tt_content' && (integer) $row['tx_flux_parent']) {
+                $where .= sprintf(
+                    ' AND tx_flux_parent=%d',
                     (integer) $row['tx_flux_parent']
                 );
             }
+
             $where .= $reference->deleteClause($table);
             $res = $this->recordService->get($table, $select, $where, '', $sortRow . ' DESC', 1);
             $previousRow = $res[0] ?? false;
--- a/Classes/View/PreviewView.php
+++ b/Classes/View/PreviewView.php
@@ -355,7 +355,8 @@ class PreviewView extends TemplateView
                 $langPointer = $row['sys_language_uid'];
                 $localizeButton = $dblist->newLanguageButton(
                     $dblist->getNonTranslatedTTcontentUids($childrenUids, $dblist->id, $langPointer),
-                    $langPointer
+                    $langPointer,
+                    $colPosFluxContent
                 );
                 $content .= $localizeButton;
             }


glucka:
I make a PR for Drag & Drop by translated Elements to. Inspired by https://github.com/svewap/flux_translationfix

Classes/Backend/TceMain.php
```diff
@@ -535,9 +536,10 @@ public function moveRecord_afterAnotherElementPostProcess($table, $uid, $destPid
         // Following block takes care of updating the immediate record, be that a placeholder, an
         // original or a versioned copy.
         $moveData = $this->getMoveData();
-        if (!$moveData) {
```
glucka: this is always false if you drag and drop in flux column

glucka:
@NamelessCoder please review

glucka:
I try to reproduce this tomorrow.

Classes/Backend/TceMain.php
```diff
@@ -535,9 +536,10 @@ public function moveRecord_afterAnotherElementPostProcess($table, $uid, $destPid
...
-            return;
-        }
+
+        $resolveUid = $this->getOriginalRecordUid($table, $uid);
+        $updateFields['colPos'] = GeneralUtility::_GET('data')[$table][$resolveUid]['colPos'];
```
NamelessCoder: Please do not access GET parameters unless there is absolutely no other possible way to do it. In this case, DataHandler ($reference) should contain the mapping array you access here.

Classes/Backend/TceMain.php
```diff
@@ -492,6 +492,7 @@ public function moveRecord_firstElementPostProcess($table, $uid, $destPid, $move
 
         // Move the immediate record, which may itself be a placeholder or an original record.
         $row['uid'] = $uid;
+        $row['colPos'] = $newColumnNumber;
```
NamelessCoder: If you look at the code, it is hopefully clear that we should not be assigning any `colPos` value here in this function. All of that should take place in `ContentService->moveRecord`. The only property we set here is the `uid` and we only do so because this property doesn't exist in the input field array but Flux methods expect it to be there when, for example, you call `RecordService->update`.

Classes/Service/ContentService.php
```diff
@@ -156,12 +156,11 @@ public function moveRecord(array &$row, &$relativeTo, $parameters, DataHandler $
         // This condition matches colPos values which are above the constant which is a virtual, assumed limit. When
         // the $relativeTo value is above this constant value it is likely a virtual value - but it may also be the
         // case for very large page trees where UID values naturally exceed this.
-        if (MiscellaneousUtility::UNIQUE_INTEGER_OVERHEAD < $relativeTo) {
+        if (MiscellaneousUtility::UNIQUE_INTEGER_OVERHEAD < $row['colPos']) {
```
NamelessCoder: Note from TceMain applies here as well - these functions shouldn't be using `$row['colPos']` - that's what `$relativeTo` is there for.

Classes/Service/ContentService.php
```diff
@@ -423,12 +433,25 @@ protected function getPreviousLocalizedRecordUid($uid, $language, DataHandler $r
             // Respect the colPos for content elements
             if ($table === 'tt_content') {
                 $where .= sprintf(
-                    ' AND colPos=%d AND tx_flux_column=\'%s\' AND tx_flux_parent=%d',
-                    (integer) $row['colPos'],
-                    $row['tx_flux_column'],
+                    ' AND colPos=%d',
```
NamelessCoder: This and the next changes to build $where looks unnecessary: the previous code results in the same output as these conditions as the values that get used match the default values of fields.

Classes/View/PreviewView.php
```diff
@@ -355,7 +355,8 @@ protected function drawGridColumn(array $row, Column $column)
                 $langPointer = $row['sys_language_uid'];
                 $localizeButton = $dblist->newLanguageButton(
                     $dblist->getNonTranslatedTTcontentUids($childrenUids, $dblist->id, $langPointer),
-                    $langPointer
+                    $langPointer,
+                    $colPosFluxContent
```
NamelessCoder: I think this is where you want to use FluxService to read the virtual colPos value that's stored in the session. Otherwise, you'll be passing `18181` as colPos for every element on the page - which you can probably cause an error with by creating several containers with several columns and put children in each, then try to translate one of them. I expect this would affect all of them then.

Instead, if we can pass this virtual colPos and this then gets extracted in TceMain / ContentService when selecting the elements to translate, that should give the correct result.

Classes/Service/ContentService.php
```diff
@@ -389,7 +389,17 @@ public function fixPositionInLocalization($uid, $languageUid, &$sourceRecord, Da
...
         }
-        $this->updateRecordInDataMap($localizedRecord[0], null, $reference);
+
+        $localizedRecord[0]['colPos'] = $sourceRecord['colPos'];
+
+        $parentLocalizedRecord = BackendUtility::getRecordLocalization('tt_content', $sourceRecord['tx_flux_parent'], $languageUid);
+        $localizedRecord[0]['tx_flux_parent'] = $parentLocalizedRecord[0]['uid'];
+        $localizedRecord[0]['tx_flux_column'] = $sourceRecord['tx_flux_column'];
+
+        $this->recordService->update('tt_content', $localizedRecord[0]);
```
NamelessCoder: This causes an additional SQL update query - please use the `updateRecordInDataMap` method to change the values in DataHandler which then saves them for us later.

Classes/Backend/TceMain.php
```diff
@@ -268,7 +268,7 @@ public function processCmdmap_postProcess(&$command, $table, $id, &$relativeTo,
         $record = $this->resolveRecordForOperation($table, $id);
 
         if ($table === 'tt_content') {
-            if ('localize' === $command) {
+            if ('localize' === $command || 'move' === $command) {
```
NamelessCoder: I'm not sure this is good - the method that gets called here is quite heavy on SQL requests and we don't want that executing every time you drag and drop anything. This should absolutely have a guard clause to make sure it doesn't get called if there is no reason to call it. That could be done with early returns from `fixPositionInLocalization` so you don't have to, for example, fetch a record here.

NamelessCoder:
Most of my concerns can be solved with refactoring - but I am quite worried this will impact performance a lot due to performing too many SQL requests when you trigger a "move" command; requests that were previously only triggered when you performed localisation (which afaik would include when you press "synchronise").

Classes/Service/ContentService.php
glucka: The problem is when $row['tx_flux_column'] or $row['tx_flux_parent'] is NULL

Classes/Service/ContentService.php
glucka: If you move an element to the first position, `$relativeTo` is always 1.
If you move an element after an other element, `$relativeTo` is a record uid of this other element.
I think we cannot use `$relativeTo` to extract the virtual colpos value.

glucka:
@NamelessCoder Thanks for your hints! Should I try to correct it?

Classes/Service/ContentService.php
NamelessCoder: `$relativeTo` should not always be `1` - it can be either a negative or positive integer, if it's negative it's the UID of the element to move the record after, if it's positive it's supposed to be the colPos. Check the PHPDOC for the method ;)

glucka: Ohh :) thanx ;)

NamelessCoder:
@glucka Please fix those issues where you're sure what needs to be done, if some are left after that I can take care or add a new review if you prefer to work on the PR yourself.

Just for the record I am almost at the point where this implementation of virtual colPos values should be replaced with a solution where you simply must specify a unique `colPos` value for every Flux column - just like you have to for page templates. If this is done, then this hardcoded `colPos` value can be saved to DB (and we can in addition save the Flux parent/column fields for legacy compat). It puts the responsibility of finding unique `colPos` values onto the developer - and things like UNIX timestamps would be ideal for that.

When I say I'm almost at the point, that means I'm about 95% decided that this is exactly what's going to happen. In theory, it would mean that nearly all of this code gets removed because the hooks simply are no longer necessary when we want to render a proper list of nested content elements from PreviewView, when we want to translate, when we want to have drop-targets and everything else that currently goes into resolving `tx_flux_parent` and `tx_flux_column` values.

You'd be welcome to give that task a try - it's not exactly easy, but I think with the time you've spent digging into the source code you would be able to do it. Hopefully you understand my points about that vision? If not then by all means please do ask for explanations where necessary!

NamelessCoder:
Btw @glucka thank you so much for all the time you spend on this! I wish I could be more available in real-time but I'm currently tied up on a clients' project and will be for the foreseeable future.

glucka:
That's why I asked. I am absolutely in your opinion. Unique colPos values would be the cleanest method. I close this PR now and go in that direction. I look ext-gridlements like they did. Thanks to flux can i quickly realized the websites ;) So I owe you more :) I would wish that I can speak English better, but English is my fourth language :) This will get better with time too.

NamelessCoder:
One last thing: it would also be possible for us to provide a migration script that should handle most common use cases, which reads records that have Flux parent relations and analyses the Columns to find a colPos value that can be saved to DB to complete the migration. It would be important for me to minimise the impact this has on editors since we'll already have to tell developers they have to change their template files.

Maybe we could even make a migration that would read template files and do a `preg_replace` to add `colPos` arguments for any Flux columns, creating the values as timestamps and pausing a second between each conversion to make sure they become truly unique.

Long story short, let's try to provide as many automatic tools as we can, to go along with our breaking changes!

glucka:
Many thank's for the Tipps!!! I understand. I only look at gridelements the methods regarding colPos. I will follow your signpost. Migration script must also be made. I try step by step because I'm more of a frontend developer. For discussion, should be good to set up an "ext-flux" channel at typo3 slack workspace because an "unofficial" flux channel exists already.

NamelessCoder:
I think you are onto something here. The trick is exactly that we need to include the record UID in a way that makes the resulting value unique, yet still has enough flexibility to make the combination of columns you need. Just to put it into words:

* You technically take the record UID and multiply by 100.
* Then you add the chosen colPos value (which implies we can have maximum 100 columns in any template, I think this is reasonable - but we could make it 1000)
* Which means the parent element's UID can also be resolved base on just the colPos
* And should make the colPos values unique enough that we don't risk collisions with (normal usages of) backend layouts' colPos values

Off the top of my head I can't find any solid objection to it, other than perhaps we'd need to redeclare `tt_content.colPos` to make it a `bigint` so it for sure can handle a UID multiplied by our offset. And it should be very easy to integrate with the rest of TYPO3 in exactly the way I envision.

Feel free to proceed with this one!

NamelessCoder:
Side note: we'll need a smarter way to do the math about colPos values so we don't ask devs to put this in templates with `{record.uid}` and adding numbers - I think `getGrid` on ContentProvider then loop all rows + columns and recalculate the colPos value if it is lower than the offset we choose, then ask devs to use standard numbers like `1` or `5` for colPos in templates.

glucka:
Thanks for the tip! That's exactly what I was looking for ;) Ok, I'll continue..

glucka:
What can we do with `<v:content.render column="{record.uid}01"/>` ? If we don't ask devs to put `{record.uid}` in templates.

NamelessCoder:
That may be a reason to keep `flux:content.*` to make it clean, so the VH can do the math for you and you only pass `colPos` - though it isn't exactly pretty, it would be possible to do something like `{record.uid + 1005}` to render `colPos=5`. Note that this math expression will then require the very latest version of Fluid to work, or it errors out when `{record.uid}` isn't set.

We maybe could also do something like `<v:content.render colPos="{form.grid.rows.123.columns.mycolumnname.colPos}" />` which would then read from Grid which already did the math on the colPos value. Don't take me up on the variable path, though. I don't remember if you need to add `.children` to target each row/column correctly, e.g. `{form.grid.children.123.children.mycolumnname.colPos}`...

For the time being I think you should keep `flux:content.*` and make it read the record and do the math. In time we can then make it a subclass of `v:content.*` and maybe just have to override one method to make it change the colPos you give it.

glucka:
I need a little help. How can i find out the provider extension name in class.ext_update.php? 
The rest is done i make a PR today to discuss ;)

glucka:
Thanks for the hints but i realized already the update script.
getRegisteredProviderExtensionKeys do not work with 'Content' maybe because i haven't fluidcontent installed? 
`
public function main()
    {

        $fluxCore = GeneralUtility::makeInstance(Core::class);
        $providerExtensionKey = $fluxCore::getRegisteredProviderExtensionKeys('Page')[0];

        $content = '';
        $extensionKey = explode('.', $providerExtensionKey)[1];
        $legacyKey = ExtensionNamingUtility::getExtensionKey($extensionKey);
        $templateRootPath = ExtensionManagementUtility::extPath($legacyKey, 'Resources/Private/Templates/Content/');

        foreach (GeneralUtility::getFilesInDir($templateRootPath, 'html') as $filename) {

            $filecontent = file_get_contents($templateRootPath . $filename);

            preg_match_all('/(flux:grid\.column.*)(name="(.*?)")/', $filecontent, $columnMatches, PREG_SET_ORDER);
            if (count($columnMatches) > 0) {
                foreach ($columnMatches as $i => $match) {
                    $areas[$match[3]] = $i;
                    $newColumnTagPart = preg_replace('/name=".*?"/', 'colPos="' . $i . '"', $match[0]);
                    $filecontent = preg_replace('/' . $match[0] . '/', $newColumnTagPart, $filecontent);
                }
            }

            preg_match_all('/(flux:content\.render.*)(area="(.*?)")/', $filecontent, $renderMatches, PREG_SET_ORDER);
            if (count($renderMatches) > 0) {
                foreach ($renderMatches as $match) {
                    $newRenderTagPart = preg_replace('/area=".*?"/', 'column="' . $areas[$match[3]] . '"', $match[0]);
                    $filecontent = preg_replace('/' . $match[0] . '/', $newRenderTagPart, $filecontent);
                }
            }

            if (count($columnMatches) > 0 || count($renderMatches) > 0) {

                $content = 'Change name attributes to colPos, area to column in grid templates:' . PHP_EOL;
                $content .= $filename . ' affected' . PHP_EOL;
                file_put_contents($templateRootPath . $filename, $filecontent);

                $content .= 'Set grid column colPos values in DB:' . PHP_EOL;
                foreach ($areas as $i => $v) {

                    if (ExtensionManagementUtility::isLoaded('fluidcontent')) {

                        $fcefile = $providerExtensionKey . ':' . $filename;
                        $GLOBALS['TYPO3_DB']->sql_query("UPDATE tt_content AS t1
INNER JOIN tt_content AS t2 ON t1.tx_flux_parent=t2.uid
SET t1.colPos = t1.tx_flux_parent * 100 + $v 
WHERE t1.tx_flux_column = '$i' AND t2.tx_fed_fcefile = '$fcefile'");

                    } else {

                        preg_match_all('/flux:form.*?id="(.*?)"/', $filecontent, $formIdMatches, PREG_SET_ORDER);
                        $formId = $formIdMatches[0][1];
                        $ctype = $extensionKey . '_' . $formId;
                        $GLOBALS['TYPO3_DB']->sql_query("UPDATE tt_content AS t1
INNER JOIN tt_content AS t2 ON t1.tx_flux_parent=t2.uid
SET t1.colPos = t1.tx_flux_parent * 100 + $v 
WHERE t1.tx_flux_column = '$i' AND t2.CType = '$ctype')");

                    }

                    $content .= $GLOBALS['TYPO3_DB']->sql_affected_rows() . ' rows affected' . PHP_EOL;

                }
            }
        }
`

The language handling is little bit tricky, i still need to use TceMain.php hooks. 
I hope next week can i present something.

NamelessCoder:
I really hate to say this, but that migration script is not safe to use. For example, you must not rewrite template files this way since there are many, many different ways you could write a template where this would completely destroy it. It really is the only sane choice to detect and warn about templates, but not try to rewrite them like this.

Also, you can't resolve providers based on which provider extensions provide Page integration and then use that to resolve Content - it's completely possible to have one without the other so we must target Content here. And it is not necessary to migrate Page implement templates as they already use the right type of colPos definitions.

I'll try to explain and illustrate the issue with the migration script.

The main problem here is that the grid can be constructed with dynamic variables that depend on the record. This is why I proposed to resolve the provider using the record, because this "thaws" the variables in the record and assigns the various variables that may be used to render the grid.

Consider the following form example:

```xml
<flux:form id="myform">
  <flux:grid>
    <flux:grid.row>
      <f:for each="{settings.sectionObjectsFromFlexForm}" as="object">
        <flux:form.column name="column{object.id}" />
      </f:for>
    </flux:grid.row>
  </flux:grid>
</flux:form>
```

Now, if you run the migration script on such a template, it will fail in the following ways:

* The column that gets checked for will be `column{object.id}`
* Only a single column will be detected when the record may define multiple ones
* If you rewrite a colPos value here, that colPos value will be the same for all child columns
* Page module will show duplications of the first column in all columns

This is quite a problem as you can see. The same type of problem happens if a dynamic variable is used as column name (for example if a condition generates the name). And you can probably imagine the problems that would arise if we tried to rewrite templates that use this kind of dynamic code.

The right way to read these columns is to extract the Grid object via the provider. That way you analyse the generated structure, not the Fluid source code that resulted in it. Because there are so extremely many variations on how a grid can be built, including using a different namespace or a VH override to make the components, we won't be able to correctly analyse these usages unless we execute the template's Configuration section and extract the Form/Grid.

My suggestion: try to get it working by resolving Content providers and reading Grid instances. If you are not able, you are of course welcome to leave that task for me. But I'm afraid we can't allow the migration script to actually rewrite templates - it'll have to just warn and give hints about fixing. That said, it's VERY okay to have the migration script perform SQL queries as needed - _after_ the developer has fixed the template so the columns all have a colPos and tested that it works ;)

NamelessCoder:
Not to mention, it's equally possible to not use a template file at all to define the Grid or even the Form, a Provider can return non-standard template paths and filenames, you can define the structures in TS as well, and you can also use inline syntax in the templates. So there are too many types of sources to correctly handle all of them; the only consistent way we have is to use the API which gives all types of Grid sources the same interface - but this will only support reporting the issues, we can't write changes back via the API.

glucka:
Unfortunately you are absolutely right :) I give a try to get it working with Content providers. 
But first of all i make some test with tt_content translate, because i don't know in this moment how many hooks we need to realize this. And many thanks for your commitment Claus, i learn a lot from it!

glucka:
I think i found a typo3 core bug witch related to this PR. 
Reported here: https://forge.typo3.org/issues/85013 Let's see what happens. 
Another thing: I think we should always take the record uid in default language as colPos.
If we use not the same colPos for each localization, native typo3 translation does not work.

NamelessCoder:
> Another thing: I think we should always take the record uid in default language as colPos.

Off the top of my head: true!

The colPos should remain the same so the synchronisation of translated columns will work, you're completely right about that.

Let's see what response you get on the issue you reported. To be honest I'm not sure myself what is expected and which flags might affect if the original/translation moves when you D&D the other...

PR #1546, opened 2018-04-18
[BUGFIX] Preview of hidden pages not possible

andinger:
Previewing disabled pages causes and endless loop because the QueryBuilder excludes disabled pages by default.


NamelessCoder:
This will not be merged, sorry. The RecordService must not remove all restrictions for every single query in order to remove restrictions from one specific use case. Instead, you must find the place where RecordService gets used - and fix it there.

It is also likely that the problem you are looking for does not exist in Flux but rather in Fluidpages.

And lastly, please make sure that you've confirmed the problem exists on `dev-development` not just of Flux but of Fluidpages (and others if you use them).

NamelessCoder:

PR #1552, opened 2018-05-17
Using colPos instead of tx_flux_column in flux:column

glucka:
@NamelessCoder I only did what I really (90%) understood ;)
Things that still need to be developed:
- The Column component in Flux should require a colPos - possibly by throwing an exception in build() if the value is not set.
- Save legacy tx_flux_column values
- DB migration tx_flux_column to colPos
- Fluidpages plugin must be updated too because Content Service no more available
- Dynamic generate TCA Column names?

Otherwise everything goes fine, drag & drop, copy/paste works with translated elements too 
(in connected and in copy mode)


glucka:
I check my typo3 core fix tomorrow again... This is maybe not the best solution.

glucka:
Now discovered: Copied grid elements lose their children. I'm looking for the reason.

glucka:
Copied grid elements lose their children because colPos need to be recalculated.

NamelessCoder:
@glucka Big thanks for all your efforts here. I will have some time during the weekend to check this out, including the core patch you made. If you don't mind I'll simply push to your branch as smaller commits (that will be squashed in the end) - that way you can follow my thoughts and if you see anything you know would be a problem, you can comment directly on the code lines ;)

My priorities would be:

* Making sure the basics are working
* Nice-to-have if translations are working
* At least a superficial test of migrations
* Seeing if there is more code and overrides we can delete from Flux

glucka:
Agreed ;) I look forward to it!

NamelessCoder:
@glucka I've found a few interesting things that have become possible with colPos usage - I will try to work on that a bit, then I will create a new branch on the Flux and Fluidpages repositories where we can continue.

Just so you are aware, I am trying now to:

* Use a simple override of the core's PageLayoutView and remove all the "draw" methods from Flux's PreviewView.
* In the PageLayoutView, use a customised BackendLayoutView which works with a complete Flux form as columns source. It looks like a custom BackendLayoutView is the key: it is how Fluidpages converts a Flux grid into a "backend_layout" array so we already have code that does this - perhaps we can in addition simplify fluidpages and put this code in flux.

If you feel like continuing I would very much appreciate if you could look into the migration of old record relationships to new colPos values!

glucka:
Sounds very good! I'm curious ;)
I'm on holiday but about in 2-3 weeks i'm sure i'll find some time to look the migration things.

NamelessCoder:
Dangnamit, I think I've got it now! It takes a bit of cheating and API abuse but it is somewhere around 5% the overrides we had before. Will create new feature branches in a moment, but from what it looks like now... it is actually working with a nested version of a nearly vanilla PageLayoutView that is provided with backend layout columns from an overridden BackendLayout.

The building of BackendLayout and corresponding arrays has been moved to the Grid object so you can call `$grid->buildBackendLayout()` and other methods to build exactly the type of resource you need. This should then be very compatible with anything the core chooses to do at a later time concerning making backend layouts more flexible with the goal of making structured content elements.

But until then... it looks like we will have a seriously good solution here.

NamelessCoder:
Replaced by https://github.com/FluidTYPO3/flux/pull/1563 - I've described what is going on in a bit more detail there, and the diffs should also be fairly clear. As mentioned there are several TODOs spread through the code - I will keep working on those when I've got time to spare.

PR #1558, opened 2018-06-12
[BUGFIX] "Show hidden content elements" in Page module doesn't show h…

flixflax:
…idden elements which are inside flux container content element #1484


flixflax:
What about this pull request? I tried to fulfill all the changes you requested in https://github.com/FluidTYPO3/flux/pull/1496 since the author of it did not. The Issue has been closed but it still persists in development branch.

rupasix:
That's true - without that PR flux-dev is almost unusable from editor standpoint. All hidden CE inside flux columns are hidden in BE Page View.

NamelessCoder:
The PR is exactly like it should be to pick up where #1496 left off - but the unit tests need to be fixed. You may need to refactor so the QueryBuilder object gets returned from a (protected) method that you can mock.

You can see which ones are failing on https://travis-ci.org/FluidTYPO3/flux/jobs/391145843 (same tests fail on all versions)

NamelessCoder:

PR #1560, opened 2018-06-15
[BUGFIX] Reset parent container after rendering grids

### LINKED ISSUE #1553: [BUG] Typo3 8.7 / Error: FluidTYPO3\Flux\Form\Field\File::getColumns() / flux dev (2018.04.24)
Hello

if have a Error since i upgarde PHP to 7.2 and need to use the newest Version of Flux.

Using Typo 3 8.7.13
PHP 7.2
and the Development-Version of flux and fluidpages.

```
Call to undefined method FluidTYPO3\Flux\Form\Field\File::getColumns() 

Error thrown in file
/www/jd.dev/beverland/doc/typo3conf/ext/flux/Classes/View/PreviewView.php in line 308.
```
in older Version of Flux it does work.

I cant understand why it is \File::getColumns()
the function getColumns() is written in class Row
```
class Row extends AbstractFormContainer implements ContainerInterface
{

    /**
     * @return array
     */
    public function build()
    {
        $structure = [
            'name' => $this->getName(),
            'label' => $this->getLabel(),
            'columns' => $this->buildChildren($this->children)
        ];
        return $structure;
    }

    /**
     * @return Column[]
     */
    public function getColumns()
    {
        return (array) iterator_to_array($this->children);
    }
}
```

can anybody help ?

thx

HenningCash:
This should fix #1553. I have no idea why this has worked in previous releases but this should be the desired behavior


NamelessCoder:
Looks reasonable. Problem in template reported in #1553 is that the grid comes before the form fields and is nested inside the form tag. Avoiding either of those will avoid the issue (until a fix can be released).

Issue #1561, opened 2018-06-18
Error when getting application context


kraftb:
With 9.2 the application context is retrieved in a different way. I have created a patch which fixes this issue for 9.2 and will submit it via a pull request.

kraftb:
The patch does not tackle the proplem of inter-version compatibility. By applying this patch the code will only work in 9.2. I guess the solution will be to create an if-condition accessing the TYPO3 version number and eventually putting the "getContext" lines in a wrapper method.

NamelessCoder:
Solved as part of https://github.com/FluidTYPO3/flux/pull/1563, will be included in release 9.0.

PR #1562, opened 2018-06-18
[BUGFIX] Error when getting application context

kraftb:
The API for retrieving the current application context has changed.

Branches: development
Resolves: #1561


NamelessCoder:
Introduced a small utility for that as part of https://github.com/FluidTYPO3/flux/pull/1563

PR #1563, opened 2018-06-21
[FEATURE] Use colPos instead of tx_flux_column

NamelessCoder:
This is a breaking change!

With this patch, Flux switches from using tx_flux_column
and tx_flux_parent to determine relations between parent
and child records, to using a pure colPos based solution.

The new behavior is designed in the following way:

* Every flux:grid.columm *must* now use “colPos” argument
* The “colPos” should be treated exactly like in page templates
  so you use numbers from 0-99 (99 is the new maximum!)
* The final colPos value is then calculated based on a math
  operation that includes the parent’s UID.

This means that a flux:grid.column with a colPos of 4,
when you use it in a grid inside a content record with
an UID of 123, the real colPos value becomes 12304.

This value can then be reverse calculated to return
the UID of the parent and the colPos in the template.

In order to render columns, flux:content.render and
the .get variant have a new “colPos” argument which
can be used with the virtual colPos value, for example
“4”, to render the real column based on the UID of the
current parent record.

It is also possible to use math to calculate a colPos
value and use this with v:content.get and v:content.render
if this is preferred. It will then of course be necessary
to manually keep track of when to use a “raw” colPos
value (in page templates) or a calculated colPos value
(in content templates).


NamelessCoder:
Must be combined with https://github.com/FluidTYPO3/fluidpages/pull/384 if you use Fluidpages.

glucka:
I try this new PageLayoutView but i see "No edit access" at every grid columns.
Backend Page Layout already set to "FluidTYPO3.Flux"

NamelessCoder:
Hmm, backend layout should be set to "fluidpages" - I tested with this. But I think I know what goes on with the "no edit access" and it is of course fixable! Thanks for the feedback. Do try with fluidpages and let me know if that also does it.

Short Q: are you admin user when trying?

glucka:
Yes, i'm admin and i see only "None" and "FluidTYPO3.Flux" as Backend Page Layout

NamelessCoder:
Got it! Pushing an update for Flux in a moment. The fluidpages BE layout will be named differently now - the ID will always be `fluidpages__grid` (so we must make a migration for this as well), but the label will now be the extension name that registers the fluid template.

This may change again but for the moment the identifier of the BE layout has to be changed.

BenjaminBeck:
I do have a similar problem with TYPO3 8.7.13: Fluidpages Backend-Layouts are working (needed a small change for TYPO3 8 compatibility), but the content-areas inside a content element only display "No Access".  I managed to trace this to its source: (https://user-images.githubusercontent.com/1185872/41783162-83a6ee90-763c-11e8-8f81-9313becf018e.png).

NamelessCoder:
@BenjaminBeck Are you using the most recent state? I force-push to this PR branch as I make changes. I did touch on that specific point and in my testing there's a valid entry in `$head` corresponding to the `columnKey` (which btw has the right value in your dump).

BenjaminBeck:
@NamelessCoder I am on flux:#b4e341c and fluidpages:4e8e795 - it looks to me that these are the most recent. I will look if i find a more root-like cause by digging deeper ..

NamelessCoder:
Excellent - and thanks for helping out!

BenjaminBeck:
Got it working with this change:
https://github.com/BenjaminBeck/flux/commit/6ea4705e68f615e73267898e0b792b8ab6fffc90
Maybe i need this because of a change in TYPO3 from 8.x to 9.x. Without it, ``$dbList->tt_contentConfig['cols']`` is never changed from it´s initialisation value (``1,0,2,3``)

NamelessCoder:
Thanks @BenjaminBeck - I've integrated that line and will push the result in a moment. Further removals have been made possible.

NamelessCoder:
2,000 lines of code removed and still going. A serious amount of DB queries are gone especially when doing move/copy operations. Some parts are still not functional - for example, the white/blacklisting of allowed content per column needs to be homogenised.

BenjaminBeck:
I have a scetchy version running for the Ajax-LocalisationController that returns the records which will be copied or localized. It works but i just realized, that the colPos values of the localized records need to be recalculated based on the new parent id..
( https://github.com/BenjaminBeck/flux/commit/a638821c113e0144cce6f9ae2befa769bf4ae687 )

NamelessCoder:
@BenjaminBeck Great that you're spending time on the localisation - that's exactly the part I haven't touched yet. But about your patch on https://github.com/BenjaminBeck/flux/commit/a638821c113e0144cce6f9ae2befa769bf4ae687 I would absolutely like to avoid the XCLASS and that has to be possible somehow (using post-processing of records which were localised, to read the new parent and generate a new colPos value).

Do note about the current state of TceMain that the code that fixes IRRE is disabled with an explicit `false` in the condition, and there's a link to a core patch which should also make it possible for us to solve this without overrides. If you did not see it already, I added a ColumnNumberUtility that can be used to first extract the "local" colPos value from the virtual one, then calculate a new virtual one based on the new parent (as looked up based on the old parent) plus the local column number.

Am I missing something or will the AJAX call already select records based on colPos (at least on 8.7) and does this not fire DataHandler methods? I haven't tested those features yet so I hope maybe you can come up with something clever to avoid the XCLASS ;)

NamelessCoder:
Also about the TceMain class - I think @glucka and myself were a bit too aggressive removing some of the hook methods. We probably still need at least the command post method to catch `localize` commands, and there's code in AbstractProvider that has to be moved to this class.

NamelessCoder:
@BenjaminBeck I've updated the patch to restore the TceMain command post-processing, and added TODOs where I think we may be able to solve the localisation colPos recalculation. If I can help you continue please let me know!

glucka:
I have removed everything directly so we can find a clean way to rebuild it ;) It looks really great!
Regarding translation, i think we have 2 options:
- For colPos generating we use the default language parent uid in the translation
(more vanilla typo3 metodhs)
- or we use the translated parent uid
(i think this is a bad way)

BenjaminBeck:
@NamelessCoder thanks for the TODOs, I will have a look at it!
@glucka there is also the possibility to have free translations that are not connected to the default language, because of this i thing using the uid of the default language is not possible.

glucka:
@BenjaminBeck yes! I think in "connected mode" (if l18_parent exists) we should use the uid of the default language and in "free mode" the translated. In "free mode" we dont need any typo3 methods to translate.

NamelessCoder:
@glucka / @BenjaminBeck I'm not sure about this, but I think this might all be handled by the RecordList component in which case we actually don't need to do anything in terms of *selecting* the right records in the page layout view? So this is all about how we should post-process records when they get localised? Correct me if I'm wrong!

But when records are localised I think we need to look at the *parent* record first, and see if that is a translated version of a parent - if it has a translation parent, we should use the parent UID of it (since it will then have the right colPos value when you render in FE and individual translated records can be overlaid). But if the *parent* was translated in free mode we should use the UID of the translated version of the parent - because then the parent will be only output in that language and will only have that UID as a relationship.

This logic should also mean that the page layout view functions correctly:

* When a translated version of page content is in "free" mode it has its own colPos value or children
* When it is a connected translation, it has the same colPos as the original - and child elements may be translated in either connected or free mode as needed without changing colPos.

NamelessCoder:
Btw I still have my fingers crossed that you two are able to come up with a nice solution for the localisation. Then I will happily take care of everything else. So you are aware, my next steps will be:

* Fixing a few misbehaviors, for example when you hold CTRL to create a copy and drop the element in a nested column, the copy gets the original's colPos for some reason.
* Implementing a cascade copy operation that copies children when you copy a parent (this is the expected behavior - right? Should we have a control for this somehow so you can copy just the container and none of the children? Or should we even solve this with a "copy children" button that can be shown in the Preview area of the copy if we detect the original has children but copy doesn't - so you selectively copy children per column.
* Finishing the "unused content" integration so it actually reports content you don't use in any root or Flux column across the entire page.
* Recreating the "column" values itemsProcFunc that lists columns of the parent element as selectable options for colPos when editing a child.
* Continuing with the migration assistant, at first just to report if/which content there is to migrate, then to do as much automatic migration of relations as possible. I'm hoping that can solve 95% of use cases and the rest can be done with manual SQL using the output of the migration tool as guide.

Let me know if you've found something that I haven't listed here which isn't related to translations!

glucka:
@NamelessCoder I think so too!
This should be the cleanest solution.

NamelessCoder:
Update re: wrong column position when copying (and also when dragging). It is reproducible by having two content elements in colPos=1 in a page template, then drag-dropping the bottom one to the top of the column. The element ends up in colPos=0 - because of the workaround code in TceMain which forces the colPos if it is not in the field array.

I'm guessing if we solve the TYPO3 core issue this also gets solved. For the time being I have adjusted the workaround code so it will only trigger if the dropped element has a translation parent.

By the way I will create individual commits from this point so you can more easily follow the changes. Just needed to get the bulk part done and tests green ;)

glucka:
@BenjaminBeck  I am now on vacation for 2 weeks. I can only look after it. :(
@NamelessCoder Cascade copy operation: I think copies children when you copy a parent is the  expected behavior without any control of them.

NamelessCoder:
Remaining tasks:

* Cascading copy when copying parent
* Migration assistant tools
* Localisation behaviors

We're almost there!

BenjaminBeck:
I created a pull request for the first set of changes.
For the translation behavior i would do it like this:
The default translation behavior is not to translate child records.
But if you use the translation wizard from the page view all contents and childs of the selected content area are translated (copied).
For this wizard to also list and use the child records I currently have the implementation using a XCLASS. The only idea I had to not need this XCLASS is to make modifications on the JavaScript side - basically a xclass for requirejs. And in this additional JavaScript we could call a flux Ajax route that provides the record list including nested content elements.
That way we could also visualize the nested structure in the translation wizard.
Would this be ok?

NamelessCoder:
@BenjaminBeck Just so I understand: the problem is that when you click to translate a column, the content records that are selected is limited to a single colPos, which means you only get the parents' UIDs? And that if you then translate this, then "localize" only gets called with those UIDs and the resulting localisations have no child content?

If this is the case, then I can see why you've gone in the direction of the XCLASS (for PHP or JS) - and I agree it would actually be a more detailed solution than what I have in mind.

Because what I had in mind, which I've come up with primarily to avoid the XCLASS, is to sacrifice the ability to *see* that child content gets translated - but cascade the translation operation to each column inside a parent, then post-processing the records to set a new colPos (if the translation was done in free mode).

So you don't see the child content in the wizard - that's what we sacrifice - but if a parent is included in the list, our TceMain hook expands that and queues translating and repositioning of all the child elements (recursively if children have children and so on). I think that's the only way we can avoid XCLASS and imho it's a reasonable sacrifice (given you can always manually adjust child localisations if you have to).

Your thoughts on that?

NamelessCoder:
This is going so very well! It looks like at this point, we just need the localisation logic to behave nicely. @BenjaminBeck I will now take a further look into the localisation logic but I'm not sure what will come out of it. I'll try to do  the same cascading copy operation, but instead of just creating a copy, creating one in the target language. That would then result in the somewhat untransparent "child content magically gets translated" behavior (I hope!).

BenjaminBeck:
@NamelessCoder 
> @BenjaminBeck Just so I understand: the problem is that when you click to translate a column, the content records that are selected is limited to a single colPos, which means you only get the parents' UIDs? And that if you then translate this, then "localize" only gets called with those UIDs and the resulting localisations have no child content?
yes, exactely

> [..] Your thoughts on that?
That is fine with me. If i find the time i can create a flux_fancyui extension (:

> This is going so very well!
Yes! - I also noticed, that after my language patches, pretty mutch worked out of the box ..

NamelessCoder:
> That is fine with me. If i find the time i can create a flux_fancyui extension (:

Oh please do! I will consider this right away and do some checks to avoid cascade-copying child content if there is already a localisation operation pending. Though there isn't when you run vanilla TYPO3, there would be if you created such an extension so I might as well design for that from the start.

NamelessCoder:
One thing @BenjaminBeck - could you take a look at the migration script I added and see if we need to make changes to migrate translated records? I think there might be a weak point when the record is a translation and is in connected mode, where the colPos would be written incorrectly as associated with the localisation, not the original...

NamelessCoder:
That's half of the localisation logic done. The connected mode translation remains to be fixed - once a parent is copied, we get the "No edit access" issues and no child content gets listed. Hopefully that's a very minor issue and only related to how we initialise the RecordList component.

So very, very close...

NamelessCoder:
Outstanding minor things:

* Refactoring outlet processing to TceMain
* Fixing the "content used?" reporting

Everything else appears to be working now, including translation support in both modes and synchronisation in connected mode.

BenjaminBeck:
> could [you] take a look at the migration script I added
I did and constructed a test. That failed with connected translations. 

For localizing i noticed three bugs:
- Reordering contents inside a content-grid by d&d in the default language does not change sorting of connected translations
- Reordering contents in the page layout (d&d to the bottom) in the default language lets translations disappear
- Localizing from sys_language_uid=1 to sys_language_uid=2 (this is only possible with the localization wizard and a new option since 7 or 8): content grid elements are not visible or not copied

When i had a closer look, I´ll give you more info on each of them.

NamelessCoder:
> deleting a connected, localized grid content element does not delete its children

We may want to implement cascading deletions as well. I think that would be logical - otherwise, when we restore the "unused content" feature, you'd just have to manually remove a lot of waste child content.

My logic: if you wanted to keep the child content, you would have moved it away before deleting the parent.

> translating (connected mode) a content element inside a content grid where the content grid already exists in the target language, does not set the right colPos value on the new created translation

I'm unable to confirm that - I've tried two ways: first, I edit the original and create a new language version. That causes the localised record to be created and it ends up in the right column and language. Then, I try the same procedure by clicking "translate" button, which then only gives me the "translate" (no "copy") option. Clicking the guide through shows no content will be translated, but finishing the guide actually translate the missing content - and puts it in the right column.

> l18n_parent points to the parent grid in the default language instead of the same element in the default language.

Hmm - you mean, child records in DB have the UID of the *grid parent*, not the *localisation parent*, in this column? If so, that is very unexpected: I don't think Flux ever writes this value... but perhaps we configure a core object with incorrect values somehow?

BenjaminBeck:
>> deleting a connected, localized grid content element does not delete its children
 
>  We may want to implement cascading deletions as well. I think that would be logical - otherwise, when we restore the "unused content" feature, you'd just have to manually remove a lot of waste child content.
> My logic: if you wanted to keep the child content, you would have moved it away before deleting the parent.

Yes, I´m the same opinion
> > translating (connected mode) a content element inside a content grid where the content grid already exists in the target language, does not set the right colPos value on the new created translation

> I'm unable to confirm that - I've tried two ways: first, I edit the original and create a new language version. That causes the localised record to be created and it ends up in the right column and language. Then, I try the same procedure by clicking "translate" button, which then only gives me the "translate" (no "copy") option. Clicking the guide through shows no content will be translated, but finishing the guide actually translate the missing content - and puts it in the right column.

I`ll check again and provide more information

>> l18n_parent points to the parent grid in the default language instead of the same element in the default language.

> Hmm - you mean, child records in DB have the UID of the grid parent, not the localisation parent, in this column? If so, that is very unexpected: I don't think Flux ever writes this value... but perhaps we configure a core object with incorrect values somehow?

Yes, i`ll check that as well

NamelessCoder:
> Hmm - you mean, child records in DB have the UID of the grid parent

Issue fixed - was a blunder in the code simply copying from parent. So it did make sense :)

> I'm unable to confirm that - I've tried two ways: first, I edit the original and create a new language version

I *think* this also got solved by proper assignment of translation source or parent in the copy loop. Knock on wood.

And cascading deletion of children is now active :)

BenjaminBeck:
Great! I worked on my branch: https://github.com/BenjaminBeck/flux/commit/3836ed58d8a3a178baa99dba823fa3a42b34dc85
I think i solved most of the problems. 
The change on processDatamap_postProcessFieldArray prevented a problem when saving a record.

NamelessCoder:
Regarding `$reference->cmdmap[$table][$id]['move'] === 'Root'` - this is how it looks on 9.3 at least. It fixes a very particular bad behavior where when you move a nested content element to the top of a nested column, a flashmessage appears saying you tried to insert it on pid=0 where you don't have access. It may not be necessary to perform this on 8.7 at all - if this D&D operation doesn't cause flashmessages and sets the right pid value in DB on that version. Hope you can test for this :)

NamelessCoder:
By the way, that issue seems to be a problem in the TYPO3 core's JS that reads the target pid value from the dropzone... I've verified the markup does contain the right value, it just doesn't seem to be correctly read when you drop elements. It isn't something we need to take care of imho, the workaround there is just fine. But it is pretty peculiar.

NamelessCoder:
> Localizing from sys_language_uid=1 to sys_language_uid=2 (this is only possible with the localization wizard and a new option since 7 or 8): elements of a content grid are not visible or not copied

I think this is the only remaining problem - and I think I know how to solve it :)

NamelessCoder:
This could be it. @BenjaminBeck did you have a chance to test the migration script (EM extension update script) on a setup that contains old Flux content which had connected and free translations?

If that works I would say we are ready for a release, after I've done some minor cleanups that are not relevant for this pull request.

This feels pretty awesome now :)

BenjaminBeck:
@NamelessCoder 
Yes i tested it with conntected translations but it failed: It look like it uses the record uid instead of the language parent uid to calculate the colPos,

There are still some todos with moving and copying of translated & connected records - but this has never been different ):
So, i think yes, this can be released after fixing the migration script.

glucka:
@NamelessCoder: I think the forge bug (85013) is back. I'm on b17b482

glucka:
I can confirm @BenjaminBeck On Typo3 v8 'Root' does not exists and the drop zone works as expected.
The issue 85013 is because the dropzone exist (of course) only in default language. No dropzone -> no colPos in localized content datamap. Currently (b17b482) issue 85013 on v8 is back. Weekend I steal time to test the migration script! :)

NamelessCoder:
Thank you both so very, very much for your help with this!

PR #1565, opened 2018-06-25
[BUGFIX] Set l18n_parent to localisation parent

BenjaminBeck:
I fixed the l18n_parent assigment and also expanded the creation of the copyRecord parameters because i think that way it is more easy to understand

### DIFF
--- a/Classes/Backend/TceMain.php
+++ b/Classes/Backend/TceMain.php
@@ -81,6 +81,91 @@ class TceMain
     }
 
     /**
+     * Command pre processing method
+     *
+     * Deletion of child records must be done before parent record is deleted,
+     * so this logic is here in the preProcess
+     *
+     * @param string $command The TCEmain operation status, fx. 'update'
+     * @param string $table The table TCEmain is currently processing
+     * @param string $id The records id (if any)
+     * @param mixed $value
+     * @param DataHandler $reference Reference to the parent object (TCEmain)
+     * @param array $pasteUpdate
+     * @return void
+     */
+    public function processCmdmap_preProcess($command, $table, $id, $value, &$reference, &$pasteUpdate){
+        $doProcessCommand = in_array($command,['delete']);
+        $doProcessCommand = $doProcessCommand && 'tt_content' == $table;
+
+
+        if ($doProcessCommand) {
+            // A copy, or localisation (which is also a copy) was made. Cascade copy operations for child records.
+            $gridChildRecords = [];
+            $resolver = $this->objectManager->get(ProviderResolver::class);
+            $originalRecord = $this->recordService->getSingle($table, '*', $id);
+            $recordCanBeGrid = !empty($originalRecord['pi_flexform']);
+            if(!$recordCanBeGrid){
+                return;
+            }
+
+
+            $primaryProvider = $resolver->resolvePrimaryConfigurationProvider(
+                $table,
+                null,
+                $originalRecord
+            );
+            $colPosCalcBaseUid = $id;
+            $gridIsTranslation = $originalRecord['sys_language_uid'] > 0;
+            if($gridIsTranslation){
+                $colPosCalcBaseUid = $originalRecord['l18n_parent'];
+            }
+            if ($primaryProvider) {
+                $childColPosValues = [];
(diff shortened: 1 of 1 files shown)


Classes/Backend/TceMain.php
```diff
@@ -80,6 +80,91 @@ public function processDatamap_postProcessFieldArray($status, $table, $id, &$fie
...
+        $doProcessCommand = in_array($command,['delete']);
+        $doProcessCommand = $doProcessCommand && 'tt_content' == $table;
+
+
+        if ($doProcessCommand) {
+            // A copy, or localisation (which is also a copy) was made. Cascade copy operations for child records.
+            $gridChildRecords = [];
+            $resolver = $this->objectManager->get(ProviderResolver::class);
+            $originalRecord = $this->recordService->getSingle($table, '*', $id);
+            $recordCanBeGrid = !empty($originalRecord['pi_flexform']);
```
NamelessCoder: This assumption is wrong - any record can be a grid, regardless whether it uses a flexform.

Classes/Backend/TceMain.php
```diff
@@ -80,6 +80,91 @@ public function processDatamap_postProcessFieldArray($status, $table, $id, &$fie
...
+                $originalRecord
+            );
+            $colPosCalcBaseUid = $id;
+            $gridIsTranslation = $originalRecord['sys_language_uid'] > 0;
+            if($gridIsTranslation){
+                $colPosCalcBaseUid = $originalRecord['l18n_parent'];
+            }
+            if ($primaryProvider) {
+                $childColPosValues = [];
+                foreach ($primaryProvider->getGrid($originalRecord)->getRows() as $row) {
```
NamelessCoder: I think this now makes sense as a method on Grid, like `buildColumnPositionValuesForParent(int $parentUid)` which returns a plain array of all possible colPos values in a grid.

Classes/Backend/TceMain.php
```diff
@@ -80,6 +80,91 @@ public function processDatamap_postProcessFieldArray($status, $table, $id, &$fie
...
+            $gridHasChildren = count($gridChildRecords) > 0;
+
+            if('delete' == $command && $gridHasChildren){
+                $cmd = [
+                    'tt_content' => []
+                ];
+                foreach ($gridChildRecords as $recordToDelete) {
+                    $cmd['tt_content'][ $recordToDelete['uid'] ] = ['delete' => 1];
+                }
+                $dataHandler = GeneralUtility::makeInstance(DataHandler::class);
```
NamelessCoder: Call on $reference instead - and I think this needs to be different. This is the command pre-processing method so we can simply extend the command map array on $reference and it should then do those, between calling this method and the post-processing method.

Classes/Backend/TceMain.php
```diff
@@ -102,11 +187,18 @@ public function processDatamap_postProcessFieldArray($status, $table, $id, &$fie
...
+        $doProcessCommand = in_array($command,['localize','copy','copyToLanguage','delete']);
+        $doProcessCommand = $doProcessCommand && 'tt_content' == $table;
+
+        if ($doProcessCommand) {
             // A copy, or localisation (which is also a copy) was made. Cascade copy operations for child records.
-            $recordsToCopy = [];
+            $gridChildRecords = [];
             $resolver = $this->objectManager->get(ProviderResolver::class);
             $originalRecord = $this->recordService->getSingle($table, '*', $id);
+            $recordCanBeGrid = !empty($originalRecord['pi_flexform']);
```
NamelessCoder: Looks like we have significant duplication here - maybe there's a way we can refactor!

NamelessCoder:
I've taken the liberty of refactoring and recreating the operations you had added here - so although this PR won't get merged, I give you full credit for pointing out the need for improvement ;)

BenjaminBeck:
Ok, i will then push the fixes and changes i am currently working on into my branch and we will then figure out how to merge

NamelessCoder:
You should pull from `feature/colpos` first, I've changed TceMain, I picked the essential stuff from this PR and refactored the original PR - those changes I hope solve the same as your changes in this branch. I'm not sure we need further changes in TceMain...

PR #1568, opened 2018-06-27
[TASK] Migration script: Use correct UID for connected translations

BenjaminBeck:


NamelessCoder:
Thanks Benjamin - you got to this one before I did, from your explanation earlier I had already picked this one out for bug extermination here this evening.

PR #1582, opened 2018-07-15
[FEATURE] Automatic grids from section objects

NamelessCoder:
Adds a flag for section objects to make each object
that gets saved in the section, a content container
column in the page layout view.

Example Form with section-objects-as-grid:

    <flux:form id="sectionObjectsAsGrid">
        <flux:form.section name="objects">
            <flux:form.object name="main" contentContainer="1">
                <flux:field.input name="test" label="Test" />
            </flux:form.object>
        </flux:form.section>
    </flux:form>


NamelessCoder:
screen shot 2018-07-15 at 16 34 26
screen shot 2018-07-15 at 16 33 56

NamelessCoder:
Hidden field for column position and input field for label are added automatically if `contentContainer="1"` is specified. The rest are the fields that you yourself specify.

pkerling:
Is it intentional to make the label mandatory? Depending on the type of CE, it might not make sense to put labels on the content areas and it would be better number them (like "Content Area #1") automatically.

NamelessCoder:
The label is optional (but the field will always be there, perhaps that's what you meant?).

Regarding automatic numbering that comes with the same challenge as the colPos value and the thing this patch attempts to solve - if you move an object in the section, the numbering changes. Outputting the colPos value as label happens if you don't fill the name/label field.

I mostly added it so that all the fields that have a special functionality based on their name (colPos, name) will also have an automatically added form field. The one thing you can't do from within a template is remove this field (that would require a custom Provider or usage of Flux hooks), but if you declare another field with that name, that field will replace the automatic one (logic sequence: is-content-container gets set on section object, fields get added when setter is called, any subsequent additions of a field with the same name replaces that field).

poveli:
How it is possible to render the content of the grids?
`<flux:content.render area="content.{iteration.index}" />` 
is deprecated now.

Thanks for your support!

PR #1583, opened 2018-07-16
[BUGFIX] Typo3 Issue 85013 workaround

glucka:
After this fix works with Typo3 ver. 8 too
@BenjaminBeck please review


NamelessCoder:
Merged manually - thanks for coming back to this one. Seems like I hadn't completely understood the purpose of the record data patching at first, but I do now :)

PR #1591, opened 2018-07-23
[BUGFIX] Increase integer size of colPos SQL column

andinger:
If you have a few more sites, the originally used SMALLINT column does not fit the longer colPos values

### DIFF
--- a/ext_tables.sql
+++ b/ext_tables.sql
@@ -2,6 +2,6 @@
 # Table structure for table 'tt_content'
 #
 CREATE TABLE tt_content (
-  tx_flux_migrated_version varchar(11) DEFAULT NULL
-  colPos BIGINT NOT NULL DEFAULT '0'
+  tx_flux_migrated_version varchar(11) DEFAULT NULL,
+  colPos bigint(20) DEFAULT '0' NOT NULL
 );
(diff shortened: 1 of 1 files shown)


NamelessCoder:
Thanks @andinger - looks okay to me.

NamelessCoder:
Adjusted schema to TYPO3 standards.

And for the record, `colPos` is *not* a `smallint(6)` unless you have gridelements installed. That extension forces the schema to use a very small int size for this field - which is surprising, given that for at least the last five years, that column has been `int(11)` in the TYPO3 core's schema. So, overriding it to a lesser size seems a bit destructive. You may want to report this to gridelements.

Even so, I'll do everyone a favor and increase this field to `bigint(20)` which will definitely leave room. And although the field is declared as `unsigned` in the core's schema, I will remove that restriction in Flux's schema.

Please be aware though that this only solves the problem if Flux gets loaded *after* gridelements. Otherwise, gridelements will override that field to make it `smallint(6)` again.

References:

* Original schema https://github.com/TYPO3/TYPO3.CMS/blob/master/typo3/sysext/frontend/ext_tables.sql#L177 (same state for this column since migrating from 4.x ext structures)
* Gridelements override of schema field: https://github.com/TYPO3-extensions/gridelements/blob/master/ext_tables.sql#L43

PR #1592, opened 2018-07-23
[BUGFIX] Implement section object colPos with JS

pkerling:
This is a RFC for the client-side JavaScript calculation

I noticed some odd behavior though: Sometimes if you delete a panel and add it again, its position is not saved correctly (it just appears as first entry int he list).


pkerling:
Note: I'd do some cleanup before actually suggesting this, this is mostly to get the idea across.

Classes/Form/Field/ColumnPosition.php
```diff
@@ -32,8 +32,8 @@ class ColumnPosition extends UserFunction
      */
     public function buildConfiguration()
     {
-        $fieldConfiguration = $this->prepareConfiguration('user');
-        $fieldConfiguration['userFunc'] = UserFunctions::class . '->renderColumnPositionField';
+        $fieldConfiguration = $this->prepareConfiguration('passthrough');
```
NamelessCoder: Causes the field to not be rendered at all - type must be `user`. Since it's the same amount of code and we don't need to access the FormEngine specifics, I would prefer to keep this as part of the UserFunctions. If nothing else then to avoid registering a node type globally which will only be used in very specific cases.

Classes/Integration/FormEngine/ColposElement.php
```diff
@@ -0,0 +1,101 @@
...
+
+        if ($inputValue === '') {
+            $rowUid = $this->data['databaseRow']['uid'];
+            if (!isset($rowUid) || !is_int($rowUid)) {
+                $rowUid = 0;
+            }
+            $usedColumnPositions = $this->determineUsedColumnPositionsWithinParent($this->data['tableName'], $rowUid);
+            $usedColumnPositionsJS = '[' . implode(',', $usedColumnPositions) . ']';
+            // FIXME better constant?
+            $maximumColPos = 99;
```
NamelessCoder: `ColumnNumberUtility::MULTIPLIER` can be used (value: `100`).

Classes/Integration/FormEngine/ColposElement.php
```diff
@@ -0,0 +1,101 @@
...
+
+            $resultArray['additionalJavaScriptPost'][] = sprintf("(function() {
```
NamelessCoder: If possible, the JS should be in a RequireJS module so we only load the module. The JS should then read all of the necessary variables from DOM properties of our hidden field. There is an existing such module loaded by the `Preview` class which interacts with our grids - perhaps you can use that as reference (and somehow make it also listen for newly added DOM nodes instead of just running once after page load).

NamelessCoder:
Peculiarity: on TYPO3 9.3 I was not able to reproduce the error you described when removing a panel. And it definitely should not be happening so that may even be a TYPO3 core bug. In any case, I don't see how the script parts you added could cause that problem...

Classes/Integration/FormEngine/ColposElement.php
pkerling: > The JS should then read all of the necessary variables from DOM properties of our hidden field.

OK, changing the data to DOM properties should be easy.

> (and somehow make it also listen for newly added DOM nodes instead of just running once after page load)

This does not run after page load, it runs each time you add a new panel.

NamelessCoder: > This does not run after page load, it runs each time you add a new panel.

This was about the grid require-js module, not the code you wrote. It needs to behave like you wrote it, but should happen with listeners instead of ad-hoc execution in an inserted script block ;)

Classes/Form/Field/ColumnPosition.php
pkerling: > Causes the field to not be rendered at all - type must be user. 

Rendering fine here on 8.7 - are you saying it's not rendered on 9.3?

> Since it's the same amount of code and we don't need to access the FormEngine specifics, I would prefer to keep this as part of the UserFunctions.

I converted it to a custom FormEngine field so we can set `additionalJavaScriptPost`, which is not possible with a `user` field. Of course we could just insert `<script>` tags into the rendered HTML in the user function instead, but I'm not sure how "clean" that is. Other option, as you mentioned, would be to have a global JS loaded and use `MutationObserver` or sth. similar to react to DOM changes. Also not that nice IMO but since the FormEngine js does not have any hooks to wire this up directly, it could very well be the best solution.

NamelessCoder: > Rendering fine here on 8.7 - are you saying it's not rendered on 9.3?

Yes - when type is `passthrough` there is no HTML output on 9.3 (possibly all of 9.x series).

Classes/Integration/FormEngine/ColposElement.php
pkerling: Ah, I misread the context. Thanks for clarifying!

Classes/Form/Field/ColumnPosition.php
NamelessCoder: > FormEngine js does not have any hooks to wire this up directly

That is exactly the problem. In `Preview` we do it like this - and I'd be fine with that same strategy in this user function:

```php
    /**
     * @return void
     */
    protected function attachAssets()
    {
        if (false === static::$assetsIncluded) {
            $doc = GeneralUtility::makeInstance(ModuleTemplate::class);
            $doc->backPath = $GLOBALS['BACK_PATH'] ?? '';

            /** @var PageRenderer $pageRenderer */
            $pageRenderer = $doc->getPageRenderer();

            $fullJsPath = PathUtility::getRelativePath(
                PATH_typo3,
                GeneralUtility::getFileAbsFileName('EXT:flux/Resources/Public/js/')
            );

            // requirejs
            $pageRenderer->addRequireJsConfiguration([
                'paths' => [
                    'FluidTypo3/Flux/FluxCollapse' => $fullJsPath . 'fluxCollapse',
                ],
            ]);
            $pageRenderer->loadRequireJsModule('FluidTypo3/Flux/FluxCollapse');

            static::$assetsIncluded = true;
        }
    }
```

pkerling: OK, then I'll try it. Do you know of an easy way to get information attached to the root form? I don't want to hard-code stuff like the maximum colPos in the JS, but have to get it from somewhere.

NamelessCoder: I'd say we just throw all of the properties on the `<input type="hidden" />` tag as `data-` properties, so we've got `data-occupied-values`, `data-min-value`, `data-max-value` for example. The values will be the same (until parent is saved) but a tiny bit of duplication in local scope is I think preferable here, as opposed to setting these on a higher level of the rendered FormEngine stack.

pkerling:
@NamelessCoder updated

I think the RequireJS script registration can be done way simpler, at least in TYPO3 v8. Please have a look. I have a patch to do it the same in the `Preview` hook subscriber and it also seems to work.

Also, you were calculating the current value of the hidden input field quite complicatedly before. Is there a reason you did not use `itemFormElValue`? Will it not work in all circumstances?

NamelessCoder:
It really is a pleasure to read/review code that's this well-structured and has appropriate code comments!

PR #1595, opened 2018-07-29
[BUGFIX] Fix wrong usage of cache API in ContentIcon hook subscriber

pkerling:
cache->get returns `false` if a key cannot be found - not `null`, so the
code will always assume that a valid cache entry was found and never
return any useful information.

Broken in 3d45411368d39afa9606a0e21d3a7347623d5864


NamelessCoder:
Oops - looks like I forgot about this PR and went ahead and did this change myself. Sorry @pkerling (also for making you wait that long!)

PR #1608, opened 2018-08-24
[BUGFIX] Show column label instead of colPos (page translation view)

BenjaminBeck:


NamelessCoder:
Thanks for opening this task - I noticed it some time after 9.0.1 was released but never got around to fixing it.

I think there's a way to solve it without using a hook, which isn't in itself a bad solution, but this particular one would execute very frequently and would re-query values that already existed when the backend layout array was generated. TYPO3 will try to read the label from the list of allowed backend columns so the problem is that these colPos values are not in the the array for some reason.

I'm not sure it *can* be solved that way, but perhaps the right method is to detect the language that gets rendered and somehow pass that to the method building the backend layout, so it can add those values to the array that contains the selector options and labels. But if that's possible, we can completely avoid additional SQL queries and hooks ;)

BenjaminBeck:
You are totally right! - There is a mutch simpler solution  and i found it (:

NamelessCoder:
Okay that was quite a bit simpler! This needs to be tested with the colPos value selection box when editing content in non-default languages, but it looks like the right way :)

NamelessCoder:
Maybe it shouldn't happen in the PreviewView, though. Perhaps we should be doing this in another place - and maybe more than one place. If possible we should avoid mutating the TCA directly.

BenjaminBeck:
Normally this code should not be needed. We already have a hook for that 
 ``$GLOBALS['TCA']['tt_content']['columns']['colPos']['config']['itemsProcFunc'] =
    \FluidTYPO3\Flux\Integration\Overrides\BackendLayoutView::class . '->colPosListItemProcFunc';``
But the code that resolves the labels does not execute the proc-function.
The really correct way would be to change that within TYPO3 - but until then i would use this one-liner.

Shall i squash the commits?

PR #1622, opened 2018-09-26
[BUGFIX] Add pageTSconfig for content types in frontend

rupasix:
According to recent changes in flux with colpos, we are trying to combine flux with ext:frontend_editing. At least in the basic version - support for new content elements and drop zones for flux:grid.column. The first problem is access to elements - they are not visible in the frontend_editing content element list. Flux only creates them for the backend (https://github.com/FluidTYPO3/flux/blob/development/Classes/Integration/ContentTypeBuilder.php#L249). Can we remove condition or expand it to include frontend_editing  (user settings or taking into account ts config.frontend_editing)? What consequences can create a modification of this condition?


NamelessCoder:
Re-labeled to be more generic as not only frontend_editing might benefit. Labeled BUGFIX to include in change log.

Issue #1604, opened 2018-08-10
flux:field.text name="tt_content.bodytext" does not work again


glucka:
just like #1250 
with latest flux

advOpk:
Flux version 9.x breaks tt_content.* named form fields to be saved in database.  

Previously in 8.x \FluidTYPO3\Flux\Backend\TceMain::processCmdmap_preProcess executed executeConfigurationProviderMethod('preProcessCommand').

Now \FluidTYPO3\Flux\Integration\HookSubscribers\DataHandlerSubscriber::processDatamap_preProcessFieldArray is dealing only with colPos typo3 issues [[BUGFIX] Add workaround for forge Issue 85013](https://github.com/FluidTYPO3/flux/commit/ec0cf25f6f4e9de07d8a309e2d5471afe2e70224#diff-93ddaa2e34830c84cb681ad5c240d4b1)

glucka:
Solved by #1636

PR #1626, opened 2018-10-03
tt_content.* named form fields

glucka:
Save tt_content.* named form fields direct in db columns
Solve #1604


glucka:
Unfortunately this works only with tt_content. I make a better version.

PR #1627, opened 2018-10-04
tt_content.* or pages.* named form fields

glucka:
Save tt_content.* or pages.* named form fields direct in db columns


NamelessCoder:
This one needs to be solved by adding quite a few loops, unfortunately.

* An outer loop must iterate every field of $fieldArray
* Then `$GLOBALS['TCA'][$table]['columns'][$fieldName]['type'] === 'flex'` must be checked and only if it is true
* Then it must be tested if at least one provider will be returned, possibly one that implements a specific ProviderInterface (I can take a closer look if you need!)
* Only if at least one provider is returned for the table and field combination, should the data be iterated over with the next two loops, which:
* Iterates over `$fieldArray[$fieldName]['data']` in order to read all sheets
* Iterates over the fields in the sheet and does the checking and transferring of value you do now.

The goal is to make this completely dynamic and not assume any field names. That means we have to iterate all the fields and check in TCA if the field might have a provider (= is a flex type field), and then check if that flex type field has a Provider (because we don't want to do this processing for "normal" FlexForms).

Lots and lots of loops... but it should execute quick enough and there is likely only one or two "flex" type fields in any given record, so it shouldn't be too bad :)

glucka:
Thanks for the hints! You're right. I correct it.

NamelessCoder:
You were almost there with this PR. I had two comments pending:

1. The resolver instance can be created once, outside the outermost loop, since it's fine to reuse that instance.
2. The check for field name matching table prefix should be a bit different:

```php
list ($possibleTableName, $columnName) = explode('.', $key, 2);
if ($possibleTableName === $table && isset($GLOBALS['TCA'][$table]['columns'][$columnName])) {
    $fieldArray[$columnName] = $value['vDEF'];
}
```

glucka:
Thanks @NamelessCoder! Its a problem that i make so many PR-s? :)

NamelessCoder:
Nono, not a problem at all! It's actually a great way to have Travis test the changes you make. It's also fine if you just keep adding commits to the same PR, we can squash them into one commit when merging :)

PR #1628, opened 2018-10-04
[BUGFIX] Dbtable.* named form fields

glucka:
Save dbtable.* named flux form fields direct in db columns
Solve #1604

### DIFF
--- a/Classes/Integration/Overrides/BackendLayoutView.php
+++ b/Classes/Integration/Overrides/BackendLayoutView.php
@@ -101,7 +101,7 @@ class BackendLayoutView extends \TYPO3\CMS\Backend\View\BackendLayoutView
         }
         if ($this->addingItemsForContent) {
             $parentRecordUid = ColumnNumberUtility::calculateParentUid($this->record['colPos']);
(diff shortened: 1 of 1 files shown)


NamelessCoder:
Looks good!

glucka:
Should/can i merge? :)

NamelessCoder:
Please hang on, I'd like to run a test on this so we have two confirmations. If it works okay I'll merge ;)

Generally speaking we don't merge our own PRs, even when approved.

glucka:
Ok, understood ;)

NamelessCoder:
I this case it's because you created the commit on the `development` branch which means you can't create additional PRs - doing that will include the old commits and detect the new PR as a replacement.

The way to avoid this is to always create a branch starting from `development`, for example `feature/my-special-fix` and have one such branch for each PR.

If you run into the situation later on that you need to do work which depends on a PR that isn't merged, feel free to ping me for more git-methods. But until you're more comfortable with one-branch-per-PR I would recommend keeping the changes small and separate, and avoid making changes depend on each other.

PR #1633, opened 2018-10-11
[TASK] Allow edit the containers with the colPos < 0

vruescas:
If the colPos of one element is minor than 0 is not possible edit this element. My update allow edit the elements with a colPos < 0.


NamelessCoder:
Thanks @vruescas - great pick for a first contribution :)

PR #1634, opened 2018-10-17
[TASK] Add renderType to Form\Field\Tree.

sven-carstens-udg:
Form/Field/Tree has not set the renderType and there are no getters or setters to add it.

The renderType is mandatory.


NamelessCoder:
Thanks Sven - this component is so rarely used I guess nobody had a problem with it until now. Patch will be included in next release :)

PR #1635, opened 2018-10-18
[BUGFIX] Remove HiddenRestriction in Preview Mode

glucka:


NamelessCoder:
I'm not sure that:

* It is a good idea to place this in the RecordService. If this is used to load records from a table that has no `hidden` field, will it cause exceptions to remove a restriction that doesn't exist?
* That doing so fixes all occurrences that may select records - I think there are cases where we use QueryBuilder directly?

Another thing that may not mix well with this, is workspaces. If you hide a record in a workspace draft and then preview that, the expected result actually is that the `hidden` restriction is respected and you don't see the content element or other.

I still have a hunch that there is a method we can use for this, in TYPO3 itself, that will allow us to read records' raw data - not rendered versions - and have all of these restrictions an workspace overlays done for us. So I think it's time I take a closer look if we can perhaps make RecordService into a wrapper for a core method and not use QueryBuilder ;)

glucka:
"It is a good idea to place this in the RecordService. If this is used to load records from a table that has no hidden field, will it cause exceptions to remove a restriction that doesn't exist?"
Do not cause exceptions because this removes only ```hidden != 0``` from the sql query.

NamelessCoder:
It might be okay for the `getSingle` method - but we do have places where we can pick multiple records that might need to be previewed, like `flux:content.get`. I will consider this thoroughly :)

PR #1636, opened 2018-10-19
[BUGFIX] Restore $table.$field naming transfer to record

NamelessCoder:
Saves fields prefixed with the table name and which exist
in TCA, into the record fields directly. This patch restores
a functionality previously bound to Providers.


NamelessCoder:
@glucka Here's an easy one - this is your previous patch restored and CGL-adjusted. No changes except for the extended comment and if you agree with that, you're welcome to press "Squash and merge" to merge your first PR into Flux :)

NamelessCoder:
🎉 🎈 ^^

PR #1639, opened 2018-10-22
Add instructions for installation

rotespferd:
Add instructions to install the flux extension via Composer or the TER.


NamelessCoder:
Thanks for this and the other one, @rotespferd :)

Issue #1645, opened 2018-11-01
Problems with nested grid elements when migrating from fluidcontent to flux


fnagel:
I'm having a hard time migrating a TYPO3 8.7.20 instance from EXT:fluidcontent (5.2.0 with flux v8.2.1) to EXT:flux (9.1.0). 

## Migration path

Here's what I do:

* Update my template (see below for new and old version)
* Change `Ctype` using SQL: `UPDATE tt_content SET CType = 'vendor_row' WHERE CType = 'fluidcontent_content' AND tx_fed_fcefile = 'Ns.Vendor:Row.html'`
* Run the EM migration wizard in EXT:flux

## Functionality 

What happens:

* Non nested grid rows (existing and newly created ones) work as expexted
* Nested grid rows with other content elements throw error messages in backend but existing ones seem to work in frontend

## Screenshots
![2018-11-01 21_59_34-vm typo3 cms 8 7 20](https://user-images.githubusercontent.com/116824/47881690-26bfe080-de27-11e8-9234-8f91de8b8824.png)
![2018-11-01 22_05_51-vm typo3 cms 8 7 20](https://user-images.githubusercontent.com/116824/47881691-27587700-de27-11e8-82eb-874439b745cf.png)
![2018-11-01 22_09_45-vm typo3 cms 8 7 20](https://user-images.githubusercontent.com/116824/47881692-27587700-de27-11e8-95e4-f8e55875a1d4.png)

## Error messages

Error message: 
``
Argument 1 passed to FluidTYPO3\Flux\Form\Container\Grid::buildExtendedBackendLayoutArray() must be of the type integer, array given, called in /var/www/typo3conf/ext/flux/Classes/Integration/Overrides/BackendLayoutView.php on line 76 
``

Sometimes this error message pops up (example):
``
[...]

PR #1648, opened 2018-11-10
[BUGFIX] Handle expanded records from FormEngine

NamelessCoder:
This patch adds an additional check for a UID value possibly being contained in an array, which happens when a record is passed from FormEngine to the Flux override for BackendLayoutView.

Though this doesn’t happen in core code, it may occur in third party code that re-uses the record FormEngine expanded and calls core functions like preview rendering with that record as input. Flux can easily handle the case.

References: #1645

### DIFF
--- a/class.ext_update.php
+++ b/class.ext_update.php
@@ -5,6 +5,12 @@ use FluidTYPO3\Flux\Provider\Interfaces\FluidProviderInterface;
 use FluidTYPO3\Flux\Provider\Interfaces\GridProviderInterface;
 use FluidTYPO3\Flux\Utility\ColumnNumberUtility;
 use TYPO3\CMS\Core\Database\ConnectionPool;
+use TYPO3\CMS\Core\Database\Query\Restriction\BackendWorkspaceRestriction;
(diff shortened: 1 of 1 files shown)


fnagel:
@NamelessCoder I used the "Update branch" button but this creates and ugly merge commit instead of doing a rebase. Not sure if you want to do this manually to keep the git history more clean...

NamelessCoder:
@fnagel The "Merge" button should by default say "Squash and merge" (I'm not sure if your account settings can override that, but repo uses squash by default). If you do that, the subject and body fields' contents become the commit subject and body and creates a single commit without any of the merge commits.

* When there's only a single commit (and any number of merge commits) then that commit message is used.
* When there are more commits they get concatenated in the body field and we like to clean that up so we only get the body from the main commit.
* The title of the PR (taken from first commit) thread is copied to the subject field - I sometimes use this to give PRs a better topic that matches the commit completely, by editing title right before squash-merge.

You'll get the hang of it soon enough I'm sure - this PR is a good one to test out the squash-and-merge since you won't have to clean up the commit body field ;)

PR #1649, opened 2018-11-10
[BUGFIX] Skip migration of deleted content records

NamelessCoder:
Do not remove the soft-delete restriction, but still remove all other restrictions for selecting content records to be migrated. This reduces the need to clean up the database before running the migration to avoid problems with old content that cannot resolve parent/child relations correctly for whichever reasons.

References: #1645


fnagel:
I've rebased and added a commit here: https://github.com/FluidTYPO3/flux/pull/1650

PR #1650, opened 2018-11-12
[BUGFIX] Skip migration of deleted content records

fnagel:
Follow up to https://github.com/FluidTYPO3/flux/pull/1649


NamelessCoder:
Makes perfect sense - if we don't migrate the records, we don't want to report them as possibly conflicting.

PR #1653, opened 2018-12-08
[FEATURE] Content type API and record-based CTypes

NamelessCoder:
This patch adds new API to use arbitrary storage formats to
store content type definitions which Flux will then render,
with support for Flux fields, nested content and so on just
like template-based content types.

An implementation based on this new API is then added which
allows site administrators to configure content types without
opening a code editor - by creating content type records.
The editing interface for content types is then tightly coupled
with validation to prevent misconfigured CTypes, and export
capability to move finished content types to Flux-based fluid
template files that are ideal for version control.


NamelessCoder:
@glucka / @BenjaminBeck I'd especially like you guys' input about this: would this implementation be sufficient for most use cases? Are there any features you'd miss the ability to use when using this API? Do you have suggestions for additional storage formats we could support?

I've got one more integration I'd like to make - a convention that looks for and if found, includes `EXT:providerextensionkey/Configuration/Flux/ContentTypes.php` which would then return an array of `ContentTypeDefinitionInterface` objects that a developer could create from any source. Then perhaps I will move the `RecordBasedContentTypeDefinition` integration from ext_localconf.php to this convention-based inclusion file inside the Flux extension dir.

NamelessCoder:
screenshot 2018-12-08 at 16 54 42
screenshot 2018-12-08 at 16 55 09
screenshot 2018-12-08 at 16 55 26
screenshot 2018-12-08 at 16 55 50
screenshot 2018-12-08 at 16 57 12
screenshot 2018-12-08 at 16 57 24

BenjaminBeck:
This function would be a great complement to the current implementation.
It would reduce the effort to create a custom content element. More specialized content elements and associated with that, fewer options, make it easier for editors.
But I don't like the idea of storing code in the database because I have all projects under version control.
It will take me a while until I have more time and can actually test it.
Here is a screenshot of a project. To have larger thumbnails, I added CSS over a backend-hook and exchanged the icons for a layout image. In my opinion the picture with the assignments of the field names to the respective positions is very good to select the right content element. The image is also displayed in the edit mask. It would be fantastic to have something like this included.
2019-01-23 21_57_55-snipping tool

NamelessCoder:
@BenjaminBeck I like the idea of enabling larger thumbnails for content - but I don't see that as necessarily a Flux feature. IIRC, Flux passes thumbnail images through without resizing, as does the new content wizard itself, so it should be possible to provide such a feature as a separate extension which does just that one single thing (and maybe allows you to configure a maximum allowed width/height and constraints to aspect).

>  I added CSS over a backend-hook

That is essentially all you'd need to do. Perhaps a bit much for a separate extension - IMHO this isn't something that would be unreasonable to simply add manually in each project that requires larger thumbnails. If larger thumbnails are requested for BE that may very well not be the only CSS thing you need to change in BE so having a custom CSS file for BE is usually a good thing ;)

Issue #1632, opened 2018-10-10
Copy & Paste child elements


mithri:
Hi, we have problems with moving translated child within flux grid columns. They are not moved or copied when nested. I`ve made a screenshot to show the problem.  Only the first child elements gets moved when moving a container element. 

The copies of all the elements are created, but not with the right colpos. After copying/moving the container elements to another page and then moving the single elements around, they appear in the right position. So this is only the matter when copying/moving to another page or copy a whole page.

I hope the image explans it. If someone needs sponsoring for this to fix, please ask.

Thanks! 
Tobias

![screen](https://user-images.githubusercontent.com/5941531/46738841-77687180-cc9f-11e8-99f1-88cf37ea42fd.png)

tobiasgraeber:
Hi @NamelessCoder,

first of all, thanks for your great work. As @mithri already points out above, we have the same problem.
We try to copy a page (containing multiple nested elements) in full as usual, but after copy, looking on the page elements, only those elements copied correctly which have a nesting of max 2 elements (so 1 container and 1 child element). If there are elements, with more childs, say 1 container, 1 child inside that container and another one inside the child - these (last childs) won´t get copied (seems like the same as in the screenshot above).

Info:
We noticed, that you could work around the issue, if you copy the elements one-by-one or even all at once (wrapping all of them into one temporary content element/container), if you dont copy the page itself but copy the single element (containing multiple nested elements) "by hand/manually" from one page to another.

So it seems there is a general problem in copying nested elements by using the "page copy" feature.

Could you maybe investigate into that / give this topic a higher priority?
We are currently facing the challenge to copy one page (as template) for up to 20-30 other pages, and doing this "by hand" would be an enormous higher effort (and also would take much much more time and is much more error-prone.

Wourd really really appreciate to hear from you (and anyone else) regarding this topic!
Thank you very much in advance!

NamelessCoder:
Hey @tobiasgraeber,

> Could you please investigate into this topic / give this topic a higher priority?

I'm afraid I don't currently have the time to invest - changing the priority wouldn't help with that. Meanwhile you can help identify the more exact cause or line of code that for whatever reason stops or fails to continue recursion.

You may end up finding that it requires a specific composition of records to trigger it.

I would also look for differences in the instance of DataHandler and its properties, as it is received by Flux's hooks in the case that works and the other that doesn't.

[...]

PR #1659, opened 2019-01-05
[BUGFIX] Copy & Paste child elements

glucka:
Should solve #1632

### DIFF
--- a/Classes/Integration/HookSubscribers/DataHandlerSubscriber.php
+++ b/Classes/Integration/HookSubscribers/DataHandlerSubscriber.php
@@ -46,10 +46,17 @@ class DataHandlerSubscriber
                         $mostRecentCopyOfParentRecord = $this->getLastCopiedVersionOfRecordInLanguage(
                             $table,
                             (int)$originalParentRecordUid,
-                            0,
+                            (int)$fieldArray['sys_language_uid'],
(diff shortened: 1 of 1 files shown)


glucka:
I think this works only in Connected Mode.
I check it in Copy Mode tomorrow.

glucka:
sorry i find  another bugs... i work on it

NamelessCoder:
The last patch state looks like it was going in the right direction - though it might be best to split the patching up into two: one to solve the free-mode behavior and another to solve the copy-mode. It's currently a little difficult to tell which changes aim to solve which behavior.

PS: I think perhaps the bug you encountered last happened because this line was changed: https://github.com/FluidTYPO3/flux/pull/1659/files#diff-93ddaa2e34830c84cb681ad5c240d4b1L42 - there's a super-important difference in how colPos values are handled: only when a record is in "copy" mode should we re-calculate the colPos like this. In "connected" mode the colPos value is the same as the original language parent's colPos, in "free" mode it's relative to the translated parent. This code *probably* should still not be executed in connected mode ;)

glucka:
Thanks for your hints Claus! You are right. 
Unfortunately i think cascading the copy cmdmap is not the right way. We should always use the $reference->copyMappingArray.
I continue to work on it!

Issue #1664, opened 2019-01-11
RecordService::getSingle does not remove restrictions for query


wazum:
TYPO3 CMS: 9.5.3
flux: 9.1.0
fluidpages: 4.3.0

InvalidTemplateResourceException for preview of hidden pages.

Problem is that \FluidTYPO3\Flux\Service\RecordService::getSingle returns false in this case (and not a record, even if 'Show hidden pages' is checked in the admin panel) and therefor the page configuration (up the root line) cannot be loaded.

wazum:
The old method looked like this
```    /**
     * @param string $table
     * @param string $fields
     * @param string $uid
     * @return array|NULL
     */
    public function getSingle($table, $fields, $uid)
    {
        $connection = $this->getDatabaseConnection();
        $record = $connection->exec_SELECTgetSingleRow($fields, $table, "uid = '" . intval($uid) . "'");
        return false !== $record ? $record : null;
    }
```

monosize:
@NamelessCoder The commit 482dd13 is not compatible for TYPO3 version 8.7.x.
This works only in TYPO3 since version 9.
Classes TYPO3\CMS\Core\Context\Context and TYPO3\CMS\Core\Context\VisibilityAspect doesn't exist in prior TYPO3 versions. 

This poses a problem because the current master version does not work optimally in 8.7.x. And it is always suggested that the development branch should be used after all, as there is no newer release yet.

A version check should be inserted here, because the development branch is marked for use with TYPO3 8.7.x.

NamelessCoder:
@monosize If you know the solution to use on 8.7 (without going back to `$GLOBALS['TYPO3_DB']`) I'd very much appreciate a pull request. Checking for version of TYPO3 or presence of VisibilityAspect class is fine for me, either way you choose.

We probably want the version check / class presence condition around https://github.com/FluidTYPO3/flux/commit/482dd13510688791d1cd551f68f35f6a6dde9721#diff-0aa463a9e905b968eb08165841e9c3e7R132 and create $frontendRestrictions in a different way on 8.7?

PR #1665, opened 2019-01-11
[BUGFIX] Remove restrictions for query

wazum:
Fix for #1664


NamelessCoder:
Implemented in a slightly different version in https://github.com/FluidTYPO3/flux/commit/482dd13510688791d1cd551f68f35f6a6dde9721 which accommodates preview requests specially and senses the restrictions to apply based on request context.

PR #1677, opened 2019-03-13
[BUGFIX] Enables renderType for checkbox field

jmverges:
Adds renderType to checkbox fields


NamelessCoder:
Replaced by https://github.com/FluidTYPO3/flux/pull/1687 which allows this option to be passed as part of an array of raw options for any field. The longer term plan is to avoid having to change Flux's API whenever TCA changes.

Issue #1683, opened 2019-03-21
Issue with copying localized grid elements


fnagel:
I'm having issues when copying localized grid elements in TYPO3 8.7.24 with EXT:flux (1d4642384f3cda38f4a5e1ae5ab55dd4cabec073). I've tested this in a custom installation and a fresh introduction package instance.

## How to reproduce

Here's what I do:

* Create a new page
* Add a grid (see template below)
* Add content within the grid element
* Localize the grid and its content
* Copy that grid element
* Localized grid content is placed within the default language grid (see screenshot)

## Screenshots

![1](https://user-images.githubusercontent.com/116824/54743792-7d0aa180-4bc5-11e9-9bbe-75e2c88a6acd.PNG)
![2](https://user-images.githubusercontent.com/116824/54743794-7d0aa180-4bc5-11e9-9fc7-81aac51ccd18.PNG)
![3](https://user-images.githubusercontent.com/116824/54743795-7d0aa180-4bc5-11e9-941f-05844faaa491.PNG)
![4](https://user-images.githubusercontent.com/116824/54743796-7d0aa180-4bc5-11e9-872a-f372ec5f6a31.PNG)
![5](https://user-images.githubusercontent.com/116824/54743797-7d0aa180-4bc5-11e9-82f8-debd0bf2d361.PNG)
![6](https://user-images.githubusercontent.com/116824/54743798-7da33800-4bc5-11e9-86d5-d239244fccc9.PNG)

## Template (simplified)

	<f:section name="Configuration">
		<v:variable.set name="columnItems" value="{
			null: 'Default (equal width)',
			'12': 'Full width',
			'10': '5/6 width',
			'9': '3/4 width',
			'8': '2/3 width',
			'6': '1/2 width',
			'4': '1/3 width',
			'3': '1/4 width',
			'2': '1/6 width'
		}" />

		<flux:form id="row" label="Grid Row" description="Add content in columns.">
			<flux:form.option.group value="common" />
			<flux:form.option.icon>EXT:project/Resources/Public/Icons/grid.svg</flux:form.option.icon>

			<flux:form.section name="columns">
				<flux:form.object name="column" label="Column">
					<flux:field.select label="Large devices" name="class-large" items="{columnItems}" />
					<flux:field.select label="Medium devices" name="class-medium" items="{columnItems}" />
					<flux:field.select label="Small devices" name="class" items="{columnItems}" />
					<flux:field.checkbox name="center" label="Center column" />
[...]

PR #1684, opened 2019-03-21
[BUGFIX] Copy & Paste localized child elements

glucka:
Should solve #1683


NamelessCoder:
Thanks for taking care of this @glucka - and thanks for testing @fnagel!

rupasix:
Late for party but yeah ... i can confirm that translating page on typo3 9 is working as intended. Thanks!

PR #1687, opened 2019-03-30
[FEATURE] Allow arbitrary TCA "config" array on fields

NamelessCoder:
Allows users to pass "config" to Flux form fields and
provide any TCA option, including those that are not
supported as ViewHelper attributes. Values specified
in this array will have priority over other properties.


NamelessCoder:
Self-merging this minor change so a release can be created.

Issue #1691, opened 2019-04-06
Wrong template is used when a custom ContentController is used


maximilian-walter:
I have updated Flux from 9.1.0 to. 9.2.0 today and now I'm getting the following Exception:

> #1257246929 TYPO3Fluid\Fluid\View\Exception\InvalidTemplateResourceException
Tried resolving a template file for controller action "Content->recordlist" in format ".html", but none of the paths contained the expected template file (Content/Recordlist.html). The following paths were checked: /public/typo3conf/ext/example_site/Resources/Private/Templates/

The following template causes this Exception:
`/public/typo3conf/ext/example_site/Resources/Private/Templates/Content/RecordList.html`.

-> The problem is the generated filename: It should be `RecordList.html`, but Fluid uses`"Recordlist.html` (with a lowercase "L").

My templates are registered in the file `ext_localconf.php` with the following call:

```php
`\FluidTYPO3\Flux\Core::registerProviderExtensionKey('MyVendor.ExampleSite', 'Content');`.
```

I'm using a custom ContentController `\MyVendor\ExampleSite\Controller\ContentController` which contains a action for this template:

```php
<?php
declare(strict_types=1);
namespace MyVendor\ExampleSite\Controller;

use FluidTYPO3\Flux\Controller\AbstractFluxController;

class ContentController extends AbstractFluxController
{
    public function recordListAction(): void
    {
        $this->view->assign('records', $this->loadRecords($this->getData()));
    }

    protected function loadRecords(array $data): array
    {
        // Load records from database
        return $loadedData;
    }
}
```

When the action is removed everything works as expected (but the template has no data...).

-> Flux behaves differently when the ContentController has a action for a template. I tried to find the change between 9.1.0 and 9.2.0 which causes this bug and I gues the following commit introduced this regression:
https://github.com/FluidTYPO3/flux/commit/c6c17094ec932a93576406a2d84f2f60cd3561c9

The method `\FluidTYPO3\Flux\Integration\ContentTypeBuilder::configureContentTypeFromTemplateFile()` has a variable $emulatedControllerAction which contains the wrong value ("Recordlist" instead of "RecordList") when $contentType is not null.
https://github.com/FluidTYPO3/flux/blob/c6c17094ec932a93576406a2d84f2f60cd3561c9/Classes/Integration/ContentTypeBuilder.php#L60

[...]

PR #1711, opened 2019-08-20
[BUGFIX] Support camel case with custom content controller action

Garbanas:
This adresses #1691 where camel case template filenames are no longer respected when being used in conjunction with an action in a custom content controller.

This PR tries to preserve the controller action name (which in turn defines the template name) only if the content type and template filename are the same in a case insensitive match.


NamelessCoder:
@Garbanas thanks for the patch - could you enable the "allow changes from repository owner" flag, or rebase on `development` and do a force-push? Thanks in advance :)

Garbanas:
@NamelessCoder The flag "Allow edits by maintainers" is on but I also just merged the latest changes from development

PR #1715, opened 2019-09-16
[BUGFIX] Handle memory leaks on large mirgation data sets

onigunn:
Migrating a large data set of content elements can/will cause memory exhausted
exceptions. Using `unset` on `$row` and `$grid` variables deals with this. Migration of an installation with over 25000 entries runs fine after this changes.


NamelessCoder:
thanks @onigunn!

PR #1716, opened 2019-09-25
[BUGFIX] Fixed language handling for nesting

Crontis:
Used TYPO3 Version: 8.7.27
Used Flux Version: dev-development

Having a parent record with sys_language_uid -1 and a child
record with sys_language_uid 0 will result in the child record
not being shown. This happens because till now the respected
sys_language_uid of the child record will be set to the parent
record, which would be -1. This results in a query looking for
the record with sys_language_uid = -1, which dosn't exist.

The normal behaviour should respect the currently selected
language in the Backend as in the Backend/PageLayoutController
shown in Line 926 (TYPO3 8).


NamelessCoder:
Thanks @Crontis - I made a little change to use the module data selection instead of the deprecated SOBE global which apparently is removed on TYPO3v10.

Issue #1704, opened 2019-06-17
[BUG][TYPO3-9.5.7]Deleted tt_content records are referenced on copy to other page


rupasix:
Hello,
We came across an error while copying content to another page:
![image](https://user-images.githubusercontent.com/9860558/59584933-97c6b680-90df-11e9-9a97-3ec2b0ef9131.png)
After a short investigation, we came to the fact that the original content and copy have a different uid than those from the message. Going further, it turned out that uid from the message are deleted=1 and are the master content for copied content. We came to the conclusion that Flux wrongly (or rightly?) does not ignore the deleted content. We decided that this line is a good starting point for this error:
https://github.com/FluidTYPO3/flux/blob/5380e40d07165a46e4da7c3d0a41d8b2eb00abfd/Classes/Integration/HookSubscribers/DataHandlerSubscriber.php#L353

Effect: 
The content is copied correctly but an error appears which stops the page refresh procedure. This is confusing for the editor. After deleting permamently the records with flag "deleted = 1" the copying process takes place without any problem.

We are using:
Typo3: 9.5.7
Flux: 9.2.0 stable
VHS: 5.2.0

If we can still help somehow, let me know.

hieslc:
Exactly the same problem here.

glucka:
Thanks for the contributing. Im going to check this next week.

nitori:
Is here potentially a regression? I'm getting exactly this error with the same steps above on TYPO3 10.4.13 with Flux 9.4.2

It seems to have been reverted in 77d66df04cec886f8d7292069e176846f8f7d486 due to issue https://github.com/FluidTYPO3/flux/issues/1745

PR #1719, opened 2019-10-09
[BUGFIX] Do not try to copy deleted records

cweiske:
When trying to copy child records, do not copy deleted ones.

Steps to reproduce:
1. Create an element that allows children (grid)
2. Create a child element
3. Delete child element
4. Copy and paste the main element
5. Error occurs, since deleted child element is tried to be copied.

Close: #1704


kh-schmidt:
I don't understand: 

This Bug is fixed in flux 9.3.2:
https://github.com/FluidTYPO3/flux/blob/9.3.2/Classes/Integration/HookSubscribers/DataHandlerSubscriber.php

But not in flux 9.4+:
https://github.com/FluidTYPO3/flux/blob/9.4.0/Classes/Integration/HookSubscribers/DataHandlerSubscriber.php

Is there a merge missing?

PR #1720, opened 2019-10-09
[BUGFIX] Adjust colPos of workspace overlay records of copied content…

cweiske:
… elements

When copying a page, the colPos of nested tt_content elements is adjusted
automatically.
Inside a workspace, TYPO3's DataHandler automatically creates shadow records
for this workspace. Their colPos needs to be adjusted, too.

Steps to reproduce:
1. Create a page, a container element and a child element
2. Switch into a workspace and copy&paste the page
3. In the pasted page, the child element is not visible


NamelessCoder:
Perfect as usual - thanks Christian!

Issue #1726, opened 2019-10-29
Class does not exist. Reflection failed.


filoucrackeur:
version
Flux:  dev-development
Fluid: 2.6.4

#1278450972 TYPO3\CMS\Extbase\Reflection\Exception\UnknownClassException
Class does not exist. Reflection failed.

in this method in AbastractFluxController line 273

```php
    /**
     * Prepares a view for the current action.
     * By default, this method tries to locate a view with a name matching the current action.
     *
     * @return ViewInterface
     * @api
     */
    protected function resolveView()
    {
        return $this->objectManager->get(is_callable([$this, 'resolveViewObjectName']) ? $this->resolveViewObjectName() : $this->defaultViewObjectName);
    }
```

```php$this->resolveViewObjectName() ```

can return false so i suggest :

```php
        return $this->objectManager->get(is_callable([$this, 'resolveViewObjectName']) && $this->resolveViewObjectName() !== false ? $this->resolveViewObjectName() : $this->defaultViewObjectName);
```

NamelessCoder:
Solved and merged, thanks for the PR @filoucrackeur :)

PR #1727, opened 2019-10-30
[BUGFIX] Check if $this->resolveViewObjectName() is not false #1726

filoucrackeur:
fix #1278450972 TYPO3\CMS\Extbase\Reflection\Exception\UnknownClassException
Class does not exist. Reflection failed.

### DIFF
--- a/Classes/Controller/AbstractFluxController.php
+++ b/Classes/Controller/AbstractFluxController.php
@@ -270,7 +270,7 @@ abstract class AbstractFluxController extends ActionController
      */
     protected function resolveView()
     {
-        $viewClassName = (!method_exists($this, 'resolveViewObjectName') ? $this->defaultViewObjectName : ($this->resolveViewObjectName()) ?: $this->defaultViewObjectName);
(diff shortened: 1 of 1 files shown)


Classes/Controller/AbstractFluxController.php
```diff
@@ -270,7 +270,7 @@ protected function initializeView(ViewInterface $view)
      */
     protected function resolveView()
     {
-        return $this->objectManager->get(is_callable([$this, 'resolveViewObjectName']) ? $this->resolveViewObjectName() : $this->defaultViewObjectName);
+        return $this->objectManager->get($this->defaultViewObjectName);
```
NamelessCoder: This will be the right solution, but not yet - while we still have support for current LTS of TYPO3 we do need to call the method. Possible check could be:

```php
return $this->objectManager->get(!method_exists($this, 'resolveViewObjectName') ? $this->defaultViewObjectName : ($this->resolveViewObjectName() ?: $this->defaultViewObjectName))`
```

Issue #1724, opened 2019-10-21
Child elements get duplicated when copying container in workspace


cweiske:
When copying a container in a workspace, child elements get duplicated.

flux 9.2.0, TYPO3 9.5.9

# Steps to reproduce

1. Switch to workspace
1. Create a new page
1. Create a new container element on that page
1. Create a child element inside that container
1. Copy the container
1. Paste the container (on the same page or a new page)
1. Child elements appear twice (get duplicated)

PR #1729, opened 2019-11-12
[BUGFIX] Correctly copy and move records in workspaces

NamelessCoder:
Fixes issues with copying and moving records in a
workspace, and rendering nested content in pages
that were moved in a workspace.

Close: #1724

### DIFF
--- a/Classes/Integration/HookSubscribers/DataHandlerSubscriber.php
+++ b/Classes/Integration/HookSubscribers/DataHandlerSubscriber.php
@@ -15,6 +15,7 @@ use FluidTYPO3\Flux\Provider\ProviderResolver;
 use FluidTYPO3\Flux\Utility\ColumnNumberUtility;
 use TYPO3\CMS\Backend\Utility\BackendUtility;
 use TYPO3\CMS\Core\Cache\CacheManager;
+use TYPO3\CMS\Core\Database\Connection;
 use TYPO3\CMS\Core\Database\ConnectionPool;
 use TYPO3\CMS\Core\Database\Query\Restriction\DeletedRestriction;
 use TYPO3\CMS\Core\DataHandling\DataHandler;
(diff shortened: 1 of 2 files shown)


NamelessCoder:
@cweiske would you mind giving this PR a try before I merge? It solves both copy and move operations in workspace.

I've confirmed operation on TYPO3 v10.2 and I'm fairly sure the paste commands are identical on 9.5 LTS.

NamelessCoder:
@cweiske Is the target page a workspace version / placeholder, and is the parent content element a workspace version / placeholder?

cweiske:
I created all pages and elements freshly in the workspace.

NamelessCoder:
Thanks - that looks the case I missed in the last test; when origin page is a placeholder. Will look into this!

cweiske:
No, I still have the same problem that the child element is not copied.

NamelessCoder:
Tested with TYPO3v10.2

* Copy from workspace version of page and content to new workspace version of page
* Copy from published page with workspace version of content to workspace version of page
* Copy from published page and content to workspace version of page
* Copy from published page and content to published page
* Copy published page in temp workspace (recurses to children of content)
* Copy published page in live workspace (recurses to children of content)

All should be working now, *if* the DataHandler command maps are the same on TYPO3v9 as on v10.2.

cweiske:
Ok, copying a container with children in the workspace works now.
It also works when copying a container with children in live.

There is a thing I do not know if it's a bug or not:
- Create container + child in live
- Switch to workspace, create a second child
- Switch back to live
- Copy container
- Paste container
- The pasted container has two children now, although live only had one.

NamelessCoder:
This probably is due to the removed workspace condition. I'll try to reintroduce this constraint, but only allow it to select records from either LIVE workspace and whichever workspace is currently active (sql `IN (0, $workspace)`). This should result in only being able to copy records that are published or belong to the active workspace (in your use case, the final copy should then have only one child record).

NamelessCoder:
Thanks Christian! Merge time :)

PR #1769, opened 2020-04-24
[BUGFIX] Corrects the selection options of the colPos field in the co…

svewap:
This patch makes sure that the colPos select element is correctly filled with values.
Without patch otherwise only the standard columns are loaded.


svewap:
No longer necessary due to the latest code changes

PR #1789, opened 2020-05-27
[BUGFIX] Make sure inline fal works with TYPO3 10

### LINKED ISSUE #1774: flux:field.inline.fal broken
```html
<html lang="en" data-namespace-typo3-fluid="true"
      xmlns="http://www.w3.org/1999/xhtml"
      xmlns:f="https://typo3.org/ns/TYPO3/Fluid/ViewHelpers"
      xmlns:flux="https://typo3.org/ns/FluidTYPO3/Flux/ViewHelpers"
      xmlns:v="https://typo3.org/ns/FluidTYPO3/Vhs/ViewHelpers">

<f:layout name="Content/Default"/>

  <flux:form id="quote" label="Zitat" description="Dieses Inhaltselement erstellt ein Zitat-Inhaltselement auf der Website.">
    <flux:form.option.icon>EXT:site_package/Resources/Public/Icons/element-content-default.svg</flux:form.option.icon>
    <flux:field.inline.fal name="settings.image" required="1" maxItems="1" minItems="1"/>
  </flux:form>
</f:section>

<f:section name="Main">
</f:section>

</html>
```

results in:

image

and the wizard is a record-wizard instead of a file wizard:

image

branch: 0c8a21a27e38e1fab680a8f913208c1eb0f5c4e5

jnugh:
This migrates the inline fal configuration from using the now removed
foreign_selector_fieldTcaOverride to overrideChildTca.

Fixes #1774

### DIFF
--- a/Classes/Form/AbstractInlineFormField.php
+++ b/Classes/Form/AbstractInlineFormField.php
@@ -1,4 +1,5 @@
 <?php
+
 namespace FluidTYPO3\Flux\Form;
 
 /*
@@ -128,7 +129,7 @@ abstract class AbstractInlineFormField extends AbstractRelationFormField impleme
     protected $levelLinksPosition = null;
 
     /**
-     * @var string
+     * @var array
      */
     protected $overrideChildTca;
 
@@ -423,9 +424,32 @@ abstract class AbstractInlineFormField extends AbstractRelationFormField impleme
     }
 
     /**
-     * @param string $overrideChildTca
+     * @param array $foreignSelectorFieldTcaOverride
      * @return RelationFieldInterface
-     * @deprecated
+     * @deprecated Please switch to overrideChildTca
+     */
+    public function setForeignSelectorFieldTcaOverride($foreignSelectorFieldTcaOverride)
+    {
+        $this->overrideChildTca = [
+            'columns' => [
+                'uid_local' => $foreignSelectorFieldTcaOverride
+            ]
+        ];
+        return $this;
+    }
+
+    /** 
+     * @return array
+     * @deprecated Please switch to overrideChildTca
+     */
+    public function getForeignSelectorFieldTcaOverride()
+    {
+        return isset($this->overrideChildTca['columns']['uid_local']) ? $this->overrideChildTca['columns']['uid_local'] : null;
+    }
+
+    /**
+     * @param array $overrideChildTca
+     * @return RelationFieldInterface
      */
     public function setOverrideChildTca($overrideChildTca)
     {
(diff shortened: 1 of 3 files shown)


NamelessCoder:
@jnugh Thanks for the patch - have you checked if this works the same on 8.7? We can disregard 7.6 but it still needs to support 8.7 at least for the time being. If it doesn't work the same there I guess we will need to duplicate those properties in TCA.

jnugh:
I actually did not try, however the old implementation was deprecated in 8.7 and I actually got the new structure from this: https://docs.typo3.org/c/typo3/cms-core/master/en-us/Changelog/8.7/Deprecation-80000-InlineOverrideChildTca.html

Classes/Form/AbstractInlineFormField.php
```diff
@@ -423,21 +423,22 @@ public function getLevelLinksPosition()
...
-        $this->foreignSelectorFieldTcaOverride = $foreignSelectorFieldTcaOverride;
+        $this->overrideChildTca = $overrideChildTca;
         return $this;
     }
 
     /**
      * @return string
      */
-    public function getForeignSelectorFieldTcaOverride()
+    public function getOverrideChildTca()
```
NamelessCoder, line 439: Old method must remain in place to avoid a breaking change. Old method can call new method transparently.

jmverges:
@jnugh @NamelessCoder do you need any help with this one?

jnugh:
@jmverges no, should be quite easy. I got distracted by an urgent task but I will come back to this once done. I'm still confident to update this PR till end of this week 😄

Classes/Form/AbstractInlineFormField.php
jnugh: I added a getter and setter that replicates the overridChildTca structure.

Classes/Form/Field/Inline/Fal.php
```diff
@@ -169,15 +169,17 @@ public function buildConfiguration()
         $configuration = $this->prepareConfiguration('inline');
         $configuration['appearance']['createNewRelationLinkTitle'] = $this->getCreateNewRelationLinkTitle();
         if (!empty($this->cropVariants)) {
-            $configuration['overrideChildTca'] = [
-                'columns' => [
-                    'crop' => [
-                        'config' => [
-                            'cropVariants' => $this->cropVariants
+            array_merge_recursive($configuration, [
```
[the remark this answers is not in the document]
jnugh: As I don't want to override `$configuration['overrideChildTca']` I need to only write to `columns`. I changed it however to detect if `$configuration['overrideChildTca']['columns']` exists (and create it if not) to set `$configuration['overrideChildTca']['columns']['crop']`

robsonrobi:
When will this fix be available ?

OrangeLeinad:
I'm also looking forward to see a stable fix.

Issue #1752, opened 2020-03-11
Flux Content under windows / wamp


supercaccia:
Under windows FCE flux content on folder example C:\wamp\www\base95\typo3conf\ext\theme\Resources\Private\Templates\Content are not found.

Page templates works correctly.

Same identically TYPO3 install works under linux OS.

In flux 9.2 all works correctly, error start on flux 9.3.

```
Oops, an error occurred!
Tried resolving a template file for controller action "Content->fce1colonna" in format ".html", but none of the paths contained the expected template file (). The following paths were checked: C:/wamp/www/base95/typo3conf/ext/theme/Resources/Private/Templates/
```

Geddo:
same for me. Files are there

ajikmi:
Same Problem on Windows 10.
on Mac and FreeBSD working correctly

NamelessCoder:
@supercaccia @Geddo @ajikmi - can one of you provide me with a complete stack trace? I don't have a Windows environment to reproduce this.

supercaccia:
[StackTrace.pdf](https://github.com/FluidTYPO3/flux/files/4604310/StackTrace.pdf)

Dear Claus,
thanks for taking this report seriously.
The required stack trace is attached in pdf format.
If you start with a clean installation there are no errors, but it is not possible to insert the FCEs because they do not appear in the list.
If you use a platform with content already inserted you can see the error if you open a page that contains the FCE.
I will be available if you need to test.
Thanks

NamelessCoder:
@supercaccia Thanks for the stack trace - this looks like it could be an issue with Fluid's TemplatePaths API, it appears your template root path is set correctly but the expected relative filename inside the template path is, for some reason, empty.

It seems to happen in https://github.com/TYPO3/Fluid/blob/master/src/View/TemplatePaths.php#L718 but since paths are not empty, I would expect to see at least one entry in the $tried variable which would then be reported as part of the exception message.

I'm not sure if you are familiar with xdebug but if you have that option, please what happens inside this loop on your system:

* The $relativePathAndFilename should definitely not be empty (the value should be something like `Content/Myfile`). The value of this variable is defined by this hardcoded array of possible relative file locations: https://github.com/TYPO3/Fluid/blob/master/src/View/TemplatePaths.php#L263
* The $paths variable shouldn't be empty before entering the loop; we expect $paths contains the same values as $this->templateRootPaths.
[...]

PR #1792, opened 2020-06-02
[BUGFIX] Respect absolute file paths on Windows systems

IchHabRecht:
The current checks for absolute file paths currently disregards Windows file paths.
This patch introduces PathUtility::isAbsolutePath to respect both, Unix and Windows paths.

Resolves: #1752


almfisch:
The patch is working on Windows 10, XAMPP, TYPO3 10.4.3.
Thank you for the patch.

NamelessCoder:
Thanks for taking care of this, @IchHabRecht!

Issue #1793, opened 2020-06-03
Extensionmanager throws TypeError when flux & fluidpages installed


jazzica:
After updating to 9.4.0 I get the following TypeError as soon as I activate fluidpages:

Return value of ext_updateaef8d0ff24e5d77ada8118a10c4da054::detectFluidPagesNotInstalledAndBackendLayoutFieldsReferenceFluidpages() must be of the type array, boolean returned

The problem is returning a false, I guess an empty array would do: 
https://github.com/FluidTYPO3/flux/blob/development/class.ext_update.php#L197

I'll open a pull request fixing the issue in a minute.

PR #1794, opened 2020-06-03
[BUGFIX] Ensure method returns array to match strict return

jazzica:
in detectFluidPagesNotInstalledAndBackendLayoutFieldsReferenceFluidpages

Fixes: #1793


NamelessCoder:
Thanks @jazzica :)

PR #1800, opened 2020-06-11
[BUGFIX] Add softref parsing on Text field

dogawaf:
Like for Input with link wizard, the Text field must have the softref TCA set to allow correct detection of soft reference.
This helps when exporting Flux contents as T3D.

### DIFF
--- a/Classes/Form/Field/Text.php
+++ b/Classes/Form/Field/Text.php
@@ -69,6 +69,7 @@ class Text extends Input implements FieldInterface
         $configuration['placeholder'] = $this->getPlaceholder();
         if (true === $this->getEnableRichText()) {
             $configuration['enableRichtext'] = true;
+            $configuration['softref'] = 'typolink_tag,images,email[subst],url';
             $configuration['richtextConfiguration'] = $this->getRichtextConfiguration();
         }
         $renderType = $this->getRenderType();


NamelessCoder:
Hi @dogawaf - I appreciate the patch but I'm inclined to reject it. The type-specific fields have been deprecated in favor of `flux:field` which can be used universally to create any field type, and through which you can specify any desired `configuration` options including the "softref" instruction. So I'd prefer to leave the type-specific fields untouched until their eventual removal - and encourage people to switch to `flux:field` if/when they encounter use cases that are not natively supported by the specific field type.

Would this be an option to solve your use case, switching to `flux:field` with custom `configuration`?

dogawaf:
Hi @NamelessCoder 

I forgot about the deprecation of the type-specific fields, so yes, switching to a `flux:field`+ a custom config will do the job.
(This is not the topic here but I will miss the type-specific fields that allow me to define flux form quite quickly and with good default).

What puzzles me in the issue I encountered is that actually `field.text` allows the users to enter rich text with links, and those links are not parsed by the softref parser, and thus not exported correctly in T3D. The same text entered in a native Textmedia is correctly exported.
For this very reason I would consider this a bug that should be fixed in the current versions of flux.

NamelessCoder (approved):

NamelessCoder:
> For this very reason I would consider this a bug that should be fixed in the current versions of flux.

That's reasonable enough, even if the text-specific field feature is deprecated.

Issue #1850, opened 2020-11-19
Column headlines not shown in translations


koehnlein:
Hi,

I'm using flux 9.4.2 in an installation with latest TYPO3 10.4. Unfortunately, the flux column headlines do not show the correct text, but only the value of the colPos field. (orange boxes)

The TYPO3 internal headlines and the flux column headlines in default language are shown correctly (green boxes)

![flux-column-headlines](https://user-images.githubusercontent.com/16088567/99691636-bc1f8300-2a89-11eb-8bb1-463fb2851bc6.png)

koehnlein:
I just saw, this problem already existed and once was fixed in #1620

NamelessCoder:
Solved by https://github.com/FluidTYPO3/flux/pull/1851

PR #1851, opened 2020-11-19
[BUGFIX] Show column label instead of colPos (page translation view) 

### LINKED ISSUE #1620: Backend: label of flux:grid.column not shown in translation
My setup:
TYPO3: 8.7.19
flux: dev-development
fluidpages: 4.2.0

I have a default language (EN) and a translation (de_DE). Now I've created a simple flux content element example with one column:
```html
    <flux:form id="container">
        <flux:grid>
            <flux:grid.row>
                <flux:grid.column colPos="0" name="column" label="Container-Content">
                </flux:grid.column>
            </flux:grid.row>
        </flux:grid>
    </flux:form>
</f:section>
```
**Problem:** in connected mode the label is missing in the german column, instead the colPos for the child records is shown.
label_connected_mode
(translated child record has uid 201870)

As you can see in the following screenshot, the label of the column "Container-Content" is shown in both languages, if the translation mode of the current page is "free mode". In this case the content elements are not translated to german but copied.
label_free_mode

BUT: my page template also has a label "Content", which is shown in every situation.
`<flux:grid.column colPos="0" name="Main" label="Content">`

koehnlein:
Copy 5fde1b6d2cd448e6ad64c7a3f70a7de4ad79190f which once fixed #1620 to solve #1850

### DIFF
--- a/Classes/Integration/PreviewView.php
+++ b/Classes/Integration/PreviewView.php
@@ -389,6 +389,11 @@ class PreviewView extends TemplateView
             $itemLabels[$name] = ($val['label'] ?? false) ? $this->getLanguageService()->sL($val['label']) : '';
         }
 
+        array_push(
+            $GLOBALS['TCA']['tt_content']['columns']['colPos']['config']['items'],
+            ...$layoutConfiguration['__items']
+        );
+
         $columnsAsCSV = implode(',', $layoutConfiguration['__colPosList'] ?? []);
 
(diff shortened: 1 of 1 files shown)


manuelselbach (commented):
This will also fix it for me. Tested with TYPO3 8.7.32 and 9.4.2 with this change applied.

NamelessCoder (approved):

PR #1869, opened 2021-02-24
[BUGFIX] Add missing parenthesis when checking for GLOB_BRACE.

kftw:
When using TYPO3 10.4.13 and PHP 7.4 and running debug mode I get the following error for the TYPO3 backend as well as the frontend:

```
PHP Warning: Use of undefined constant GLOB_BRACE - assumed 'GLOB_BRACE' (this will throw an Error in a future version of PHP) in /www/public/typo3conf/ext/flux/Classes/Utility/MiscellaneousUtility.php line 60
```


NamelessCoder:
Thanks for the PR @kftw - solved by merging https://github.com/FluidTYPO3/flux/pull/1871 which came above yours in the PR list. Usually I would have merged yours since it was created first but I didn't see until too late that they solve the same problem.

PR #1885, opened 2021-04-30
[BUGFIX] readd child elements toggler to flux containers

franzkugelmann:
With flux 9.2.0 and TYPO3 8 there was a toggler icon in the backend page module to hide the child elements of a flux (fluidcontent) container. In TYPO3 with the current flux version it is not shown anymore. The proposed patch readds that toggler.
flux-container-toggler


NamelessCoder:
Manually picked as https://github.com/FluidTYPO3/flux/commit/54d2732b53418ede0d85881da4c0f919cd4690b2 - thanks for the patch!

PR #1891, opened 2021-06-04
[FEATURE] Add support for field description properties.

nightsplinter:
In addition to labels, TYPO3 9.4 add a description TCA field property that can be used as help text.

This patch makes the property available for flux fields.

The property is only added to fields in TYPO3 version 9.4+.

See https://docs.typo3.org/c/typo3/cms-core/master/en-us/Changelog/9.4/Feature-85410-AllowTCADescriptionProperty.html


cweiske:
@nightsplinter is my apprentice; I tested the patch already.

NamelessCoder (approved):

Issue #1898, opened 2021-08-12
Subpages' layout cannot be determined if parent page is hidden or deleted


moe2k:
TYPO3 9.5.28
fluidpages: 9.5.0
flux: 9.5.0

Observed behavior:

- Below the root page, there is a page (I call it parent page), below are subpages
- the parent page has a certain layout set in "Layout - this page" ("layout - subpages" is set to "Same as parent").
- Subpages' "Layout - this page" is set to "same as parent" ("Layout - subpages" also).
- Hide or delete the parent page.
- The subpage cannot be rendered anymore, in the FE I get an error like "Oops, an error occurred! Code: 2021081212553580da354a" which is random. Exception behind it is: `TYPO3Fluid\Fluid\View\Exception\InvalidTemplateResourceException: Tried resolving a template file for controller action "Page->default" in format ".html", but none of the paths contained the expected template file (). The following paths were checked: [truncated]/private/typo3conf/ext/fluidpages/Resources/Private/Templates/ in [truncated]/vendor/typo3fluid/fluid/src/View/TemplatePaths.php:598`

Suspected Reason: The subpage cannot resolve the parent page's layout anymore because frontend databasre restrictions apply.

Expected behavior:
- When encountering an inaccessible parent page, flux walks up in the rootline and tries to resolve the layout from the next page up. It only fails when the top of the tree is reached without being able to resolve the layout.

supercaccia:
I confirm the problem.

In the file "typo3conf/ext/flux/Classes/Service/RecordService.php" replace the "function getSingle" with this code:

```
public function getSingle($table, $fields, $uid)
    {
        if (TYPO3_MODE === 'BE') {
            return BackendUtility::getRecord($table, $uid, $fields);
        }

        $queryBuilder = $this->getQueryBuilder($table);
        $restrictions = $queryBuilder->getRestrictions();
        $restrictions = $restrictions->removeByType(HiddenRestriction::class);
        $queryBuilder->setRestrictions($restrictions);

        $results = $queryBuilder
            ->from($table)
            ->select(...explode(',', $fields))
            ->where(sprintf('uid = %d', $uid))
            ->execute()
            ->fetchAll() ?: [];
        return reset($results) ?: null;
    }
```

We create a pull requst as soon as possible ;)

PR #1904, opened 2021-10-08
Fix: Subpages' layout cannot be determined if parent page is hidden or deleted

supercaccia:
Fixes FluidTYPO3/flux#1898


NamelessCoder:
Solved by a different approach in https://github.com/FluidTYPO3/flux/commit/86220d1072b7a9716358a42768adbbbd75cb1df1 which also improves performance.

PR #1924, opened 2022-02-08
[BUGFIX] Change cache name

### LINKED ISSUE #1923: Rename cache_runtime
I am getting deprecation entries
```
component="TYPO3.CMS.deprecations": Core: Error handler (BE): TYPO3 Deprecation Notice: Accessing a cache with the "cache_" prefix as in "cache_runtime" is not necessary anymore, and should be called without the cache prefix. in /var/www/html/public/typo3/sysext/core/Classes/Cache/CacheManager.php line 150
```

This line uses `cache_runtime`:

https://github.com/FluidTYPO3/flux/blob/3cc50474d8504bbbf2b043410b04110882332b69/Classes/Content/ContentTypeManager.php#L119

References:
https://docs.typo3.org/c/typo3/cms-core/main/en-us/Changelog/10.0/Deprecation-88366-DefaultCachingFrameworkCacheNamesChanged.html
The requirement seems arbitrary but it's also easy to change it.

crazy252:
Fixes #1923

### DIFF
--- a/Classes/Content/ContentTypeManager.php
+++ b/Classes/Content/ContentTypeManager.php
@@ -125,7 +125,7 @@ class ContentTypeManager implements SingletonInterface
                 }
 
                 $cacheName = 'runtime';
-                if (version_compare($version, 11, '<')) {
+                if (version_compare($version, 10, '<')) {
                     $cacheName = 'cache_runtime';
                 }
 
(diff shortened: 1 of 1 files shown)


NamelessCoder (changes requested):
On hold until support for TYPO3 8.7 and 9.5 is dropped, at which point the cache name will simply be changed to `runtime` in all occurrences. I'll do my best to remember and reuse this PR when that time comes!

NamelessCoder:
Fixed via https://github.com/FluidTYPO3/flux/commit/3c944b576e01745d0c77f49e977df539597b7b81

Issue #1824, opened 2020-09-03
Child Content Element's Colpos is not Updated on Copy


simare:
The colpos of a chid element  is not updated when copying a page if the field sorting of the child is greater than sorting of the parent.
This is because the child record is processed before the parent and thus ```DataHandlerSubscriber::$copiedRecords``` does not contain the uid of the parent record, so no colpos calculation is triggered.

KevinMcLean:
I have the same problem with a three-column container when I insert child elements there.

If you copy the whole container, it will only insert the container and no child elements.

Could someone adjust this?

baschny:
We also struggled with exact this same problem. This makes the whole system unusable to the customer where the workflows requires to constantly copy pages with nested content elements. "Sometimes it works" is the feedback from the customer and my debugging came to the same conclusion here: it depends on the order that the DataHandler processed the input if it "works or not". 

Is there any better solution in sight, or is this problem known already, @NamelessCoder?

baschny:
The DataHandler.php copies the content elements in reverse "sortby" order, this is why this happens:
https://github.com/TYPO3/TYPO3.CMS/blob/master/typo3/sysext/core/Classes/DataHandling/DataHandler.php#L3320-L3322

simare:
As far as I recall the problem occurs with container elements that were created before the colpos handling for child records changed.
You just need to move the container around on the same page so the sorting fixes itself. After that you can copy containers at will.

baschny:
@simare good comment! That could be a potential reason! Maybe then we could create a "migration" for this situation? have you managed to achieve such a thing?

simare:
@baschny I thought about creating a migration but it is a bit tricky to figure out the correct sorting value. The container should not change its relative position but the sorting value has to be greater than that of the child records. So you might have to update sorting for all content elements on the page.

Maybe some kind of brute force approach like adding 1000 to all content elements one a page, except for the children, would work.
We only had a couple of elements and fixed it manually.

NamelessCoder:
@baschny I don't have a concrete suggestion how to avoid this case, yet. I suspect it would be possible to sort the command map before the copy happens, to make sure that parents are copied before children, but that's as far as I've thought about it at this point. I'm hoping it is possible to do this without performing additional queries and without modifying the sorting value in DB but that all depends on how much data is available in the command map before it starts to process.

If we're lucky then it can be done with one of the pre-functions in the DataHandler hook using a relatively simple `usort`...

[...]

PR #1929, opened 2022-03-08
[FEATURE] Upgrade wizard for problematic content element "sorting" values

cweiske:
Copying pages will fail to adjust the colPos values when the "sorting" value
of child elements is larger than that of its parent record.
This leads to "lost" content records that are visible in the list view,
but not in the web view in the TYPO3 backend.

This problem happens at least when migrating from TYPO3 v7 to v8 while
upgrading to flux 9.5.0.
Old content records exhibit this problem, while newly created content
is fine - TYPO3 takes care of modifying the "sorting" values correctly.

This wizard fixes the problematic sorting values by finding the largest
child sorting value and adjusting the parent's "sorting" accordingly,
as well as all following content element siblings.

Resolves: https://github.com/FluidTYPO3/flux/issues/1824

### DIFF
--- a/Classes/Updates/MigrateColPosWizard.php
+++ b/Classes/Updates/MigrateColPosWizard.php
@@ -1,0 +1,165 @@
+<?php
+namespace FluidTYPO3\Flux\Updates;
+
+use FluidTYPO3\Flux\Utility\ColumnNumberUtility;
+use TYPO3\CMS\Core\Database\ConnectionPool;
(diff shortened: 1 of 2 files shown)


NamelessCoder (approved):

NamelessCoder (approved):

NamelessCoder:
Excellent - thanks @cweiske!

Issue #1927, opened 2022-02-17
Preview FCEs moved in DRAFT dont works


supercaccia:
TYPO3 v8,v10 flux9-dev

In draft workspace create for example a page that contains an FCE, for example a 2 column grid with content inside it and publish it.
After, in DRAFT move for example the FCE in anoter page layout column.
At this point if you preview the page on the frontend the FCE has lost its contents. The problem is solved by publishing.
According to my analysis this happens because moving an FCE creates a version of the FCE record but not of its contents.
The versioned FCE refers to the colPos of the live records and not to the draft ones (which do not exist).

joneppie:
Deleting and copying hidden elements also returns the following error:
```Argument 1 passed to FluidTYPO3\Flux\Provider\AbstractProvider::getGrid() must be of the type array, null given, called in /var/www/typo3-10er/public/typo3conf/ext/flux/Classes/Integration/HookSubscribers/DataHandlerSubscriber.php on line 433```

cweiske:
@joneppie The problem with deleted elements is #1939, and can be solved with #1957.

Issue #1914, opened 2021-12-15
Error moving FCEs in DRAFT workspace


supercaccia:
Reproduced in TYPO3 v8, v10 with flux master and flux development

Steps to reproduce:

1. In workspace create a page with an FCE (for example a grid) that contains elements **and publish the page**

![1](https://user-images.githubusercontent.com/7404955/146141727-3527569b-d1c9-4539-9b62-1b233c97e6a2.png)

2. For example move the FCE to another column of the page layout backend

![2](https://user-images.githubusercontent.com/7404955/146142276-3e4614ab-0b0f-42db-94c2-7a67eece69a1.png)

3. Now moves the FCE back to the same column it was in before the move

![3](https://user-images.githubusercontent.com/7404955/146141729-17760626-47bf-44f4-9af8-58c7afffc25d.png)

At this point, the following errors appear:

1: Record cannot be versioned because it is a placeholder for a moving operation
1: Could not be edited in offline workspace in the branch where found (failure state: 'Online record was not in a workspace!'). Auto-creation of version failed!

**Note: if you reload the page, the moving operation seems to have been successful anyway even if the error appeared**!

Issue #1905, opened 2021-10-11
Copy & paste FCE’s into workspace includes deleted elements and duplicate some contents


supercaccia:
Reproduced on TYPO3 8.7.32 & 10.4.21 with latest flux-development

To reproduce follow these steps:

1. Go DRAFT Workspace, insert an FCE (for example 1 column) and insert 3 contents (for example text & media) and publish all. If you copy & paste the FCE “1 column” all works correctly 
 
![1](https://user-images.githubusercontent.com/7404955/136744210-e853f8a0-8849-448d-bf3b-9df5b69ab3f7.JPG)

2. Delete for example the Elements3 and publish the page
 
![2](https://user-images.githubusercontent.com/7404955/136744211-5a2945ef-3540-4c70-a5c8-85310b839e6a.JPG)

Now if you copy & paste the FCE “1 column” this error appears (because the system tries to copy and paste the previously deleted record as well). BUT after refresh the copy and paste seems to have worked correctly
 
![3](https://user-images.githubusercontent.com/7404955/136744212-c0a685db-f2c3-4d45-b9ad-ecde8f2c51b6.JPG)
![4](https://user-images.githubusercontent.com/7404955/136744214-9c1926cf-16c4-4266-aa3a-d8743158570d.JPG)

3. Now try copy & paste the latest pasted FCE “1column”. After copy & paste the elements into FCE “1 column” are duplicated!

![5b](https://user-images.githubusercontent.com/7404955/136744883-57ab25a4-8db4-4118-bd1b-ba59ed034ace.JPG)

Possibile related:
#1895
#1896

PR #1933, opened 2022-03-29
[BUGFIX] Workspaces full compatibility #1927 #1914 #1905 #1942

supercaccia:
[BUGFIX] Workspaces full compatibility

This change makes FLUX full compatible with workspaces.
This version has been tested hard and used in production in an environment with 250 editors working with the workspaces.

Fixes FluidTYPO3/flux#1927
Fixes FluidTYPO3/flux#1914
Fixes FluidTYPO3/flux#1905
Fixes FluidTYPO3/flux#1942


supercaccia:
Fully Hard Tested in v8 and v10, but ia used for this patch "BackendUtility::getMovePlaceholder" which doesn't work in version 11 :(
I will try to correct as soon as possible, contributions to this patch are welcome :)

Kanti:
For everyone that used the branch in the past, We repushed it in our fork:
So you can require it from there: https://github.com/andersundsehr/flux/tree/pr-1933

Issue #1934, opened 2022-03-29
New upgrade wizard for problematic content element "sorting" values breaks upgrade wizard in TYPO3 11


fnagel:
The new upgrade wizard for ` [FEATURE] Upgrade wizard for problematic content element "sorting" values`  breaks the upgrade wizard BE module in TYPO3 11.5.8. Introduced in commit bc93c5f2dee10a89136151cac18ece0a286248de.

Screenshot:

![image](https://user-images.githubusercontent.com/116824/160717722-dc2e5cb8-eb51-4079-b2e4-60d9c18112e1.png)

@cweiske @NamelessCoder

cweiske:
I indeed only tested it on TYPO3 v9.
(It was introduced in #1929 with commit 0c269b0a5bf81727918c66904aa740bc6e4d00d5)

NamelessCoder:
> How should this be fixed? Do we need to drop support for typo3 v8 in flux for this?

I would of course use the new interface and then avoid registering the upgrade for v8 - it won't be very long until v8 support is dropped completely anyway.

fnagel:
This is still an issue in version 9.6.1 (24536535af7cc2f2390906d4641792046b0afbfa).

cweiske:
@fnagel did you try the patch #1935?

MohsinQK:
TYPO3 v10, Flux v9.6.1 problem is still there.

cweiske:
@MohsinQK did you try the patch https://github.com/FluidTYPO3/flux/pull/1935?

fnagel:
@cweiske Sorry for the late response on this one, I've never noticed you created an PR. Sorry for that. 

Please see the PR for my feedback: https://github.com/FluidTYPO3/flux/pull/1935#issuecomment-1117965028

MohsinQK:
> @MohsinQK did you try the patch #1935?

yes @cweiske it worked.

fnagel:
I've tested this in TYPO3 11.5.9 without issues. Thank you :-)

PR #1935, opened 2022-03-30
[BUGFIX] Use TYPO3v9 API for colPos upgrade wizard

cweiske:
The v8 API is not available on TYPO3v10 and v11, so to support them
we have to use the new upgrade wizard API.

We drop support for TYPO3v8 for that feature.

See
https://docs.typo3.org/c/typo3/cms-core/9.5/en-us/Changelog/9.4/Feature-86076-NewAPIForUpgradeWizards.html

Resolves: https://github.com/FluidTYPO3/flux/issues/1934
Related: https://github.com/FluidTYPO3/flux/pull/1929

### DIFF
--- a/Classes/Updates/MigrateColPosWizard.php
+++ b/Classes/Updates/MigrateColPosWizard.php
@@ -2,8 +2,10 @@
 namespace FluidTYPO3\Flux\Updates;
 
 use FluidTYPO3\Flux\Utility\ColumnNumberUtility;
+use Symfony\Component\Console\Output\OutputInterface;
 use TYPO3\CMS\Core\Database\ConnectionPool;
 use TYPO3\CMS\Core\Database\Query\Restriction\DeletedRestriction;
+use TYPO3\CMS\Core\Registry;
 use TYPO3\CMS\Core\Utility\GeneralUtility;
 
 /**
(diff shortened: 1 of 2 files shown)


NamelessCoder:
@cweiske Unforttunately I didn't have time to include this in the last release - would you mind rebasing this, then I can include it in the next one?

fnagel:
Tested this PR successfully with TYPO3 11.5.9. When using the commit before this PR (which means this commit: a496951565fe9817d5234bb9ce31cbc7cec7967a) the upgrade wizard module crashes, with this PR it does not.

Ping me, I can do do another test in TYPO3 11.5 (and 10.4 if needed) when the PR has been rebased.

cweiske:
I've rebased against development. No conflicts.

NamelessCoder (approved):

akroii:
This problem still occurs in v10.4.28
I also tried patch 1935

cweiske:
@akroii Two people used this patch and it worked for them. It works for me. Open a new issue and post the exception message and full stack trace there, but I think it's you doing something wrong.

michael-binder:
I can also confirm that the patch works with my setup
TYPO3 11.5.17
Flux 9.6.1 + patch
PHP 7.4

PR #1938, opened 2022-04-06
[BUGFIX] Error "Language -1 does not exist on site"

imilhormel:
Hello there!

In one of our projects we use a multi-lingual TYPO3 10.4 together with our custom flux elements extension. After enabling the TYPO3 feature "Fluid based page module" this error occurred on any page in the page module if the accessed page contains a container content element which got assigned to all languages:
image

We could reproduce this error using containers from an external flux elements extension as well.
With this adjustment the page module renders as expected:
image


NamelessCoder:
Fixed in https://github.com/FluidTYPO3/flux/commit/94a832aad86391c1f88fa08583016e5573a94b16 - thanks for your effort!

PR #1945, opened 2022-04-22
[TASK] Adjust wizard for new typo3 api

fanat98:
The update wizard does not work in TYPO3 11.
with these changes it worked for me


fanat98:
it is duplicated

Issue #1953, opened 2022-05-12
Using "cache_" prefix is deprecated in TYPO3v10


cweiske:
flux uses CacheManager to fetch cache instances, e.g. in `ContentTypeManager::getCache()`: https://github.com/FluidTYPO3/flux/blob/ad2bcab291ba87d5451e9b253ff74fc6d7e6a445/Classes/Content/ContentTypeManager.php#L119

In TYPO3 v10, the "cache_" prefix for all the cache names has been removed. Calling `CacheManager::getCache('cache_runtime')` is deprecated since then: https://docs.typo3.org/c/typo3/cms-core/10.4/en-us/Changelog/10.0/Deprecation-88366-DefaultCachingFrameworkCacheNamesChanged.html

Using it generates deprecation notices:
> TYPO3 Deprecation Notice: Accessing a cache with the "cache_" prefix as in "cache_runtime" is not necessary anymore, and should be called without the cache prefix. in /var/www/site/htdocs/typo3/sysext/core/Classes/Cache/CacheManager.php line 151

With TYPO3 v12, the fallback will be removed.

How can this be solved, given that flux supports TYPO3 8.7 and 9?
Core version specific code?

cweiske:
Seems to be a duplicate of #1923.

NamelessCoder:
Fixed via https://github.com/FluidTYPO3/flux/commit/3c944b576e01745d0c77f49e977df539597b7b81

PR #1954, opened 2022-05-13
[TASK] Do not use "cache_" prefix for caches in TYPO3v10+

cweiske:
In TYPO3 v10, the "cache_" prefix for all the cache names has been removed.
Calling CacheManager::getCache('cache_runtime') is deprecated since then:
https://docs.typo3.org/c/typo3/cms-core/10.4/en-us/Changelog/10.0/Deprecation-88366-DefaultCachingFrameworkCacheNamesChanged.html

Flux generated deprecation notices on TYPO3v10:

 TYPO3 Deprecation Notice: Accessing a cache with the "cache_" prefix as
 in "cache_runtime" is not necessary anymore, and should be called without
 the cache prefix.
 in typo3/sysext/core/Classes/Cache/CacheManager.php line 151

Resolves: https://github.com/FluidTYPO3/flux/issues/1953

### DIFF
--- a/Classes/Content/ContentTypeManager.php
+++ b/Classes/Content/ContentTypeManager.php
@@ -116,7 +116,8 @@ class ContentTypeManager implements SingletonInterface
             try {
                 $cache = $cacheManager->getCache('flux');
             } catch (NoSuchCacheException $error) {
-                $cache = $cacheManager->getCache('cache_runtime');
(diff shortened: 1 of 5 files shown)


crazy252:
I made also a PR for this change and the maintainer said it is on hold. #1924 is the PR

NamelessCoder:
Fixed via https://github.com/FluidTYPO3/flux/commit/3c944b576e01745d0c77f49e977df539597b7b81

Issue #1939, opened 2022-04-11
Delete content element not working if it is hidden


stoutik:
TYPO3 11.5.5
flux: dev-development

When I try to delete a hidden content element, I have the following error:

![image](https://user-images.githubusercontent.com/10999921/162787983-d8414248-c867-4a32-8b4b-3185f3b45ebe.png)

The problem comes from this file:

Classes/Integration/HookSubscribers/DataHandlerSubscriber.php

To solve the problem, you could implement the following adjustments:

```
 protected function getSingleRecordWithRestrictions(string $table, int $uid, string $fieldsToSelect): ?array
     {
         $queryBuilder = GeneralUtility::makeInstance(ConnectionPool::class)->getQueryBuilderForTable($table);
         $queryBuilder->select(...GeneralUtility::trimExplode(',', $fieldsToSelect))
             ->from($table)
             ->where($queryBuilder->expr()->eq('uid', $queryBuilder->createNamedParameter($uid, \PDO::PARAM_INT)));
         return $queryBuilder->execute()->fetch() ?:  [];
     }
```

```
protected function getSingleRecordWithoutRestrictions(string $table, int $uid, string $fieldsToSelect): ?array
    {
        $queryBuilder = GeneralUtility::makeInstance(ConnectionPool::class)->getQueryBuilderForTable($table);
        $queryBuilder->getRestrictions()->removeAll();
        $queryBuilder->select(...GeneralUtility::trimExplode(',', $fieldsToSelect))
            ->from($table)
            ->where($queryBuilder->expr()->eq('uid', $queryBuilder->createNamedParameter($uid, \PDO::PARAM_INT)));
        return $queryBuilder->execute()->fetch() ?: [];
    }
```

DanYouCan:
Hi there
Sorry for rehandle this issue. I use TYPO3 V10.4.28 with Flux V9.6.1. If I remove or delete a hidden flux content element I still got following error:

_Argument 1 passed to FluidTYPO3\Flux\Provider\AbstractProvider::getGrid() must be of the type array, null given, called in /home/webtecha/www/webtech.ch/typo3104/public/typo3conf/ext/flux/Classes/Integration/HookSubscribers/DataHandlerSubscriber.php on line 433_

If I activate that content element and then remove or delete it, it does work. Could you please recheck it and search for a fix?
Thanks in advance for your effort.

cweiske:
This bug has been caused by #1896.

@peterreichenbach correctly wrote there:
[...]

PR #1957, opened 2022-05-18
[BUGFIX] Allow deleting hidden content elements

cweiske:
Deleting content elements that are restricted in some way
(hidden, starttime, endtime, fe_group) is broken since commit
> [BUGFIX] Prevent error when moving/pasting content with deleted child

That commit was meant to fix a problem with deleted content elements.
This commit here actually only filters deleted elements instead of all.

Resolves: https://github.com/FluidTYPO3/flux/issues/1939


develth:
What to do to get this released?

NamelessCoder:
This would break the `undelete` command in DataHandlerSubscriber (triggered when you restore an element from Recycler). The method `getSingleRecordWithoutRestrictions` was intentionally introduced to allow loading records completely without restrictions, including deleted records. So we need a different approach here.

I've pushed b1d352e45de76c89185b3f77bf6da4319006deaf which aims to solve the same issue without changing the functionality of `getSingleRecordWithoutRestrictions`. If you can confirm that it solves the issue, it would be great if you could close the PR/issues ;)

PR #1982, opened 2022-08-18
[BUGFIX] Correctly load extension config

### LINKED ISSUE #1981: Extension configuration not loaded correctly in TYPO3 CMS 10 and above
I noticed, that the page layouts from `design/Template` weren't loaded in TYPO3 CMS 10.4. After debugging I found out, that the `plugAndPlay` setting always evaluated to false.

thommyhh:
Fixes #1981


NamelessCoder:
Duplicate of https://github.com/FluidTYPO3/flux/pull/1978 which came first. Regardless, thanks for your effort in reporting and proposing a fix! :)

Issue #1894, opened 2021-06-20
"Clear value" wizard in page config not present any more


yol:
Adding wizards via `['config']['wizards']` was removed for TYPO3 8.6: https://docs.typo3.org/c/typo3/cms-core/master/en-us/Changelog/8.6/Deprecation-79440-TcaChanges.html

The "clear value" wizard (rendered when `clearable="1"` on a form field) is still being included with this method: https://github.com/FluidTYPO3/flux/blob/3121e6d/Classes/Form/AbstractFormField.php#L227-L235

As far as I can tell, this should not work any more with any TYPO3 version supported by current vhs. It would have to be replaced with one of the new wizard registration methods.

cweiske:
Flux needs to convert the old "renderClearValueWizardField" into a FieldControl class as described in https://docs.typo3.org/m/typo3/reference-coreapi/10.4/en-us/ApiOverview/FormEngine/Rendering/Index.html#node-expansion

(The field tag attribute is `clear="1"`, not "clearable".)

NamelessCoder:
Solved with the PR contributed by @cweiske

PR #1989, opened 2022-08-25
WIP: [BUGFIX] Make "clear value" wizard work in TYPO3 v8+

cweiske:
The "clear value" field wizard that can be enabled with the "clear="1" attribute
on "<flux:field.*>" elements is broken in TYPO3 v8 and later, because
the registration of field wizards changed:
https://docs.typo3.org/c/typo3/cms-core/8.7/en-us/Changelog/8.6/Deprecation-79440-TcaChanges.html

We register a "field wizard" that renders a checkbox that allows removing
the value instead of simply writing an empty string into flexform.

WHAT DOES NOT WORK YET:
Actually removing the field from flexform, because
AbstractProvider::postProcessRecord does not get called anymore since
commit e9127768d7b64681becbf1fb7b3774cf21bb52f3
("[FEATURE] Use colPos instead of tx_flux_column")

Resolves: https://github.com/FluidTYPO3/flux/issues/1894

:warning:  I have no idea how AbstractProvider::postProcessRecord shall be merged into TceMain.php, so I'm leaving this PR as "Work in progress" open for others to continue.

### DIFF
--- a/Classes/Form/AbstractFormField.php
+++ b/Classes/Form/AbstractFormField.php
@@ -225,13 +225,9 @@ abstract class AbstractFormField extends AbstractFormComponent implements FieldI
         }
         $wizards = $this->buildChildren($this->wizards);
         if (true === $this->getClearable()) {
-            array_push($wizards, [
-                'type' => 'userFunc',
-                'userFunc' => UserFunctions::class . '->renderClearValueWizardField',
-                'params' => [
-                    'itemName' => $this->getName(),
-                ],
-            ]);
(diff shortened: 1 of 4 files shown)


NamelessCoder:
Processed locally and pushed as https://github.com/FluidTYPO3/flux/commit/8cf7381c328b49a4648eed63a66a17d23f7d1671

Calling of postProcessRecord has been restored via the Provider as it was before. Moving the implementation to DataHandlerSubscriber wouldn't give much benefit anyway - but would prevent some use cases from working and would still require two different implementations for pages/other tables.

Thank you *so much* for taking care of this one!

PR #1990, opened 2022-08-25
[BUGFIX] First argument to is_callable() in pages.php, since php8 has…

TomekZiaja:
… to be a class object instead of the classname

[BUGFIX] Change 1st Parameter for is_callable() in pages.php

Otherwise causes php warnings when editing page properties with php ^8.


NamelessCoder:
Duplicate of https://github.com/FluidTYPO3/flux/pull/1994 which is merged

PR #1996, opened 2022-09-12
Fix BUG calculateColumnNumberForParentAndColumn float instead of int returned

PHP-Expert:
Fix BUG 
Return value of FluidTYPO3\Flux\Utility\ColumnNumberUtility::calculateColumnNumberForParentAndColumn() must be of the type int, float returned

### DIFF
--- a/Classes/Utility/ColumnNumberUtility.php
+++ b/Classes/Utility/ColumnNumberUtility.php
@@ -45,7 +45,7 @@ abstract class ColumnNumberUtility
 
     public static function calculateColumnNumberForParentAndColumn($parentUid, $columnNumber): int
     {
-        return ($parentUid * static::MULTIPLIER) + $columnNumber;
+        return (int) floor(($parentUid * static::MULTIPLIER) + $columnNumber);
     }
 
     public static function calculateParentUidAndColumnFromVirtualColumnNumber($virtualColumnNumber): array


NamelessCoder:
I'm not sure I understand this one - given that all the involved variables are (or _should definitely be_) integers and no division is taking place, the output should always be an integer. What is the exact use case that causes it to be a float? I suspect the cause could be that $columnNumber is somehow a float (problem from Fluid template perhaps?) and forcibly converting it to integer risks hiding the actual problem instead of fixing it.

PHP-Expert:
This problem was raised due to other function calls. I haven't had called the function directly but got error

Return value of FluidTYPO3\Flux\Utility\ColumnNumberUtility::calculateColumnNumberForParentAndColumn() must be of the type int, float returned

NamelessCoder:
There should be a stack trace and ideally also some information about what the input values were. With that information you should be able to determine the actual cause of the problem. I need to know the true cause - if the problem is, as I suspect, that somewhere further up the call stack $columnNumber is defined as a float, then the cause is some kind of misconfiguration and the right solution would be to throw a more specific error or cast compatible values at the place where it is configured. If the method receives a value like `1.5` and we floor that to become `1` then this would only cause a different kind of problem (like orphaned child records or inability to load correct child records in FE).

NamelessCoder:
Please check if this is still an issue on `dev-development` - several patches have been made recently to cast various values and these patches have also introduced casting in `flux:content.get` which is where this problem starts in your particular use case. These new patches should prevent the invalid input, regardless if you use an integer colPos value or a string column name reference when you use `flux:content.get`.

Even if the error is prevented:

* You should also make sure that the template or templates involved in the specific FE request that triggers this problem, use only integers as value for the `area=""` argument on `flux:content.get`.
* And you should make sure that you do not have any occurrences in any template of `flux:grid.column` where the value of the `colPos` argument is anything but an integer. Only integer values from 0 to 99 are valid.

And if you find none of the two issues described above, some background:

This could very well be caused by Fluid itself. In some variations of template syntax, variable usages and PCRE versions you can sometimes see that an argument declared like `arg="123"` is internally converted to `(float)123`. Fluid tries to avoid problems by pseudo-casting such values by doing `$value = $value + 0` (which converts any numeric value stored in a string to the appropriate native type) but if the value is already a float (which can happen in some cases where PCRE matching returns an matched pattern as a float and other versions return it as a string) then the result of this pseudo-cast is also going to be a float.

PR #1999, opened 2022-10-05
 PHP 8.0.8 Typo3 11.5.16

davidk1982:
Hi,
here are some enhancements for using PHP 8 and Typo3 11.5 to prevent output fatal errors in frontend and backend. i hope it's helps to make the extension "flux" more stablity.

Best regards
DavidK

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -1,0 +1,5 @@
+include:
+  - project: devops/gitlab-ci
+    ref: release/1.1.0
+    file: /templates/composer.gitlab-ci.yaml
+
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -1,5 +1,0 @@
-include:
-  - project: devops/gitlab-ci
-    ref: release/1.1.0
-    file: /templates/composer.gitlab-ci.yaml
-
--- a/Classes/Controller/AbstractFluxController.php
+++ b/Classes/Controller/AbstractFluxController.php
@@ -24,7 +24,7 @@ use TYPO3\CMS\Extbase\Configuration\ConfigurationManagerInterface;
 use TYPO3\CMS\Extbase\Mvc\Controller\ActionController;
 use TYPO3\CMS\Extbase\Mvc\Exception\StopActionException;
 use TYPO3\CMS\Extbase\Mvc\View\ViewInterface;
-use TYPO3\CMS\Extbase\Mvc\Response;
+use TYPO3\CMS\Extbase\Mvc\response;
 use TYPO3\CMS\Fluid\View\TemplatePaths;
 use TYPO3Fluid\Fluid\Core\Rendering\RenderingContextInterface;
 use function get_class;
@@ -341,6 +341,13 @@ abstract class AbstractFluxController extends ActionController
      */
     protected function performSubRendering($extensionName, $controllerName, $actionName, $pluginSignature)
     {
+    
+    	if (isset($this->responseFactory)) {
+            $response = $this->responseFactory->createResponse();
+        } else {
+            $response = $this->objectManager->get(Response::class);
+        }
+    
         $shouldRelay = $this->hasSubControllerActionOnForeignController($extensionName, $controllerName, $actionName);
         if (!$shouldRelay) {
             if ($this->provider instanceof FluidProviderInterface) {
@@ -384,7 +391,7 @@ abstract class AbstractFluxController extends ActionController
                 'view' => $this->view,
                 'content' => $content,
                 'request' => $this->request,
-                'response' => $this->response,
+                'response' => $response,
                 'extensionName' => $extensionName,
                 'controllerClassName' => $foreignControllerClass,
                 'controllerActionName' => $actionName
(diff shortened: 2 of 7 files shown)


.gitlab-ci.yml
```diff
@@ -0,0 +1,5 @@
+include:
```
NamelessCoder, line 1: This file would need to be removed

Classes/Controller/AbstractFluxController.php
```diff
@@ -24,7 +24,7 @@
 use TYPO3\CMS\Extbase\Mvc\Controller\ActionController;
 use TYPO3\CMS\Extbase\Mvc\Exception\StopActionException;
 use TYPO3\CMS\Extbase\Mvc\View\ViewInterface;
-use TYPO3\CMS\Extbase\Mvc\Response;
+use TYPO3\CMS\Extbase\Mvc\response;
```
NamelessCoder, line 27: Unintended change? The class name, if present, definitely starts with an uppercase "R" and this would cause problems with composer autoloading on TYPO3 versions that have this class.

Classes/Integration/HookSubscribers/DataHandlerSubscriber.php
```diff
@@ -124,19 +124,25 @@ public function processDatamap_preProcessFieldArray(array &$fieldArray, $table,
...
                 );
 
                 if ($primaryConfigurationProvider && is_array($fieldArray[$fieldName]) && array_key_exists('data', $fieldArray[$fieldName])) {
                     foreach ($fieldArray[$fieldName]['data'] as $sheet) {
-                        foreach ($sheet['lDEF'] as $key => $value) {
-                            list ($possibleTableName, $columnName) = explode('.', $key, 2);
-                            if ($possibleTableName === $table && isset($GLOBALS['TCA'][$table]['columns'][$columnName])) {
-                                $fieldArray[$columnName] = $value['vDEF'];
-                            }
+                        foreach ($sheet['lDEF'] as $key => $value)
```
NamelessCoder, line 135: This chunk would need a whitespace/indentation and CGL fix for spaces and curly braces.

Classes/Integration/HookSubscribers/DataHandlerSubscriber.php
```diff
@@ -358,6 +364,7 @@ protected function recursivelyMoveChildRecords(string $table, int $parentUid, in
     protected function getSingleRecordWithRestrictions(string $table, int $uid, string $fieldsToSelect): ?array
     {
         $queryBuilder = GeneralUtility::makeInstance(ConnectionPool::class)->getQueryBuilderForTable($table);
+	$queryBuilder->getRestrictions()->removeAll()->add(GeneralUtility::makeInstance(DeletedRestriction::class));
```
NamelessCoder, line 367: This would break an intentional change to move soft-deleted child records when the parent is moved. Without this, undeleting a child record of a parent that was moved to another page would cause the child record to become orphaned. So this shouldn't be included - but in order to understand the reason for it, why did you add this constraint?

Classes/ViewHelpers/Content/GetViewHelper.php
```diff
@@ -127,9 +129,10 @@ public static function renderStatic(
         }
         $templateVariableContainer = $renderingContext->getVariableProvider();
         $record = (array) $renderingContext->getViewHelperVariableContainer()->get(FormViewHelper::class, 'record');
-
-        if ($GLOBALS['BE_USER']->workspace) {
-            $placeholder = BackendUtility::getMovePlaceholder('tt_content', $record['uid']);
+        
+        $workspaceId = GeneralUtility::makeInstance(Context::class)->getPropertyFromAspect('workspace', 'id');
```
NamelessCoder, line 133: This breaks backwards compatibility with TYPO3 versions that don't have the Context class. It is not safe to assume that it exists. We would need a check for `class_exists` and a fallback to use the old method in case it doesn't exist.

NamelessCoder (changes requested):
General: indentations must be converted to spaces.

Classes/Integration/HookSubscribers/DataHandlerSubscriber.php
davidk1982: I have added this contraint, because in the current development state it was not possible to delete hidden records. It endsup in an exception.

.gitlab-ci.yml
davidk1982: Yes, it's not needed here. I'm sorry my fault.

Classes/Controller/AbstractFluxController.php
davidk1982: Has been restored.

Classes/Integration/HookSubscribers/DataHandlerSubscriber.php
davidk1982: Fix for spaces and curly braces

Classes/ViewHelpers/Content/GetViewHelper.php
davidk1982: The Fallback was implemented

NamelessCoder:
I've recreated the pull request on https://github.com/FluidTYPO3/flux/pull/2005.

For future contributions:

* Please create one pull request per change (ideally your PR should be a single commit changing a single thing)
* Please create one new branch per change (each PR should use a unique branch - and don't use `development` as this will mess with your ability to sync your local branch after the PR is merged)

This makes it a lot easier to process the pull requests. FYI: next time you pull from `development` you will need to either reset your local branch to the state before the commits in this PR - or delete the branch and check it out again.

Thanks for your contrubution!

davidk1982:
Thank you for your hints. I will handle it like this in the future.

PR #2003, opened 2022-10-14
Fix undefined array key warning in PHP 8.1

dennismetz:


NamelessCoder:
Handled in https://github.com/FluidTYPO3/flux/commit/ac70e4ff4c8b16c9fba8eff1ff896e18f426ffc9 - but thanks for your effort in proposing a fix!

PR #2028, opened 2022-11-09
BUGFIX: In PageService getTemplateConfiguration.

monosize:
Although the PageLayout setting of the page is set, it is ignored. Instead, the layout of the parent page is used. This is not correct. I think the OR operator must be used here.
Tested it and it works.


monosize:
I close the request because it doesn't work after all.

PR #2030, opened 2022-11-09
[BUGFIX] Load entire record in BackendLayoutDataProvider->getBackendlayout

monosize:
only the uid of the page is read. But here all fields must be read, because later in PageProvider->getControllerActionReferenceFromRecord is checked whether the main template of the page is set. And this is then never set! Then the layout from the settings of the parent page is used by mistake.


NamelessCoder (approved):

NamelessCoder:
Thanks @monosize :)

Issue #2034, opened 2022-11-22
T3v10.4.32 - Flux 9.7.1 assumes extbase ActionController processRequest function to return a request object and throws exception


SadikAtMNM:
I am getting an exception with a Content Controller (which extends AbstractFluxController) action: "Uncaught TYPO3 Exception: Call to a member function getBody() on null | Error thrown in file /html/typo3/typo3conf/ext/flux/Classes/Controller/AbstractFluxController.php in line 492". 
I guess because in line 472 there is no check for version < 11 anymore and extbase does not return a request in v10.

Greetings

NamelessCoder:
Fixed with https://github.com/FluidTYPO3/flux/commit/f97c457998b4a965e0a399a109e550649ca6b2fc#diff-62ebf087ca0cba0e6dcc06b87cc0e4d3007fccfabaca8ce9d73967c3218f8d97R467

mrnpnv:
Hi, I just updated Flux to the latest version and can still reproduce the issue. I am currently using
- Flux 9.7.2
- TYPO3 9.5.37
- PHP 7.4.33

and receive the same error log message. Thanks!

kitzberger:
Same here. Exception still occurs with TYPO3 10 and Flux 9.7.2.

maximilian-walter:
As a workaround, I added the following method to my ContentController:

```php
public function processRequest(RequestInterface $request, ResponseInterface $response)
{
    parent::processRequest($request, $response);

    return $this->response;
}
```

Flux 9.7.2, TYPO3 10.4.36, PHP 7.4

PR #2035, opened 2022-11-22
Re-added version check for Typo3

hieblmedia:
On Typo3 v10 an error (Call to a member function getBody()) occurs on sub action controllers. 
Also reported here: https://github.com/FluidTYPO3/flux/issues/2034


NamelessCoder:
Thanks for the patch - but the issue was fixed with https://github.com/FluidTYPO3/flux/commit/f97c457998b4a965e0a399a109e550649ca6b2fc#diff-62ebf087ca0cba0e6dcc06b87cc0e4d3007fccfabaca8ce9d73967c3218f8d97R467

PR #2046, opened 2022-12-06
[FIX/2045] ColumnNumberUtility::calculateParentUid colPos parameter

### LINKED ISSUE #2045: FluidTYPO3\Flux\Utility\ColumnNumberUtility::calculateParentUid(): Argument #1 ($colPos) must be of type int, string given, called in /var/www/.../public/typo3conf/ext/flux/Classes/Integration/Overrides/BackendLayoutView.php on line 119
Parameter colPos should be cast as an integer

lucmuller:
should be cast as an integer

Fixes #2045


NamelessCoder:
Thanks for bringing this to my attention. Meanwhile, a type-cast has been added through a different commit and the problem is solved (in `dev-development`).

PR #2105, opened 2023-07-18
[BUGFIX] Cast "id" Form attribute to string

mrwhy-orig:
If $arguments['id'] is an integer an error gets thrown. This way the integer gets casted to a string and doesn't break the processing.


NamelessCoder:
Be careful when casting in combination with a null coalesce; the cast will be processed first and the null coalesce will stop working, unless you add parenthesis around the expression being null coalesced.

NamelessCoder:
I would also recommend not using integer values for the form ID - while it does work in most cases, the ID is used in several contexts where a string value is expected (e.g. XLF). Though I wouldn't go so far to throw an exception in this case, it's probably better to cast it.

mrwhy-orig:
Thanks for the heads up! Will do so.

Issue #2143, opened 2023-12-13
Incompatibility with 12.4.9 at translated content


svewap:
The error occurs as soon as I want to edit a translation.

```
Uncaught TYPO3 Exception: FluidTYPO3\Flux\Integration\MultipleItemsProcFunc::execute(): Argument #2 ($formDataProvider) must be of type TYPO3\CMS\Core\DataHandling\ItemProcessingService, TYPO3\CMS\Backend\Form\FormDataProvider\TcaSelectItems given | TypeError thrown in file /var/www/html/vendor/fluidtypo3/flux/Classes/Integration/MultipleItemsProcFunc.php in line 36
```

I fixed it by modifying MultipleItemsProcFunc and adding ItemProcessingService:

```
    public function execute(array &$parameters, FormDataProviderInterface|ItemProcessingService $formDataProvider): void
    {
```

tobiasgraeber:
@NamelessCoder We also just ran into this and debugged already a bit. 

We think the reason tracks down to this changes here:

- https://github.com/TYPO3/typo3/commit/1c2b6a80101866451719c99af607192aac82c491

To be more specific in that function here: 

- https://github.com/TYPO3/typo3/commit/1c2b6a80101866451719c99af607192aac82c491#diff-83b6e1dbbf02a43046bc231ce4b936bf1bf15ea866eed93b58c208b74f14d826L1272

Around line 1335 - 1340 (old Version) the if-clause changed.

And now, we are not running into the return at https://github.com/TYPO3/typo3/commit/1c2b6a80101866451719c99af607192aac82c491#diff-83b6e1dbbf02a43046bc231ce4b936bf1bf15ea866eed93b58c208b74f14d826L1272 anymore 

We also thought about the same solution like suggested by @svewap

PR #2144, opened 2023-12-14
[BUGFIX] Allow ItemProcessingService as param in execute function of …

tobiasgraeber:
Resolves https://github.com/FluidTYPO3/flux/issues/2143 also see https://github.com/FluidTYPO3/flux/issues/2143#issuecomment-1855795050

### DIFF
--- a/Classes/Integration/MultipleItemsProcFunc.php
+++ b/Classes/Integration/MultipleItemsProcFunc.php
@@ -10,6 +10,7 @@ namespace FluidTYPO3\Flux\Integration;
 
 use TYPO3\CMS\Backend\Form\FormDataProviderInterface;
 use TYPO3\CMS\Core\Utility\GeneralUtility;
+use TYPO3\CMS\Core\DataHandling\ItemProcessingService;
 
 class MultipleItemsProcFunc
 {
@@ -32,7 +33,7 @@ class MultipleItemsProcFunc
         $GLOBALS['TCA'][$table]['columns'][$field]['config']['itemsProcFunc'] = $newFunction;
     }
 
-    public function execute(array &$parameters, FormDataProviderInterface $formDataProvider): void
+    public function execute(array &$parameters, ItemProcessingService | FormDataProviderInterface $formDataProvider): void
     {
         $table = $parameters['table'];
         $field = $parameters['field'];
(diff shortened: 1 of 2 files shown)


Classes/Integration/MultipleItemsProcFunc.php
```diff
@@ -32,7 +33,7 @@ public static function register(string $table, string $field, ?string $additiona
         $GLOBALS['TCA'][$table]['columns'][$field]['config']['itemsProcFunc'] = $newFunction;
     }
 
-    public function execute(array &$parameters, FormDataProviderInterface $formDataProvider): void
+    public function execute(array &$parameters, ItemProcessingService | FormDataProviderInterface $formDataProvider): void
```
NamelessCoder, line 36: Unfortunately I can't accept code with union types - Flux still has to support TYPO3v10 which means supporting PHP 7.4 which sadly doesn't support union types. If this method has to support two different classes as second argument, we have to remove the type and add a PHPDOC instead (which only describes this single argument and notes the two possible classes as type; we don't PHPDOC things that are implied by strict types).

Tests/Unit/Integration/MultipleItemsProcFuncTest.php
```diff
@@ -31,10 +32,12 @@ public function testRegistersFunction(): void
 
     public function testExecutesFunction(): void
     {
-        $provider = $this->getMockBuilder(FormDataProviderInterface::class)->getMockForAbstractClass();
+        $formDataProviderInterface = $this->getMockBuilder(FormDataProviderInterface::class)->getMockForAbstractClass();
```
NamelessCoder, line 35: If one of these classes only exists on some versions of TYPO3, then this test method should be split in two and a check made for `class_exists` and if it doesn't, do `$this->markTestSkipped('Skipped on v12');` so the test won't run and always fail on that TYPO3 version.

NamelessCoder:
(preconfigured reply about branching)

Because you created this pull request based on your fork's development branch, once it is merged you will need to either force-push, re-fork or re-create the development branch in your fork. For your own benefit, future pull requests should be created from a branch that you create in your fork with starting point development. That way you won't have conflicts in your fork's development branch; and you can safely create multiple pull requests with one commit each and one branch per pull request.

All FluidTYPO3 repositories use git flow (https://datasift.github.io/gitflow/IntroducingGitFlow.html). Your pull requests should all be created based on feature branches.

NamelessCoder:
Thanks for taking care of this, @tobiasgraeber

tobiasgraeber:
Thank you too! 😊

PR #2153, opened 2024-03-07
[BUGFIX] Handle possible DBALException

JanStorm:
We can ignore a missing database connection at this point. We cant do anything about a missing database connection. This Scenario is possible during first install.

In a Scenario, where flux is installed before first install, this happens:
1. Upon the last setup step, when finishing by sending the admin user credentials and some metadata, the BootCompletedEvent is fired which brings up the FluidTYPO3\Flux\Integration\Event\BootCompletedEventListener.
2. The `SpooledConfigurationApplicator` then in its `processData` calls `$this->spoolQueuedContentTypeRegistrations` in Line 71.
3. Then `$this->contentTypeBuilder->registerContentType` in Line 181 **requires a Database Connection** for registering the content types.
4. Because the **Database is not fully setup** yet, DBAL returns a TableNotFoundException, which is a subclass of the more generic DBALException.
5. This is nowhere caught and therefore we cant successfully end the first install process.

Tested in Typo3 11.5.36 with flux 10.0.9

### DIFF
--- a/Classes/Integration/Configuration/SpooledConfigurationApplicator.php
+++ b/Classes/Integration/Configuration/SpooledConfigurationApplicator.php
@@ -8,6 +8,7 @@ namespace FluidTYPO3\Flux\Integration\Configuration;
  * LICENSE.md file that was distributed with this source code.
  */
 
+use Doctrine\DBAL\DBALException;
 use FluidTYPO3\Flux\Builder\ContentTypeBuilder;
 use FluidTYPO3\Flux\Builder\RequestBuilder;
 use FluidTYPO3\Flux\Content\ContentTypeManager;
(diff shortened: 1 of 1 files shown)


NamelessCoder:
The exception appears to have a different class name on the version of Doctrine that's used by TYPO3v12. We'll need to handle the newer exception type, and if it isn't backwards compatible, handle both types and add the resulting warnings to the phpstan baseline.

JanStorm:
Thanks for looking into it! The exception class was deprecated and then renamed in the newest doctrine version. I've updated the catch case to a more specific class that exists in both the older and new versions of Doctrine.

NamelessCoder (approved):

PR #2166, opened 2024-07-05
[BUGFIX] no check for the existence of the aspect frontend.preview

idp001:
There is no check for the existence of the "frontend.preview" aspect, otherwise we will get an AspectNotFoundException and the site will crash.


NamelessCoder (approved):

NamelessCoder:
Thanks Denis!

PR #2176, opened 2024-08-20
[BUGFIX] make sure to have correct pid

franzkugelmann:
When you add a new page from list view via page (select position) then BeforeFlexFormDataStructureIdentifierInitializedEvent gives us the negative value of the page uid of the sibling to the choosen position. We need the real parent id though (as the documentation implies is always given).

### DIFF
--- a/Classes/Builder/FlexFormBuilder.php
+++ b/Classes/Builder/FlexFormBuilder.php
@@ -70,6 +70,10 @@ class FlexFormBuilder
                 // template selection values as part of our DS identifier.
                 // This is NOT necessary if the input record contains an explicitly selected page layout, hence the
                 // added check above before entering this condition block.
+                if ((int)$record['pid'] < 0) {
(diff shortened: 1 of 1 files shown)


NamelessCoder (approved):

NamelessCoder:
Thanks for the fix @franzkugelmann!

Because you created this pull request based on your fork's development branch, once it is merged you will need to either force-push, re-fork or re-create the development branch in your fork. For your own benefit, future pull requests should be created from a branch that you create in your fork with starting point development. That way you won't have conflicts in your fork's development branch; and you can safely create multiple pull requests with one commit each and one branch per pull request.

All FluidTYPO3 repositories use git flow (https://datasift.github.io/gitflow/IntroducingGitFlow.html). Your pull requests should all be created based on feature branches.

NamelessCoder (approved):

PR #2198, opened 2024-12-04
[BUGFIX] Fix localization record merging logic

monosize:
Updated the logic to correctly merge localized records by checking for the existence of '_LOCALIZED_UID' in the $contentObject data. The $record was missing the '_LOCALIZED_UID' field, and therefore proper data localization never occurred.

This change ensures that records are only merged if a valid localized version is available.

I extracted $this->recoredService->getSingle(...) from the array_merge function because on my system this never merged the localized record. Therefore, getSingle is stored in a separate variable and then, if available, merged.

### DIFF
--- a/Classes/Controller/AbstractFluxController.php
+++ b/Classes/Controller/AbstractFluxController.php
@@ -583,15 +583,18 @@ abstract class AbstractFluxController extends ActionController
             );
         }
 
-        if ($record['_LOCALIZED_UID'] ?? false) {
-            $record = array_merge(
-                $record,
-                $this->recordService->getSingle(
-                    (string) $this->getFluxTableName(),
-                    '*',
-                    $record['_LOCALIZED_UID']
-                ) ?? $record
+        if ($contentObject->data['_LOCALIZED_UID'] ?? false) {
+            $localized = $this->recordService->getSingle(
+                (string) $this->getFluxTableName(),
+                '*',
+                $contentObject->data['_LOCALIZED_UID']
             );
+            if ($localized) {
+                $record = array_merge(
+                    $record,
+                    $localized
+                );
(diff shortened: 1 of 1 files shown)


monosize:
Tested with TYPO3 12.4.23

monosize:
Maybe the content from the $contentObject can be used completely – without array_merge; I haven't tested that yet.

monosize:
I ran a test that uses $contentObject->data.
It worked for me. Is there any reason not to use it directly?


```php
        if ($contentObject->data['_LOCALIZED_UID'] ?? false) {
            return $contentObject->data;
//            $test = $this->recordService->getSingle(
//                (string) $this->getFluxTableName(),
//                '*',
//                $contentObject->data['_LOCALIZED_UID']
//            );
//            $record = array_merge(
//                $record,
//                $test ?: $record
//            );
        }
        return $record;
```

NamelessCoder (changes requested):
> Is there any reason not to use it (`cObj->data`) directly?

This may be different on TYPO3 versions, but in at least some of my testing back when writing the method, `cObj->data` was not trustworthy. It behaved differently in content vs. pages, and in `INT` and cached contexts. In some cases it even contained the data of a totally different element - which would be really bad.

So we do need to use `$record` and we do need to perform the overlay. At least for now.

NamelessCoder:
Sorry about the delay here. I believe this is fixed, either by https://github.com/FluidTYPO3/flux/pull/2253 whcih fixes the overlaying that happens inside the record service, or by https://github.com/FluidTYPO3/flux/commit/f71fe74e848718c3972a73af9cccb815c3aad183 which circumvents the record lookup to use the cObj data in almost all cases.

PR #2205, opened 2025-01-08
[BUGFIX] Overlay record with language specific variables if translation is available

### LINKED ISSUE #2204: BUG: Nested grid-elements are not overlayed/translated properly in multilanguage site in version 10.1.0.
**I have checked that the bug exists in the `dev-development` branch**
_Yes_

**I have checked that there are no already open issues or recently closed issues about this bug**
_Yes_

**Describe the bug**
By multilanguages pages grid elements always displaying "default" language, and not the selected one.

**To Reproduce**
1. Create page with at least 2 languages. (in my case DE as default and PL as secondary language)
2. Create grid element container and fill it with some content. Do the same for second language, or just translate default language elements. In my test case I have the content in "connected mode".
3. Open page using secondary language (in my case PL).
4. The grid container itself as well as content elements inside grid container are shown in default language (DE), and not as expected in secondary language (PL).

**Expected behavior**
Expected is to display content in selected language.

**Screenshots**

image

image

image

image

**Additional context**
The function "getSingle" in class "FluidTYPO3\Flux\Service\WorkspacesAwareRecordService" return data for selected row without variable "_LOCALIZED_UID", so it deliver different data than previous implementation "$record = $contentObject->data;". The "_LOCALIZED_UID" is missing so the content elements cannot be rendered in correct language.

Here small diff which is fixing the issue:
[flux-lang.txt](https://github.com/user-attachments/files/18336856/flux-lang.txt)

SylwesterDev:
Fixes: #2204

### DIFF
--- a/Classes/Controller/AbstractFluxController.php
+++ b/Classes/Controller/AbstractFluxController.php
@@ -585,6 +585,10 @@ abstract class AbstractFluxController extends ActionController
                 1729864698
             );
         }
+        
+        if (is_array($record)) {
+            $record = $tsfe->sys_page->getLanguageOverlay($table, $record);
+        }
 
         if ($record['_LOCALIZED_UID'] ?? false) {
             $record = array_merge(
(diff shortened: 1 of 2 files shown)


tgriessbach (approved):

dacostafilipe:
Is there something we can do here to help out?

NamelessCoder:
Please re-check this in either the `development` branch or the upcoming `11.1` release. The problem should be fixed by https://github.com/FluidTYPO3/flux/commit/f71fe74e848718c3972a73af9cccb815c3aad183.

PR #2206, opened 2025-01-13
Update ColumnPositionItems.php

Vincent-typo3-sp:


NamelessCoder:
Could you tell me a bit more about the context? A content record should never be passed to this function _without_ it having the colPos column in the record - and resolving a parent with a ficticious column number could cause trouble (returning no parent even if the record does have a parent that we're just not seeing). So I'm wondering what causes this to happen and thinking we probably need to catch this issue further up before we end up in this code.

Vincent-typo3-sp:
Thank you. I could not reproduce the bug.

PR #2208, opened 2025-01-20
[BUGFIX] Prevent array key exception in FieldViewHelper

derhansen:
Thie PR prevents the following exception being thrown.

```
Uncaught TYPO3 Exception FluidTYPO3\Flux\Form\AbstractFormField::setProtectable(): Argument #1 ($protectable) must be of type bool, null given.
```

### DIFF
--- a/Classes/ViewHelpers/FieldViewHelper.php
+++ b/Classes/ViewHelpers/FieldViewHelper.php
@@ -97,8 +97,8 @@ class FieldViewHelper extends AbstractFieldViewHelper
         /** @var array $arguments */
         $parent = static::getContainerFromRenderingContext($renderingContext);
         $field = Field::create($arguments instanceof ArgumentCollection ? $arguments->getArrayCopy() : $arguments);
-        $field->setClearable($arguments['clear']);
-        $field->setProtectable($arguments['protect']);
+        $field->setClearable($arguments['clear'] ?? false);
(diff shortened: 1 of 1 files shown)


NamelessCoder:
Hi @derhansen - this is a known problem caused by not flushing the Fluid template cache after upgrading an extension which uses a ViewHelper who has had argument changes. Specifically, it's caused by the compiled template having an incomplete set of arguments because the compiled template hasn't been recompiled. When templates are compiled, a complete set of arguments are built including all default values - and these two arguments do have default values.

So, null-coalescing or conditioning these arguments only hides the problem in this specific instance - the right solution, which works for any extension and any ViewHelper with this kind of changes, is to ensure Fluid's template cache gets flushed.

NamelessCoder:
I'll close this PR since we know the solution and attempts to work around it would just open a can of worms where every VH argument would have to implement fallback cases, which would potentially hide and obscure more issues than it would actually solve. There are also aspects of this problem that simply wouldn't be solved at all by null-coalescing; e.g. a change to a default argument value wouldn't cause an error but is almost as bad.

I would recommend pushing _in the TYPO3 copre_ for a better, more verbose and more explicit handling of Fluid (and other SimpleFileBackend!) cache flushes. I would personally consider a Fluid cache _not being flushed without any reported problems_ when `cache:flush` is called, a definite bug. But not one that can or should be solved in Flux.

PR #2228, opened 2025-02-04
[TASK] Add TCA "required" config option for TYPO3v12+

cweiske:
"eval=required" is not used anymore in TYPO3v12+, as described in the TCA migration message:
> The TCA field 'title' of table 'content_types' defines "required"
> in its "eval" list.
> This is not evaluated anymore and should be replaced
> by `'required' => true`.

To keep backwards compatibility with TYPO3v10 and 11, "eval=required" is kept.

Changelog entries:
https://docs.typo3.org/c/typo3/cms-core/main/en-us/Changelog/12.0/Deprecation-97035-RequiredOptionInEvalKeyword.html
https://docs.typo3.org/c/typo3/cms-core/main/en-us/Changelog/12.0/Feature-97035-UtilizeRequiredDirectlyInTCAFieldConfiguration.html

This change is necessary for the flux own content_types table, as well as configuration generated for flux form fields.

### DIFF
--- a/Classes/Form/AbstractFormField.php
+++ b/Classes/Form/AbstractFormField.php
@@ -12,6 +12,7 @@ namespace FluidTYPO3\Flux\Form;
 use FluidTYPO3\Flux\Form\Container\Section;
 use FluidTYPO3\Flux\UserFunction\ClearValueWizard;
 use TYPO3\CMS\Core\Utility\GeneralUtility;
+use TYPO3\CMS\Core\Utility\VersionNumberUtility;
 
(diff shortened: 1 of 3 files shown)


cweiske:
I added a second commit to remove deprecated TCA configuration in TYPO3v12, so that no TCA migration notices are shown anymore.

cweiske:
I've also adjusted the creation of flux form field configuration now.

cweiske:
Can I do something to help getting this merged?

PR #2229, opened 2025-02-04
[BUGFIX] Always fetch ContentObject from request in TYPO3v12+

cweiske:
The ContentObjectFetcher was introduced in #2190 to have a common API for fetching the current content object across different TYPO3 versions.

This class logs a deprecation notice in TYPO3v12:
> TYPO3 Deprecation Notice:
> ConfigurationManager->getContentObject() is deprecated since TYPO3 v12.4
> and will be removed in v13.0.
> Fetch the current content object from request attribute
> "currentContentObject" instead in
> vendor/typo3/cms-extbase/Classes/Configuration/ConfigurationManager.php
> line 96

To prevent this from happening, we check for the TYPO3 version and always use the request instead of the configuration manager in TYPO3v12+.

Note: In TYPO3v12, "$request->getAttribute('currentContentObject')" does not always return the content object.
We thus fall back to TypoScriptFrontendController, which always has it.

Related: https://github.com/FluidTYPO3/flux/pull/2190
https://github.com/FluidTYPO3/vhs/issues/1920
https://github.com/FluidTYPO3/vhs/issues/1921
https://github.com/FluidTYPO3/vhs/pull/1931

Documentation: https://docs.typo3.org/m/typo3/reference-coreapi/12.4/en-us/ApiOverview/RequestLifeCycle/RequestAttributes/CurrentContentObject.html

### DIFF
--- a/Classes/Utility/ContentObjectFetcher.php
+++ b/Classes/Utility/ContentObjectFetcher.php
@@ -22,16 +22,15 @@ class ContentObjectFetcher
             ? $configurationManager->getRequest()
             : ($GLOBALS['TYPO3_REQUEST'] ?? null);
 
-        if ($request && $configurationManager === null) {
+        if ($request) {
             $contentObject = static::resolveFromRequest($request);
         }
 
-        if ($contentObject === null) {
-            if ($configurationManager !== null && method_exists($configurationManager, 'getContentObject')) {
-                $contentObject = $configurationManager->getContentObject();
-            } else {
-                $contentObject = static::resolveFromRequest($request);
-            }
+        if ($contentObject === null
+            && $configurationManager !== null
+            && method_exists($configurationManager, 'getContentObject')
+        ) {
+            $contentObject = $configurationManager->getContentObject();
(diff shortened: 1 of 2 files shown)


Classes/Utility/ContentObjectFetcher.php
```diff
@@ -22,23 +23,29 @@ public static function resolve(?ConfigurationManagerInterface $configurationMana
             ? $configurationManager->getRequest()
             : ($GLOBALS['TYPO3_REQUEST'] ?? null);
 
-        if ($request && $configurationManager === null) {
+        if ($request
```
NamelessCoder, line 26: We can simplify this to just if-request-then-fetch-from-request.

The code below then only enters the fetch-from-configurationmanager case if it wasn't already found in request. So this should give the same result, but gives preference to Request instead of ConfigurationManager on pre-v12 (which should be fine).

cweiske: Adjusted.

PR #2230, opened 2025-02-04
[BUGFIX] Use associative array for TCA items

cweiske:
TYPO3v12 introduced associative array keys for items in TCA select fields: label, value, icon, group.
The indexed item arrays were deprecated at the same time, throwing warnings in TYPO3v12+:

> TYPO3 Deprecation Notice:
> FlexFormTools did an on-the-fly migration of a flex form data structure.
> This is deprecated and will be removed.
> Merge the following changes into the flex form definition "content.field":
> The TCA field 'dummyField' of table 'dummyTable' uses the legacy way of
> defining 'items'.
> Please switch to associated array keys:
> label, value, icon, group, description.
> in vendor/typo3/cms-core/Classes/Configuration/FlexForm/FlexFormTools.php
> line 1127

Changelog:
https://docs.typo3.org/c/typo3/cms-core/main/en-us/Changelog/12.3/Feature-99739-AssociativeArrayKeysForTCAItems.html#feature-99739-associative-array-keys-for-tca-items https://docs.typo3.org/c/typo3/cms-core/main/en-us/Changelog/12.3/Deprecation-99739-IndexedArrayKeysForTCAItems.html#deprecation-99739-indexed-array-keys-for-tca-items

### DIFF
--- a/Classes/Form/AbstractMultiValueFormField.php
+++ b/Classes/Form/AbstractMultiValueFormField.php
@@ -8,6 +8,7 @@ namespace FluidTYPO3\Flux\Form;
  * LICENSE.md file that was distributed with this source code.
  */
 
+use FluidTYPO3\Flux\Integration\FormEngine\SelectOption;
 use FluidTYPO3\Flux\Service\TypoScriptService;
 use TYPO3\CMS\Core\Utility\GeneralUtility;
 use TYPO3\CMS\Extbase\DomainObject\DomainObjectInterface;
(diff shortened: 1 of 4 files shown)


NamelessCoder:
I wrote a little helper class a while back to deal with this version difference: `\FluidTYPO3\Flux\Integration\FormEngine\SelectOption`.

Usage example:

```php
$items[] = (new SelectOption($optionLabel, $optionValue, $optionalOptionIcon))->toArray();
```

This helper should be used so we can keep the behavior switching in one place.

It's perfectly fine to also use this class in unit tests to create expected select options for comparisons.

Dealer's choice whether to use GU::makeInstance or just `new` to make the instances. I don't expect anyone to ever need to make an override of this class.

cweiske:
I've adjusted the code to use SelectOption.

NamelessCoder (approved):

PR #2245, opened 2025-05-22
Use new defaultValues key instead of tt_content_defValues for filtering items in content wizard

svewap:
Otherwise, the list of elements remains empty as soon as you set <flux:form.variable name=“allowedContentTypes” value=“xyz”/>

### DIFF
--- a/Classes/Integration/WizardItemsManipulator.php
+++ b/Classes/Integration/WizardItemsManipulator.php
@@ -46,7 +46,7 @@ class WizardItemsManipulator
                 $items = array_filter(
                     $items,
                     function (array $item) use ($enabledContentTypes, $fluxContentTypeNames) {
-                        $contentType = $item['tt_content_defValues']['CType'] ?? null;
+                        $contentType = $item['defaultValues']['CType'] ?? null;
                         return !in_array($contentType, $fluxContentTypeNames, true) ||
                             in_array($contentType, $enabledContentTypes, true);
                     }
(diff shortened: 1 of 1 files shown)


NamelessCoder:
@svewap Thanks - but is this also compatible with TYPO3v10?

svewap:
Puh, good question. I don't think. perhaps insert a version switch? Off the top of my head, I have no idea when this was changed in the core.

Rall3n:
@NamelessCoder @svewap This was changed in the core with version 13.0 (see [changelog for feature #102834](https://docs.typo3.org/c/typo3/cms-core/main/en-us/Changelog/13.0/Feature-102834-Auto-registrationOfNewContentElementWizardViaTCA.html#feature-102834-auto-registration-of-new-content-element-wizard-via-tca)).

For reasons of compatibility, would it not be feasible to chain the null coalescing operators to include the renamed item?

```php
$item['tt_content_defValues']['CType'] ?? $item['defaultValues']['CType'] ?? null
```

NamelessCoder:
@Rall3n That's a great solution - although I think I'd put the "modern" name first in the sequence of variables to check. No strict reason for that, it's just a preference (code elsewhere also prefers current practice first, legacy as fallback).

I'll merge and publish if you make that change!

ChristianOsswald:
[flux_wizard-item-manipulator.patch](https://github.com/user-attachments/files/20929136/flux_wizard-item-manipulator.patch)
With the suggested way. First checking defaultValues then tt_content_defValues

NamelessCoder:
Fixed by https://github.com/FluidTYPO3/flux/pull/2249

PR #2249, opened 2025-06-26
[BUGFIX] Fix `allowedContentTypes` in TYPO3 v13+ by using new `defaultValues` key

Rall3n:
Replaces #2245.

Fixes the option `allowedContentTypes` in grid configurations by using the new `defaultValues` key (in usage since TYPO3 v13), while also ensuring compatibility with versions lower than v13.


Rall3n:
@NamelessCoder Could we get this PR reviewed and the workflow approved?

NamelessCoder:
Thanks @Rall3n - sorry about the delay!

Issue #2241, opened 2025-04-16
BUG: ColumnNumberUtility::calculateParentUid fails in Workspace


typo3rel:
**I have checked that the bug exists in the `dev-development` branch**
_Yes_

**I have checked that there are no already open issues or recently closed issues about this bug**
_Yes_

**Describe the bug**
Workspace changes are not showing. Ajax error:
FluidTYPO3\Flux\Utility\ColumnNumberUtility::calculateParentUid(): Argument #1 ($colPos) must be of type int, null given, called in ...vendor/fluidtypo3/flux/Classes/Integration/HookSubscribers/ColumnPositionItems.php on line 32

**To Reproduce**
1. create on a workspace environment a new Flux element with flux:grid 
2. **move** element inside / maybe on move out too
3. show workspace changes 
4. See no List / See Ajax error

No error with:
```
public static function calculateParentUid(int $colPos): int
```
to
```
public static function calculateParentUid(int|null $colPos): int
```

Typo3 12.4.25
Flux 10.1.0

Apen:
+1 for this easy fix, i have the same error with direct_mail who call flux in TYPO3 v13

julebbb:
Hello I have the same error when I check content history in TYPO3 v12.4.34 and Flux 11.0.3

RKlingler:
I'm seeing the same issue, but in the regular history. I'm not using workspaces.
Adding the following check and early return to the start of _colPosListItemProcFunc_ seems to solve the issue:
```
if (!isset($parameters['row']['colPos'])) {
    return;
}
```
I think that would be preferable over making _ColumnNumberUtility::calculateParentUid_ accept null and consequentially return wrong or non-existing uids, as suggested in previous comments.

PR #2252, opened 2025-07-09
Add null to CalculateParentUid ColumnNumberUtility.php

julebbb:
Bugfix for this issues https://github.com/FluidTYPO3/flux/issues/2241


NamelessCoder:
Thanks for the pull request! This has however been fixed previously in https://github.com/FluidTYPO3/flux/pull/2254. Note we cannot accept mulitiple-typed arguments since we still have to support PHP 7.4.

PR #2272, opened 2025-11-04
[BUGFIX] fix all Implicitly marked null parameters (PHP 8.4)

Kanti:


RKlingler:
@NamelessCoder Would it be possible to have this merged and released soon? It would be really helpful for users who want to upgrade to PHP 8.4 or PHP 8.5 without opting for less strict error reporting.

NamelessCoder:
@RKlingler due to the public API method signature changes this means a new major version, so this will be held back. The next major version will be made as part of v14 compatibility which also drops v10 and v11 support along with making PHP 8.3 the minimum requirement.

RKlingler:
@NamelessCoder Thanks for your quick reply.

You’re probably right, but I’m not sure I fully understand. The concerned method arguments were nullable before and are nullable after the change. The minimum requirement of flux (as of now) is PHP 7.4, which supports nullable types. So is the method signature change not just a technicality, rather than a breaking change?

As you can tell I was hoping for a different answer. But regardless - thank you for clarifying. I will be looking forward to the next major version then.

Kanti:
My understanding is also that this is not a breaking change. 

See here this is not breaking in any of the supported version:
https://3v4l.org/W7YWZ#veol

NamelessCoder:
Superseded by https://github.com/FluidTYPO3/flux/pull/2288

